PR #3, opened 2018-05-15
Tempdeck gcodes added

andySigler:
## overview

This PR integrates the gcode commands and responses, such that they operate the device as expected (eg: set temp, get temp, get info, etc.)

In addition, some bugs and basic refactorings were addressed in this PR.

### DIFF
--- a/modules/temp-deck/temp-deck-arduino/gcode.cpp
+++ b/modules/temp-deck/temp-deck-arduino/gcode.cpp
@@ -1,0 +1,109 @@
+#include "Gcode.h"
+
+Gcode::Gcode() {
+    Serial.begin(115200);
+    Serial.setTimeout(2);
+}
+
+void Gcode::_send_ack(){
+    Serial.println("ok");
+    Serial.println("ok");
+}
+
+void Gcode::_strip_serial_buffer(){
+  for(uint8_t i=0;i<SERIAL_BUFFER_STRING.length();i++) {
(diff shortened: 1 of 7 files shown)


sanni-t (commented):
1. .ino code doesn’t give an error or any indication when a non existent gcode (or any non-gcode input) is received. It gives a valid ACK.
2. activate_bootloader() not being called anywhere
3. Does code.pop_command() have to be checked in a while loop? We are taking in one command at a time anyway since received_newline() only reads the serial buffer until ‘\r\n’ ? Changing that to an if{} and moving code.send_ack() inside of it would also solve pt 1 issue
4. update_target_temp name sounds misleading. We’re just updating the current temperature, right? Not updating the target to a different value?
5. We can use constrain() to constrain the value for setPWM (lights.cpp) instead of checking if it’s in range every time

andySigler:
1. Yes, that is how Smoothieware works too. I just decided to mimic it ¯\_(ツ)_/¯
2. added it 👍 
3. I'm still mimicking Smoothieware here. You could send, for example, the string `M104S25 M105 M115\r\n`, and it would process each command in order, and then finally return a single ack. Also with Smoothieware, as long as a `\r\n` was received, it always responds with an ack.
4. Good point, I changed it to `stabilize_to_target_temp`
5. Terrific!

sanni-t:
Ah! Ok. Good to know that's how smoothieware works

sanni-t (approved):
Looks good

PR #5, opened 2018-05-18
EEPROM writer

sanni-t:
Firmware to write temp deck Serial number and Model to EEPROM and read the same from EEPROM.


andySigler (approved):
🍠

PR #8, opened 2018-05-28
Temp-Deck: adds PID control, startup-animation, and overall cleanup

andySigler:
## overview

This PR, in addition to a fair amount of code-cleanup and bug fixing, mainly adds a PID control loop to control the peltiers' direction and intensity. This is done using [a simple arduino library](https://github.com/br3ttb/Arduino-PID-Library), with the TARGET and CURRENT temperatures as input, and the peltiers' intensity/direction as the OUTPUT.

## PID Tuning Values (K)

The tricky thing here was finding the correct Kp, Ki, and Kd values. Not only are these values different depending of the device is heating up or cooling down, BUT they are also different depending on the temperature region the device is targetting.

The K tuning values can be set over GCode with the `M104` command, to help testing and/or debugging. Add the arguments `P`, `I`, or `D` to set the corresponding tuning K values (floating point). Below is an example command setting the tuning values while rising up to 99-deg target:
```
M104 S99 P0.2 I0.1 D0.0
```
Some [hard-coded K tuning values are found at the top](https://github.com/Opentrons/opentrons-modules/blob/82051d2ab0845f6020dbf843a83a18b5fab155d9/modules/temp-deck/temp-deck-arduino/temp-deck-arduino.ino#L50) of `temp-deck-arduino.ino`. These worked very well, but I expect these will need to be updated in the future, as the values I've set were found while running the device at 31.5 volts (not the production 36 volts).

## Graphing the PID Loop

You can graph what the PID loop is doing, by using the Arduino IDE's Serial Plotter graphing tool. This is super useful when finding tuning values.

To have the firmware print over serial a format the graph is expecting, uncomment [the following lines](https://github.com/Opentrons/opentrons-modules/blob/82051d2ab0845f6020dbf843a83a18b5fab155d9/modules/temp-deck/temp-deck-arduino/temp-deck-arduino.ino#L77) from `temp-deck-arduino.ino`:
```arduino
#define DEBUG_PLOTTER_ENABLED
```

## Startup LED Animation

When the device first powers on, both 7-segment displays are at full brightness, and the color-bar is all the way off. If the firmware attempts to change these states during `setup()`, then a noticeable flash happens when it changes.

To avoid this, the `Lighs::startup_animation()` method has been updated to do the following:
1. Fade both 7-segment displays DOWN to 0% power (now everything is off)
2. Set the (powered-off) 7-segment displays to show the current temperature
3. Fade the 7-segment displays, and the color-bar, UP to their normal states


sanni-t (commented):
Acc to the pwm driver datasheet an 'i2c_write_delay' of 4.7us minimum is required. So i think a 10us delay should be a good value.. short yet gives enough margin

sanni-t (approved):
💡

PR #10, opened 2018-06-01
Temp-Deck: Adds custom bootloader and watchdog timeout to trigger

andySigler:
## overview

Adds a bootloader image to the temp-deck folder, `Caterina-Leonardo_modified.hex`, which pulses the green LED on while the device while in bootloader mode.


sanni-t (approved):
⏱

PR #16, opened 2018-07-02
Temp deck: updates after testing PP1

andySigler:
## overview

- Removes linear function converting thermistor values to an averaged plate-temp. Now it just uses the raw thermistor temperature which was found to be more accurate
- Lowers memory used by a bit, through removing unnecessary global variables
- Re-enables the power to be turned off when the temperature is set, to avoid over-powering the power-supply and resetting the MCU


sanni-t (commented):
What's the temp-deck-log-1-second.py file used for? Debugging?

andySigler:
@sanni-t yup it's for logging a device's temperature during QC

sanni-t (approved):
🥁

PR #23, opened 2018-08-20
feat(modules, bootloader): Common bootloader for modules, home mageck on power up 

### LINKED ISSUE #22: MagDeck: Disengage on Powercycle
As a MagDeck user, I would like my MagDeck to disengage as soon as it is turned on, so that I can put labware on top.

## Acceptance Criteria
- MagDeck disengages as soon as it is powered on

## Implementation Notes
- We will eventually want to add user-initiated disengage at other points in the run setup flow, since even if we home on boot the user can still end up trying to put labware on a fully raised magdeck (e.g. if they ended a run in the 'engaged' position and then went to run a different protocol). This can be done later without FW changes.

sanni-t:
- This PR adds a common bootloader for both modules, switching out the old ones. This bootloader is functionally the same as the previous individual bootloaders (that differed only in their PIDs). 
- The bootloader has its own PID which makes it easier to detect when a module switches to its bootloader when performing a DFU
- Also adds a feature to magdeck to home/ disengage every time it's powered on

Closes #22


andySigler (approved):

modules/mag-deck/mag-deck-arduino/mag-deck-arduino.ino
```diff
@@ -316,6 +323,7 @@ void loop() {
           gcode.print_device_info(device_serial, device_model, device_version);
           break;
         case GCODE_DFU:
+          gcode.send_ack(); // Send ack here since we not reaching the end of the loop
```
andySigler, line 326: good call! we should add this to the temp-deck as well

andySigler (changes requested):

modules/mag-deck/mag-deck-arduino/mag-deck-arduino.ino
sanni-t: Done

sanni-t:
Created new package with version 1.2.0 to be used with arduino. User should be able to simply update to new version from boards manager

andySigler (approved):
🍯

PR #27, opened 2018-08-22
(feat) Thermistor offset, and temp-deck QC script

andySigler:
## overview

This PR adds a small offset to the thermistor when going either below or above room temperature, to more accurately reflect the temperature of the top plate (found through QC'ing).

This PR also updates the QC script to no longer us the IR sensor, but only an external thermistor (found to be far more accurate).


modules/temp-deck/QC/test_temp_deck_EXTERNAL_SENSOR.py
```diff
@@ -0,0 +1,197 @@
...
+
+def connect_to_temp_deck(default_port=None):
+    tempdeck = temp_deck.TempDeck()
+    ports = []
+    for p in comports():
+        if p.pid == PID_TEMPDECK:
+            ports.append(p.device)
+    if not ports:
+        raise RuntimeError('Can not find a TempDeck connected over USB')
+    if default_port and default_port in ports:
```
sanni-t: Should this be `if default_port and not default_port in ports:` ?

modules/temp-deck/QC/test_temp_deck_EXTERNAL_SENSOR.py
```diff
@@ -0,0 +1,197 @@
...
+
+
+def connect_to_external_sensor(default_port=None):
+    ports = []
+    for p in comports():
+        if p.pid == PID_FTDI:
+            ports.append(p.device)
+    if not ports:
+        raise RuntimeError('Can not find a External Sensor connected over USB')
+    if default_port and default_port in ports:
```
sanni-t: Same here

sanni-t (approved):
🐙

PR #28, opened 2018-08-22
fix(eeprom): Fixes bug where magdecks were not able to have IDs writt…

andySigler:
…en to

### DIFF
--- a/eepromWriter/write_module_memory.py
+++ b/eepromWriter/write_module_memory.py
@@ -17,7 +17,7 @@ MODELS = {
 }
 
 
-def connect_to_temp_deck():
+def connect_to_module():
     port = None
     for p in comports():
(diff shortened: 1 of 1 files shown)


eepromWriter/write_module_memory.py
```diff
@@ -67,7 +67,9 @@ def _user_submitted_barcode(max_length):
         raise Exception(BAD_BARCODE_MESSAGE.format(barcode))
     # remove all characters before the letter T
     # for example, remove ASCII selector code "\x1b(B" on chinese keyboards
-    barcode = barcode[barcode.index('T'):]
+    for m in MODELS.values():
```
sanni-t: I think this should be `for m in MODELS.keys():`?

sanni-t (approved):
🔑

PR #30, opened 2018-08-28
(mag-deck): Lower motor current and update QC script

andySigler:
## overview

This PR lowers the motor's default current to be equal to the minimum current found to have all devices pass QC, plus an addition safety margin of 25%, bringing the current to 0.25 amps.

Also adds a small delay for after the stepper has been enabled, which seems to have eliminated the occasional skipping by giving the driver time to boot up and apply current to the motor.

And finally this PR makes three changes to the mag-deck QC script. 1) it updates the expected motor currents to match the firmware's default current, 2) it adds a check to see if the serial/model numbers have been saved to the device, and 3) adds a "probing" motion to each cycle of the QC test, where it moves upwards at low current and slowly, and the script checks to make sure there was no skipping on the motor when at that low current.


sanni-t (approved):
👾

PR #46, opened 2018-11-13
feat(modules): add firmware upload script

sanni-t:
This PR adds a python script that would enable uploading new firmware onto modules by ssh-ing into the robot. This would be useful in case usual methods of firmware update (server endpoint/ run app) aren't working.
It can also be used outside of the robot, i.e, to upload firmware on a module directly from a computer. Use the optional arguments to specify custom paths for config file, avrdude and bootloader port


andySigler (approved):
👍

PR #50, opened 2019-02-21
ci(repo): setup Travis-CI

sanni-t:
Closes [#2997](https://github.com/Opentrons/opentrons/issues/2997)

## overview ##
This PR implements CI for the modules repository. 
Complete process involves:
1. Installing Arduino on the machine
2. Linking libraries from the repository
3. Installing required boards
4. Building magdeck, tempdeck and thermocycler
5. Saving hex/bin files to a known location

## changes ##
Added:
- .travis.yml
- Makefile with arduino setup and build recipes
- All library files required by the three modules

Changed:
- all modules header files' references to lowercase

## todos ##
(Mike's comment)
* If we configure the `build-*1 tasks as non-PHONY file targets (i.e. the `hex` output) we can have incremental builds based on source
* The directories in `libraries` should probably be configured as [git submodules](https://git-scm.com/book/en/v2/Git-Tools-Submodules) rather than file copies


.travis.yml
```diff
@@ -0,0 +1,17 @@
+language: c
+sudo: false
```
mcous: `sudo: false` was [recently deprecated](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration) by travis, so this line has no effect. Can you remove it?

Makefile
```diff
@@ -0,0 +1,72 @@
...
+	NO_ARDUINO := true
+endif
+
+# setup Arduino
+.PHONY: setup-ino
+setup-ino:
+	# Download Arduino
+	mkdir -p $(HOME)/arduino_ide
+	# if not already cached, download and install/unpack arduino IDE
+	$(if $(NO_ARDUINO), wget --quiet $(DOWNLOAD_LINK), @echo "Arduino already present")
```
mcous: `wget` is not present by default on macOS; can this script use `curl -O` instead?

.travis.yml
```diff
@@ -0,0 +1,17 @@
...
+
+cache:
+  directories:
+    - ~/arduino_ide
+
+install:
+  - make setup-ino
```
mcous: It's a little redundant, but you may want to consider passing the OS here as a form of documentation:

```shell
make setup-ino CI_SYSTEM=$TRAVIS_OS_NAME
```

Makefile
```diff
@@ -0,0 +1,72 @@
...
+	$(if $(NO_ARDUINO), $(UNPACK_CMD))
+	# Link all the libraries that the modules use
+	mkdir -p $(ARDUINO_SKETCHBOOK_LOC)/libraries/
+	cp -a $(LIBRARIES_DIR)/. $(ARDUINO_SKETCHBOOK_LOC)/libraries/
+	echo -n "Libraries installed:"
+	ls $(ARDUINO_SKETCHBOOK_LOC)/libraries
+	# Add board packages
+	rm -rf $(ARDUINO15_LOC)
+	$(ARDUINO) --pref "boardsmanager.additional.urls=https://adafruit.github.io/arduino-board-index/package_adafruit_index.json, https://s3.us-east-2.amazonaws.com/opentrons-modules/package_opentrons_index.json" --save-prefs
```
mcous: 1. This Adafruit URL looks like it should be in an environment variable up top for readability
2. Is this opentrons JSON file looks like it's in the repository; are we not able to require it from there instead of this S3 link?

Makefile
```diff
@@ -0,0 +1,72 @@
...
+BUILD_PATH := $(HOME)/.arduino_builds
+DOWNLOAD_LINK := https://downloads.arduino.cc/$(ARDUINO_ARCHIVE)
+
+ifeq ($(wildcard $(ARDUINO)), )
+	NO_ARDUINO := true
+endif
+
+# setup Arduino
+.PHONY: setup-ino
+setup-ino:
```
mcous: Nitpick, but GNU (and also - to an extent -our) conventions would probably be to call this task `install`.

Something that we've (recently) found to be helpful is to also have an `uninstall` task that undoes whatever `install` does. In this case, it would wipe out `$(HOME)/arduino_ide` I think

Makefile
```diff
@@ -0,0 +1,72 @@
+# opentrons modules makefile
+# https://github.com/Opentrons/opentrons-modules
+
+SHELL := /bin/bash
+
+LIBRARIES_DIR := libraries
+MODULES_DIR := modules
+INO_DIR := $(HOME)/arduino_ide
+
+ARDUINO_VERSION ?= 1.8.5
```
mcous, line 12: 💯

mcous (changes requested):
This looks really good! I've got a couple little change requests.

It looks like this PR also pulls in some external libraries. Would it make sense to pull them in as [git submodules](https://git-scm.com/book/en/v2/Git-Tools-Submodules) rather than copying the source? It would make it easier to pull in updates / contribute upstream if necessary

Makefile
sanni-t: 1. Totally
2. The arduino board files need to be in a specific location which can't be changed easily (unless you modify arduino source code I think). And I've always found it to be a cleaner approach to install board files through Arduino instead of copy+pasting (I forget why I switched from copy+pasting). To reduce the extra efforts in installing board files every single time, I've pushed a changed to cache the board files and install only the missing ones

Makefile
sanni-t: @sfoster1 thoughts on `install`? I think if we're going to use this to set up arduino on our computers and build firmware without using the IDE then the `install` convention doesn't seem too bad..

Makefile
sanni-t: Thanks to @sfoster1

Makefile
sfoster1: Per https://www.gnu.org/prep/standards/html_node/Standard-Targets.html , the `install` target is for installing _the program the makefile is meant to produce_. GNU doesn't have an equivalent for our usage of install, which is "install the dependencies required to build the target" - usually GNU autotools projects would discover missing dependencies in the `configure` step, complain about it, and trust the user to fix it.

Consider that muscle memory autotools sequence: `./configure && make && [sudo] make install`. When we use `install` to install dependencies, we're using it in a totally different way.

/rant

Makefile
```diff
@@ -0,0 +1,96 @@
...
+	ARDUINO15_LOC := $(HOME)/Library/Arduino15
+else
+	ARDUINO := $(INO_DIR)/arduino
+	ARDUINO_ARCHIVE := arduino-$(ARDUINO_VERSION)-linux64.tar.xz
+	UNPACK_CMD := tar xf $(ARDUINO_ARCHIVE) -C $(INO_DIR)/ --strip-components=1
+	ARDUINO_SKETCHBOOK_LOC := $(HOME)/Arduino
+	ARDUINO15_LOC := $(HOME)/.arduino15
+endif
+
+BUILDS_DIR := $(HOME)/.arduino_builds
```
sfoster1: [nitpick] If at all possible let's try and keep this in the tree, like `./build` or something

Makefile
mcous: @sfoster1 @sanni-t our current `install` usage is definitely a convention I've either introduced or pushed strongly. My bad for somehow entirely skipping over the "install" section of the standard targets doc every time I've read it and going straight to what my brain wants `make install` to be based on my experience with `npm install`.

I'm fine calling what we currently call `install` and `uninstall` something else. `setup` and `teardown` come to mind?

sfoster1: `setup` and `teardown` sounds great!

sanni-t: Cool

.travis.yml
sanni-t: I made a change in the makefile to check the uname to determine MacOS or not. Would this still be useful?

mcous (approved):
⚽️ this looks great!

For the future:
- If we configure the `build-*1 tasks as non-PHONY file targets (i.e. the `hex` output) we can have incremental builds based on source
- The directories in `libraries` should probably be configured as [git submodules](https://git-scm.com/book/en/v2/Git-Tools-Submodules) rather than file copies

sfoster1 (approved):
Concur with Mike about both points. Good for now though.

PR #51, opened 2019-03-04
feat(modules): implement thermocycler communication using gcodes

sanni-t:
## overview ##
This PR adds gcode communication to thermocycler firmware

GCodes used: https://docs.google.com/document/d/1pfZpxrgS3ecJ3Wtdm7R4jHNMOnyOxCzrRnPm7S7BEYs/edit?usp=sharing

## changes ##
- added gcode class which basically takes over the serial port and listens to, recognizes and responds to gcodes received
- added tc_timer class which keeps track of 'hold time' set for a target temperature (works well but subject to change if it doesn't turn out to be reliable.. in which case, we could explore using timer interrupts)
- added readable lid status
- pieced together gcode requests and appropriate actions & responses

Chores- 
- moved all .ino initializations/declarations to a separate header file
- made the thermistor lookup table a static const

## review requests ##
The code works on a dummy thermocycler board (not connected to peripherals).
Please review if the gcodes and their responses are as desired. 
Also review code in general.

## note ##
Previous code that uses direct serial communication is still left in there to allow h/w team to continue their tests (since it is more verbose). Gcodes feature can be toggled by setting `USE_GCODES` to `false` (thermo-cycler.h, line 133)

Closes [#2950](https://github.com/Opentrons/opentrons/issues/2950), Closes [#2952](https://github.com/Opentrons/opentrons/issues/2952)


modules/thermo-cycler/thermo-cycler-arduino/gcode.h
```diff
@@ -0,0 +1,56 @@
...
+  GCODE_SET_LID_TEMP,
+  GCODE_DEACTIVATE_LID_HEATING,
+  GCODE_SET_PLATE_TEMP,
+  GCODE_GET_PLATE_TEMP,
+  GCODE_SET_RAMP_RATE,
+  GCODE_EDIT_PID_PARAMS,
+  GCODE_PAUSE,
+  GCODE_DEACTIVATE_ALL,
+  GCODE_DFU
+}Gcode_enum;
```
sfoster1: Instead of having a separate definition for `TOTAL_GCODE_COMMAND_CODES` you could have a `GCODE_MAX` as the last element of the enum here

modules/thermo-cycler/thermo-cycler-arduino/gcode.h
```diff
@@ -0,0 +1,56 @@
...
+        void targetting_temperature_response(float target_temp,
+                                             float current_temp, float time_left);
+        void idle_temperature_response(float current_temp);
+        bool read_number(char key);
+        void response(String msg);
+        void response(String param, String msg);
+
+    private:
+
+        String COMMAND_CODES[TOTAL_GCODE_COMMAND_CODES];
```
sfoster1: This is a static lookup table. It should either be file-scoped to gcode.cpp, if used only in gcode.cpp, or a `static const[expr if supported]` member of this class if used elsewhere.

Since this is C++, we also don't need to do the anonymous-typedef trick, and if we make this an enum class with explicit scoping we probably don't need to have GCODE_ at the beginning of all the names or the name of the enum itself.

In addition, it can be a little easier for this common enum-with-string-lookup case to do something like

```
#define GCODES_TABLE
GCODE_DEF(GCODE_GET_LID_STATUS, 0, "Mxxx") \
GCODE_DEF(GCODE_OPEN_LID, 1, "Myyy) \
....

#define GCODE_DEF(name, num, _) ##name=num,
typedef enum {
GCODE_TABLE
} Gcode_enum;

#undef GCODE_DEF
#define GCODE_DEF(_1, _2, str) ##str,

static String GCODE_NAMES[] = {
GCODE_TABLE
};
```

modules/thermo-cycler/thermo-cycler-arduino/gcode.h
```diff
@@ -0,0 +1,56 @@
...
+        String GCODE_BUFFER_STRING;
+        String SERIAL_BUFFER_STRING;
+        char CHARACTERS_TO_STRIP[3] = {' ', '\r', '\n'};
```
sfoster1: Same with the static lookup table. Should be `const static` or at file scope in gcode.cpp.

modules/thermo-cycler/thermo-cycler-arduino/gcode.cpp
```diff
@@ -0,0 +1,141 @@
...
+  code = GCODE_NO_CODE;
+  while (GCODE_BUFFER_STRING.length()) {
+    for (uint8_t i=GCODE_GET_LID_STATUS; i<=TOTAL_GCODE_COMMAND_CODES; i++) {
+      if (GCODE_BUFFER_STRING.substring(0, COMMAND_CODES[i].length()) == COMMAND_CODES[i]) {
+        GCODE_BUFFER_STRING.remove(0, COMMAND_CODES[i].length());
+        code = i;
+        return true;
+      }
+    }
+    GCODE_BUFFER_STRING.remove(0, 1);
```
sfoster1: Per [the source](https://github.com/arduino/ArduinoCore-sam/blob/master/cores/arduino/WString.cpp), `String::remove` incurs a `strcpy` every time you call it. It may be better to restructure this code to loop through the serial buffer, copying completed gcodes, and stopping when you find the first non-complete gcode. Then at the end you can trim the serial buffer with a single `remove`.

modules/thermo-cycler/thermo-cycler-arduino/gcode.cpp
```diff
@@ -0,0 +1,141 @@
+#include "gcode.h"
+
+Gcode::Gcode() {
+}
+
+void Gcode::_strip_serial_buffer() {
+  for(uint8_t i=0;i<SERIAL_BUFFER_STRING.length();i++) {
+    for (uint8_t c=0;c<sizeof(CHARACTERS_TO_STRIP);c++) {
+      if(SERIAL_BUFFER_STRING[i] == CHARACTERS_TO_STRIP[c]) {
+        SERIAL_BUFFER_STRING.remove(i, 1);
```
sfoster1: Remove incurs a strcpy and probably shouldn't be used in a tight loop like this. Consider changing this to a [range-for](https://en.cppreference.com/w/cpp/language/range-for) that copies characters if they're not in the `CHARACTERS_TO_STRIP` blacklist.

modules/thermo-cycler/thermo-cycler-arduino/gcode.cpp
```diff
@@ -0,0 +1,141 @@
...
+    }
+    GCODE_BUFFER_STRING.remove(0, 1);
+  }
+  return false;
+}
+
+bool Gcode::received_newline() {
+  if (Serial.available()){
+    if (SERIAL_BUFFER_STRING.length() > MAX_SERIAL_BUFFER_LENGTH) {
+      SERIAL_BUFFER_STRING = "";
```
sfoster1: We should probably indicate this failure more prominently.

modules/thermo-cycler/thermo-cycler-arduino/gcode.cpp
```diff
@@ -0,0 +1,141 @@
...
+    }
+    SERIAL_BUFFER_STRING += Serial.readStringUntil('\n');
+    if (SERIAL_BUFFER_STRING.charAt(SERIAL_BUFFER_STRING.length() - 1) == '\r') {
```
sfoster1: What's the alternative case? Do we require a `\r` before every `\n`?

modules/thermo-cycler/thermo-cycler-arduino/gcode.cpp
```diff
@@ -0,0 +1,141 @@
...
+  Serial.println("ok");
+}
+
+bool Gcode::read_number(char key) {
+  // NOTE: Has a flaw where if there's 2 `set_temperature` gcodes in the buffer,
+  //       with the first one without a hold time and the next having one,
+  //       then the next hold time will be captured and associated with the
+  //       first gcode
+  int starting_key_index = GCODE_BUFFER_STRING.indexOf(key);
+  if (starting_key_index >= 0) {
```
sfoster1: Could we use [strtof](https://en.cppreference.com/w/cpp/string/byte/strtof) instead of this?

modules/thermo-cycler/thermo-cycler-arduino/gcode.cpp
```diff
@@ -0,0 +1,141 @@
...
+  Serial.print(F("serial:"));
+  Serial.print(serial);
+  Serial.print(F(" model:"));
+  Serial.print(model);
+  Serial.print(F(" version:"));
+  Serial.print(version);
+  Serial.println();
+}
+
+void Gcode::targetting_temperature_response(float target_temp, float current_temp,\
```
sfoster1: I don't think we need the backslash.

modules/thermo-cycler/thermo-cycler-arduino/gcode.h
```diff
@@ -0,0 +1,56 @@
...
+        bool SEND_ACK_NEXT_UPDATE = false;
+
+        void _strip_serial_buffer();
+        void _parse_gcode_commands();
```
sfoster1: Is this implemented somewhere?

modules/thermo-cycler/thermo-cycler-arduino/gcode.h
```diff
@@ -0,0 +1,56 @@
...
+        void response(String msg);
+        void response(String param, String msg);
+
+    private:
+
+        String COMMAND_CODES[TOTAL_GCODE_COMMAND_CODES];
+        String GCODE_BUFFER_STRING;
+        String SERIAL_BUFFER_STRING;
+        char CHARACTERS_TO_STRIP[3] = {' ', '\r', '\n'};
+        bool SEND_ACK_NEXT_UPDATE = false;
```
sfoster1: Is this used?

modules/thermo-cycler/thermo-cycler-arduino/gcode.h
```diff
@@ -0,0 +1,56 @@
...
+  GCODE_SET_PLATE_TEMP,
+  GCODE_GET_PLATE_TEMP,
+  GCODE_SET_RAMP_RATE,
+  GCODE_EDIT_PID_PARAMS,
+  GCODE_PAUSE,
+  GCODE_DEACTIVATE_ALL,
+  GCODE_DFU
+}Gcode_enum;
+
+class Gcode{
```
sfoster1: We should have some more comments here describing how the interface to this class is supposed to work.

modules/thermo-cycler/thermo-cycler-arduino/lid.h
```diff
@@ -40,6 +40,13 @@
 #define STEPS_PER_MM 15  // full-stepping
 #define PULSE_HIGH_MICROSECONDS 2
 
+typedef enum{
+  UNKNOWN = 0,
+  CLOSED,
+  OPEN
+}Lid_status;
+
+static const char * LID_STATUS_STRINGS[] = {"Unknown", "Closed", "Open"};
```
sfoster1: This shouldn't be in the header, it will cause it to be defined in every translation unit. This should either be declared here and defined in the file or declared and defined as a `static const` public member of the `Lid` class.

modules/thermo-cycler/thermo-cycler-arduino/lid.h
```diff
@@ -40,6 +40,13 @@
 #define STEPS_PER_MM 15  // full-stepping
 #define PULSE_HIGH_MICROSECONDS 2
 
+typedef enum{
+  UNKNOWN = 0,
+  CLOSED,
+  OPEN
+}Lid_status;
```
sfoster1: C++ enums don't need anonymous typedefs, and could probably be `enum class`

modules/thermo-cycler/thermo-cycler-arduino/tc_timer.h
```diff
@@ -0,0 +1,25 @@
+#ifndef TC_TIMER_H_
+#define TC_TIMER_H_
+
+typedef enum{
+  IDLE,
+  RUNNING,
+  COMPLETE
+}Timer_status;
```
sfoster1: C++ enums don't need anonymous typedefs and could be `enum class`

modules/thermo-cycler/thermo-cycler-arduino/tc_timer.cpp
```diff
@@ -0,0 +1,48 @@
+#include "Arduino.h"
+#include "tc_timer.h"
+// #include "limits.h"
```
sfoster1: Commented out code

modules/thermo-cycler/thermo-cycler-arduino/tc_timer.h
```diff
@@ -0,0 +1,25 @@
...
+  public:
+    unsigned int total_hold_time;
+    TC_Timer();
+    void reset();
+    bool start();
+    unsigned long time_left();
+    Timer_status get_status();
+    void update();
+  private:
+    unsigned long _total_hold_time_millis = 0;
```
sfoster1, line 22: It's probably fine, but keep in mind that per [the arduino docs](https://www.arduino.cc/reference/en/language/functions/time/millis/) `millis` overflows after about 50 days, and institutions might keep their thermocycler plugged in all the time. It may be worth adding overflow handling.

modules/thermo-cycler/thermo-cycler-arduino/gcode.cpp
```diff
@@ -0,0 +1,141 @@
...
+      if(SERIAL_BUFFER_STRING[i] == CHARACTERS_TO_STRIP[c]) {
+        SERIAL_BUFFER_STRING.remove(i, 1);
+        i--;
+        break;
+      }
+    }
+  }
+}
+
+bool Gcode::pop_command() {
```
sfoster1: Do we need this to set a class variable that's polled later? Couldn't we have this return the code?

modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler.h
```diff
@@ -0,0 +1,141 @@
+#ifndef THERMOCYCLER_H
```
sfoster1, line 1: This is a general comment about this file - objects should really not be instantiated in header files. They can be declared here (usually as pointers or whatever) but their definitions should really live in a `.cpp` or `.ino` file. We should also really consider whether these need to be global, though because of arduino's `setup` and `loop` style they may in fact need to be.

sfoster1 (changes requested):
The overall structure is good but I have comments about a number of the individual parts of the system.

modules/thermo-cycler/thermo-cycler-arduino/gcode.h
```diff
@@ -0,0 +1,56 @@
...
+                                             float current_temp, float time_left);
+        void idle_temperature_response(float current_temp);
+        bool read_number(char key);
+        void response(String msg);
+        void response(String param, String msg);
+
+    private:
+
+        String COMMAND_CODES[TOTAL_GCODE_COMMAND_CODES];
+        String GCODE_BUFFER_STRING;
```
sfoster1: Also we should really leave the `ALL_CAPS_VARIABLE_NAMES` for data macros and `static consts`

modules/thermo-cycler/thermo-cycler-arduino/tc_timer.h
sanni-t: It's okay if the thermocycler is plugged in for more than 50 days because the timer will be running only in short durations while it's implementing cycles (which I doubt would be as long as 50 days). Here's a quick explanation on why this is overflow-safe: https://arduino.stackexchange.com/a/12591

modules/thermo-cycler/thermo-cycler-arduino/gcode.cpp
sanni-t: Yes, we require that each gcode line should end with a '\r\n'

modules/thermo-cycler/thermo-cycler-arduino/gcode.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+void GcodeHandler::_strip_serial_buffer()
+{
+  char temp[_serial_buffer_string.length()];
+  char * p;
+  p = temp;
+  bool skip = false;
+  for (const auto& sb_char : _serial_buffer_string)
+  {
+    for (const auto& c: _CHARACTERS_TO_STRIP)
```
sfoster1, line 21: [nitpick] If you declare `skip` inside the outer loop you won't have to reset it every time

modules/thermo-cycler/thermo-cycler-arduino/gcode.cpp
```diff
@@ -0,0 +1,202 @@
...
+    }
+    skip = false;
+  }
+  *p = '\0';
+  _serial_buffer_string = temp;
+}
+
+/* Searches for gcode in the given string and updates the index reference with
+ * index of gcode found */
+bool GcodeHandler::_find_command(String _string, uint8_t * code_int, uint8_t * index)
```
sfoster1:
```suggestion
bool GcodeHandler::_find_command(const String& _string, uint8_t * code_int, uint8_t * index)
```

This way you don't copy the string in every time

modules/thermo-cycler/thermo-cycler-arduino/gcode.cpp
```diff
@@ -0,0 +1,202 @@
...
+  int key_index = _command.args_string.indexOf(key);
+  if (key_index >= 0)
+  {
+    // Convert to number
+    String number_string = _command.args_string.substring(key_index+1);
+    uint8_t array_len = number_string.length()+1; // the number + array end char
+    char c_array[array_len];
+    number_string.toCharArray(c_array, array_len);
+
+    char *p = c_array;
```
sfoster1: If you're not modifying it you could use `number_string.c_str()` and avoid making `c_array` separately

modules/thermo-cycler/thermo-cycler-arduino/gcode.h
```diff
@@ -0,0 +1,83 @@
+#ifndef GCODE_H
+#define GCODE_H
+
+#include "Arduino.h"
+#include <cerrno>
```
sfoster1: Consider whether these includes have to be in the header or can live in `gcode.cpp`. If possible, they shouldn't live here.

sfoster1 (changes requested):
Looking a lot better, just a handful more small changes

sfoster1 (approved):
Looks great! Thanks for addressing all the feedback

PR #52, opened 2019-03-21
feat(modules): stub out serial and model for now

b-cooper:
Until there is a static place for serial number and model, let's stub them out for downstream
development in the meantime.


Laura-Danielle (approved):
🦇

sanni-t (changes requested):
I think instead of changing the response it'd be better to initialize device_serial and device_model to 'dummy_...'. That way when we do add the real serial and model numbers, it'd get overwritten

sanni-t (approved):

PR #53, opened 2019-03-25
feat(modules): add pause and resume to timer and gcodes

b-cooper:
Add gcodes for pause and resume function that map to according timer control calls.

This only controls the timer and doesn't effect the holding temperature.  The desired functionality is that a received "pause" gcode should hold at current temp and a sequential "resume" gcode should continue the ramp toward the target.  In place of the temperature-effect logic, I've placed TODO comments.


b-cooper:
Pause and Resume at the UX level should actually be handled by indefinite hold at temps (wait for temps). This means that the pause and resume functionality can be removed from the firmware

sfoster1:
Code LGTM

PR #54, opened 2019-03-26
feat(modules): add thermocycler lid position detection

sanni-t:
Closes [#2951](https://github.com/Opentrons/opentrons/issues/2951)

## overview 
This PR adds lid status detection by adding a timer interrupt mechanism that polls the lid's end switches. It's a non blocking mechanism that always knows the status of the lid and hence makes it easy to detect any unexpected behavior.

## changes
Added:
- timer interrupt setup
- timer ISR
- `Lid::switch_poller()`

Changed the pin names to match those in schematic

## review requests
Check for code sanity. 
Also test on a TC board. If using a dummy board, comment out lines 791-198 in the .ino file and make sure DUMMY_BOARD in `lid.h` is true. 
To test, pull the `PIN_COVER_SWITCH`(pin 8) and `PIN_BOTTOM_SWITCH`(pin 9) pins to GND to simulate a button press. On the dummy board, pin D8- the `PIN_COVER_SWITCH` isn't exposed, so you can change line 9 to `#define PIN_COVER_SWITCH PIN_SPI_MOSI` and use the pin labelled MOSI instead. 

## updated behavior after addressing comments
- lid detection is switched to a hardware interrupt with a debounce check
- DUMMY_BOARD macro is now enabled via make command eg `make build-thermocycler DUMMY_BOARD=true`

### DIFF
--- a/modules/thermo-cycler/thermo-cycler-arduino/lid.cpp
+++ b/modules/thermo-cycler/thermo-cycler-arduino/lid.cpp
@@ -1,6 +1,10 @@
 #include "lid.h"
 
 const char * Lid::LID_STATUS_STRINGS[] = { STATUS_TABLE };
+volatile bool cover_switch_toggled = false;
+volatile bool bottom_switch_toggled = false;
+volatile unsigned long cover_switch_toggled_at = 0;
+volatile unsigned long bottom_switch_toggled_at = 0;
 
 Lid::Lid()
 {}
@@ -46,6 +50,7 @@ void Lid::_setup_digipot()
 
 Lid_status Lid::status()
 {
+  _update_status();
   return _status;
 }
 
@@ -70,52 +75,30 @@ void Lid::_update_status()
     case 0x02:
       _status = Lid_status::open;
       break;
-    case 0x03:
+    default:
       _status = Lid_status::error;
       break;
   }
 }
 
-/* This poller keeps track of changes on the lid switches and updating the status.
- * Uses a bit shifting technique described here:
- * https://forum.arduino.cc/index.php?topic=451164.0
- */
-void Lid::switch_poller()
+void Lid::check_switches()
 {
-  bool cover_pin_bit = 0;
-  bool bottom_pin_bit = 0;
-  switch(_status)
+  if (cover_switch_toggled)
   {
-    case Lid_status::in_between:
-      cover_pin_bit = digitalRead(PIN_COVER_SWITCH);
-      bottom_pin_bit = digitalRead(PIN_BOTTOM_SWITCH);
-      break;
-    case Lid_status::closed:
-      cover_pin_bit = digitalRead(PIN_COVER_SWITCH);
-      bottom_pin_bit = digitalRead(PIN_BOTTOM_SWITCH) ^ (1 << 0);
-      break;
-    case Lid_status::open:
-      cover_pin_bit = digitalRead(PIN_COVER_SWITCH) ^ (1 << 0);
-      bottom_pin_bit = digitalRead(PIN_BOTTOM_SWITCH);
-      break;
-    case Lid_status::error:
-      cover_pin_bit = digitalRead(PIN_COVER_SWITCH) ^ (1 << 0);
-      bottom_pin_bit = digitalRead(PIN_BOTTOM_SWITCH) ^ (1 << 0);
-      break;
+    if (millis() - cover_switch_toggled_at >= 200)
+    {
+      cover_switch_toggled = false;
+      _is_cover_switch_pressed = !bool(digitalRead(PIN_COVER_SWITCH));
+    }
   }
-  _debounce_state[TO_INT(_Lid_switch::cover_switch)] = (_debounce_state[TO_INT(_Lid_switch::cover_switch)] << 1) |
-                                                       cover_pin_bit | 0xF800;
-  _debounce_state[TO_INT(_Lid_switch::bottom_switch)] = (_debounce_state[TO_INT(_Lid_switch::bottom_switch)] << 1) |
-                                                        bottom_pin_bit | 0xF800;
-  if (_debounce_state[TO_INT(_Lid_switch::cover_switch)] == 0xFC00)
+  if (bottom_switch_toggled)
   {
-    _is_cover_switch_pressed = !_is_cover_switch_pressed;
+    if (millis() - bottom_switch_toggled_at >= 200)
+    {
+      bottom_switch_toggled = false;
+      _is_bottom_switch_pressed = !bool(digitalRead(PIN_BOTTOM_SWITCH));
+    }
   }
-  if (_debounce_state[TO_INT(_Lid_switch::bottom_switch)] == 0xFC00)
-  {
-    _is_bottom_switch_pressed = !_is_bottom_switch_pressed;
-  }
-  _update_status();
 }
 
 void Lid::solenoid_on()
@@ -197,6 +180,7 @@ bool Lid::move_millimeters(float mm)
   {
     _motor_step(dir);
     _update_acceleration();
+    check_switches();
     if (dir)
     {
       if (_is_cover_switch_pressed)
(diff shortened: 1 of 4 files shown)


modules/thermo-cycler/thermo-cycler-arduino/lid.cpp
```diff
@@ -185,7 +244,30 @@ void Lid::setup()
 	pinMode(PIN_STEPPER_DIR, OUTPUT);
 	pinMode(PIN_STEPPER_ENABLE, OUTPUT);
 	motor_off();
-	pinMode(PIN_COVER_OPEN_SWITCH, INPUT);
-	pinMode(PIN_COVER_CLOSED_SWITCH, INPUT);
-	_setup_digipot();
+#if DUMMY_BOARD
```
sfoster1, line 245: Can we change this so that
- The presence of the `DUMMY_BOARD` define changes all the other stuff noted in the PR comment as well
- The `DUMMY_BOARD` define is not in code, but set by passing a `-DDUMMY_BOARD` on the compiler command line, and the presence of that define can be triggered with a makefile argument

modules/thermo-cycler/thermo-cycler-arduino/lid.cpp
```diff
@@ -46,28 +46,76 @@ void Lid::_setup_digipot()
 
 Lid_status Lid::status()
 {
-  if(_is_open_switch_pressed() && !_is_closed_switch_pressed())
+  return _status;
```
sfoster1, line 54: This can be a `const` method if it's just an accessor

modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
```diff
@@ -710,16 +762,32 @@ void setup()
   strip.show();
   startTime = micros();  // for the rainbow test
   rainbow_test();
+  setup_timer_interrupt();
 }
 
 void loop()
 {
 #if USE_GCODES
```
sfoster1, line 717: If we're going to make `DUMMY_BOARD` determined by compiler flag we could make `USE_GCODES` one too

modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
```diff
@@ -625,6 +625,58 @@ void read_from_serial() {
...
+
+/* Set up a timer interrupt to check lid switches' status every 20ms.
+ * Keeping the frequency high helps in getting rid of switch debounce. The timer ISR
+ * sets a flag when it's time to check the status of the switches. The loop()
+ * checks this flag and calls lid.switch_poller
+ */
+void setup_timer_interrupt()
+{
+  /* 48MHz/240 = 200kHz (8-bit divisor) | Select Generic Clock (GCLK) 4*/
+  REG_GCLK_GENDIV = GCLK_GENDIV_DIV(240) | GCLK_GENDIV_ID(4);
```
sfoster1: Where's the boardfile that provides these included? Should probably explicitly include it here.

[request, not required] Also, could we consider using [CMSIS](https://developer.arm.com/embedded/cmsis) for this? It's the standard cortex HAL and is a little more structured than your standard register bit twiddling. It looks like it's in the arduino library somewhere per a quick `find`.

modules/thermo-cycler/thermo-cycler-arduino/lid.cpp
```diff
@@ -46,28 +46,76 @@ void Lid::_setup_digipot()
...
+      break;
   }
-  else if(_is_closed_switch_pressed() && !_is_open_switch_pressed())
+}
+
+/* This poller keeps track of changes on the lid switches and updating the status.
+ * Uses a bit shifting technique described here:
+ * https://forum.arduino.cc/index.php?topic=451164.0
+ */
+void Lid::switch_poller()
```
sfoster1: There's a lot of logic here and I wonder if it might be simpler to set up these pins as edge-triggered interrupts and do the debouncing part in the main loop rather than a timer interrupt.

modules/thermo-cycler/thermo-cycler-arduino/lid.cpp
```diff
@@ -139,9 +186,8 @@ void Lid::_update_acceleration()
   _calculate_step_delay();
 }
 
-void Lid::move_millimeters(float mm, bool top_switch, bool bottom_switch)
+bool Lid::move_millimeters(float mm)
```
sfoster1, line 172: I think there's an issue here. This function is intended to loop until either
- it hits the specified distance
- the appropriate switch is triggered

But, I don't think it can sense the switch. The switch state is updated in `Lid::switch_poller`, which is called synchronously from `loop()` when the timer interrupt sets a flag. However, `move_millimeters` is _also_ called synchronously from loop. So `_is_cover_switch_pressed` and `_is_bottom_switch_pressed` can't actually be updated while this function is happening. This function looks like it will move the full distance every time and ignore the switches.

modules/thermo-cycler/thermo-cycler-arduino/lid.cpp
```diff
@@ -46,28 +46,76 @@ void Lid::_setup_digipot()
...
+    case 0x00:
+      _status = Lid_status::in_between;
+      break;
+    case 0x01:
+      _status = Lid_status::closed;
+      break;
+    case 0x02:
+      _status = Lid_status::open;
+      break;
+    case 0x03:
```
sfoster1: can we change this `case 0x03:` into a `default:` just for my anxiety

modules/thermo-cycler/thermo-cycler-arduino/lid.cpp
```diff
@@ -151,29 +197,42 @@ void Lid::move_millimeters(float mm, bool top_switch, bool bottom_switch)
...
 
 void Lid::open_cover()
 {
-  if (_is_open_switch_pressed())
+  if (_is_cover_switch_pressed)
   {
      return;
   }
-  move_millimeters(300, true, false);
+  move_millimeters(300);
```
sfoster1: This and the `close_cover` distance as magic numbers might want to be in headers or something I guess. Not a big thing they'll still be magic wherever they are

modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
```diff
@@ -625,6 +625,58 @@ void read_from_serial() {
...
+
+  REG_TC4_INTFLAG |= TC_INTFLAG_MC1 | TC_INTFLAG_MC0 | TC_INTFLAG_OVF;  // Clear interrupt flags
+  REG_TC4_INTENSET = TC_INTENSET_OVF;     // Enable TC4 ovf interrupt
+
+  REG_TC4_CTRLA |= TC_CTRLA_PRESCALER_DIV16 |   // Set prescaler to 16, f=12.5kHz (legal values 1,2,4,8,16,64,256,1024)
+                   TC_CTRLA_ENABLE;             // Enable TC4
+  while (TC4->COUNT8.STATUS.bit.SYNCBUSY);      // Wait for synchronization
+}
+
+volatile bool check_lid = false;
```
sfoster1: This should be declared and defined at the top of the file so it doesn't get lost in the shuffle and there's no declaration-position issues

sfoster1 (changes requested):
Couple style changes and a possible bug.

modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
sanni-t: arduino library has it for external hardware interrupts. Couldn't find anything for configuring timer interrupts. Maybe I missed it?

modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
sanni-t: We are never not going to use gcodes. The flag only exists for AndyC while he does PID tuning using Serial to dump all temperature values. We'll be removing that part soon

sfoster1 (approved):

sfoster1:
Tested on a feather, LGTM

PR #57, opened 2019-04-08
feat(thermocycler): switch to 25MHz pwm, add debug mode, use NC lid switches

sanni-t:
## overview
This PR makes a few changes required for the TC to function better
Closes [#2955](https://github.com/Opentrons/opentrons/issues/2955), Closes [#3316](https://github.com/Opentrons/opentrons/issues/3316)

## changes

1. Switched to higher frequency PWMing (25kHz)
2. Added a debug mode in gcode land. It can be used to get status readings of peripherals that aren't exposed to the user/api
3. Corrected lid switches' behavior (Changed from Normally Open to Normally Closed switches)
4. Implemented `set_lid_temp` gcode
5. Changed PID sampling time (more suitable sampling frequency)
6. Removed ramping up/down with different PID values (might change after science testing)
7. Refined fan control logic

## review requests

Check if the code makes sense and I have a dummy board to test on if you need


modules/thermo-cycler/thermo-cycler-arduino/fan.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include "high_frequency_pwm.h"
+
+Fan::Fan()
+{}
+
+void Fan::set_percentage(float p)
+{
+  enable();
+  uint8_t val = p * 255.0;
+  if (val < 0)
```
sfoster1: `val` is a `uint8_t` so these checks won't do much. Also, there's a good chance the value will be just truncated to 8 bits, not saturated, which might provide surprising results:

```
dev/opentrons ➤ cat test.c
#include <stdio.h>
#include <stdint.h>

int main(int ignore1, char** ignore2) {
  uint8_t i = 1024;
  printf("%d\n", i);
}
dev/opentrons ➤ gcc test.c
test.c:5:15: warning: implicit conversion from 'int' to 'uint8_t'
      (aka 'unsigned char') changes value from 1024 to 0 [-Wconstant-conversion]
  uint8_t i = 1024;
          ~   ^~~~
1 warning generated.
dev/opentrons ➤ ./a.out
0
```

All in all, it's probably better to check the range of `p` as a float before the conversion.

modules/thermo-cycler/thermo-cycler-arduino/fan.cpp
```diff
@@ -0,0 +1,49 @@
...
+    val = 255;
+  }
+  current_power = p;
+  hfq_analogWrite(_pwm_pin, val);
+}
+
+void Fan::setup_pwm_pin(uint8_t pwm_pin)
+{
+  _pwm_pin = pwm_pin;
+  _pwm_controlled = true;
```
sfoster1, line 21: Is there any way to switch off pwm?

modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
```diff
@@ -46,11 +49,11 @@ PID PID_Cover(
 
 void set_heat_pad_power(float val)
 {
-  int byte_val = max(min(val * 255.0, 255), 0);
+  uint8_t byte_val = max(min(val * 255.0, 255), 0);
   pinMode(PIN_HEAT_PAD_CONTROL, OUTPUT);
-  if (byte_val == 255) digitalWrite(PIN_HEAT_PAD_CONTROL, HIGH);
+  if (byte_val >= 255) digitalWrite(PIN_HEAT_PAD_CONTROL, HIGH);
```
sfoster1: Probably don't need the `>=` if this is a `uint8_t`

sfoster1 (changes requested):
Couple small changes but the structure looks good

modules/thermo-cycler/thermo-cycler-arduino/fan.cpp
sanni-t: Nope. Not in this code. The "switching off" is basically achieved by setting a pwm value of 0

sfoster1: 👍

sfoster1 (approved):
Changes look good!

PR #58, opened 2019-04-11
feat(modules): add lid heat control via gcodes and some cleanup

sanni-t:
## overview
This PR adds the gcode logic for controlling the lid's heat pad.
Also changes the format of the debug responses in order to be easy to parse.

Also added USE_GCODES and LID_WARNING (required for science testing since the prototype doesn't have lid movement control enabled) to compiler flags.

For science testing, compile thus:
```  make build-thermocycler DUMMY_BOARD=false USE_GCODES=true LID_WARNING=false ```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,9 +77,9 @@ setup:
 .PHONY: build
 build: build-magdeck build-tempdeck build-thermocycler
 
-DUMMY_BOARD := false
-USE_GCODES := true
-LID_WARNING := true
+DUMMY_BOARD ?= false
+USE_GCODES ?= true
+LID_WARNING ?= true
 
 .PHONY: build-magdeck
 build-magdeck:
--- a/Makefile
+++ b/Makefile
@@ -78,6 +78,8 @@ setup:
 build: build-magdeck build-tempdeck build-thermocycler
 
 DUMMY_BOARD := false
+USE_GCODES := true
+LID_WARNING := true
 
 .PHONY: build-magdeck
 build-magdeck:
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: Makefile [hunk 1]
```diff
@@ -78,6 +78,8 @@ setup:
 build: build-magdeck build-tempdeck build-thermocycler
 
 DUMMY_BOARD := false
+USE_GCODES := true
+LID_WARNING := true
```
sfoster1: Oh these should be assigned with [?=](https://www.gnu.org/software/make/manual/html_node/Setting.html), that way you can do `make build-rthermocycler USE_GCODES=false`

modules/thermo-cycler/thermo-cycler-arduino/lid.cpp
```diff
@@ -17,7 +17,7 @@ void Lid::_i2c_write(byte address, byte value)
   byte error = Wire.endTransmission();
   if (error)
   {
-    Serial.print("Digipot I2C Error: "); Serial.println(error);
+    // Serial.print("Digipot I2C Error: "); Serial.println(error);
```
sfoster1: Do we need this?

modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
```diff
@@ -362,10 +361,18 @@ void ramp_temp_after_change_temp()
             gcode.response("Lid", lid.LID_STATUS_STRINGS[static_cast<int>(lid.status())]);
             break;
           case Gcode::open_lid:
+            // Send ack right away since opening cover takes time > serial_response_timeout (driver)
```
sfoster1: Should we just increase the serial response timeout?

modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
```diff
@@ -362,10 +361,18 @@ void ramp_temp_after_change_temp()
...
+            gcode.send_ack();
             lid.open_cover();
+            // Indicate lid open action over
+            gcode.response("Lid opened");
```
sfoster1: Is there a gcode response for this?

sfoster1 (changes requested):
Couple minor changes, looks good overall though

modules/thermo-cycler/thermo-cycler-arduino/lid.cpp
sanni-t: I want to keep it so we know when the i2c isn't working but we don't have anything in the api to catch this message in a good way. The poller in the driver would freeze since there isn't a serial ack at the end of the message. I'm thinking adding an ack after that message would work so that the poller interrupt would catch it..

b-cooper (approved):
Apart from Seth's points, LGTM 🦅

sfoster1 (approved):
👍

PR #61, opened 2019-05-22
feat(thermocycler): make compatible with HW version 3

sanni-t:
## overview ##
This PR introduces TC firmware changes to account for changes in version 3 hardware.

Summary of changes is in this sheet: **(Check sheets 'Changes' and 'Pin change comparison')** https://docs.google.com/spreadsheets/d/1c5pcv1EJcuCT7ymIgZqZeUZNTy9E2JoBwoeIdMooxvE/edit?usp=sharing

Main changes that have been implemented are: 
1. Peltier pin interchanging
2. Some thermistors<->adc connection interchange
3. Addition of motor driver reset and fault detection
4. Switching to using internal uC DAC instead of digital potentiometer for generating motor Vref
5. Adding heat pad on/off (previously was just pwm control)

The new pin assignments also reassigns an RxLED pin whose behavior is built into the Feather M0 board definition. So, instead of having to modify Adafruit's board definitions every time, I have created 'Opentrons SAMD' board definitions (based on Adafruit's definitions) and added to our existing Opentrons package for Arduino. (Having our own board definitions will also make future revisions much easier)
To use this package:
-  paste this url into Arduino->preferences->Additional board URLs `https://s3.us-east-2.amazonaws.com/opentrons-modules/package_opentrons_index.json`
- go to Tools->Board->Boards manager and search for 'Opentrons SAMD' and install
OR
If you're on a Mac/Linux machine, do `make setup` from opentrons-modules repo. 

For review, you can ignore the board definition files since they are copied from Adafruit. The only change is in the `variant.h` file (commented out RxLED and TxLED defs)

You will also see that a few more flags have been introduced- 
HFQ_PWM: to switch between high & low freq PWM
OLD_PID_INTERVAL: to switch between more accurate but untested PID computation and a tested but slightly inaccurate one

There are a bit too many flags but they are being introduced so as to keep a single active branch instead of the 2-3 we have right now. They will eventually go down in number


sfoster1 (approved):
LGTM!

PR #63, opened 2019-06-01
fix(thermocycler): correct thermistor connections & heatsink fan cntrl

sanni-t:
## overview

This PR has corrections made by @andy-g-calderon after testing on version 3 hardware

## changes
- interchanged front left and right thermistor connections
- changed heatsink fan from active low to active high


andy-g-calderon (approved):
These  changes look good to me.

PR #66, opened 2019-06-13
feat(thermocycler): open lid on button press, fix heat pad enable for v3

sanni-t:
## overview ##

This PR adds hardware interrupt functionality to front button. Button press deactivates thermocycler and opens the lid. This is an interim behavior and could change in the future after TC UX design has been finalized.

Also fixes heatpad issue by adding the missing pin mode declaration

## changes ##
- add front button function
- change up/down direction of motor for HW v3 and use correct motor Vref
- define heatpad pin mode
- change motor movement function to use angle of movement instead of distance
- make LED strip the same color as button LED and correct to GRBW instead of RGBW

### DIFF
--- a/modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
+++ b/modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
@@ -10,7 +10,7 @@ Fan cover_fan;
 Fan heatsink_fan;
 
 #if RGBW_NEO
-Adafruit_NeoPixel_ZeroDMA strip(NUM_PIXELS, NEO_PIN, NEO_RGBW);
+Adafruit_NeoPixel_ZeroDMA strip(NUM_PIXELS, NEO_PIN, NEO_GRBW);
 #else
 Adafruit_NeoPixel_ZeroDMA strip(NUM_PIXELS, NEO_PIN, NEO_GRB + NEO_KHZ800);
 #endif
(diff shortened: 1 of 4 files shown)


modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
```diff
@@ -702,6 +702,27 @@ void read_from_serial() {
...
+  {
+    front_button_pressed = true;
+  }
+}
+
+void front_button_check()
+{
+  if (front_button_pressed)
+  {
+    delay(100); // debounce
```
sfoster1: This is a really long delay for something that will block the mainloop. 

We could probably fix this by changing the logic to
- Save the last time the button was hit
- When this is called, check to see if it's been less than some timeout since then, and don't do anything if so (to prevent this from getting called multiple times by the bounce)
- If it's been long enough that we should actuate again, write the current millis() to a variable somewhere and actuate the lid

sfoster1 (changes requested):
The debounce logic's delay is a bit long I think.

Also shouldn't we cancel heating when we pop the lid like this?

modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
sanni-t: Totally. I think I was in a hurry to just add some functionality to test the button.
I'll make these changes

sanni-t:
Updated PR description

sfoster1 (approved):

PR #67, opened 2019-06-17
fix(libraries, thermocycler): switch ADS1X15 adc library, fix TC max() bug

sanni-t:
## overview ##
Previous ADS1X15 library from adafruit had a bug where it wasn't waiting long enough for the adc conversion to be over. As a result, the firmware was reading the converted value of the previous reading. So the sequence of thermistor update would be: 
REQUEST therm1_val -> wait -> GET therm0_val  (previous conversion)
<therm1_val conversion completes>
REQUEST therm2_val -> wait -> GET therm1_val  (previous conversion)
<therm2_val conversion completes>
and so on ..

However, this bug wasn't apparent in our firmware because of another bug introduced by these lines- 
`max(adc_a->readADC_SingleEnded(index % ADC_PER_DEVICE), 0)`, `max(adc_b->readADC_SingleEnded(index % ADC_PER_DEVICE), 0)`

 `max()` is defined as a macro fn in Arduino.h: `((a)>(b)?(a):(b))`
Because of this, if we pass a function to it, the function will be executed twice. In our case, we were requesting thermistor readings twice, which was leading to the following sequence of things-
```
UPDATE therm1_val:
REQUEST therm1_val -> wait -> GET therm0_val  (previous conversion, wrong but never used)
<therm1_val conversion completes>
REQUEST therm1_val -> wait -> GET therm1_val   (previous conversion, correct and used)
<therm1_val conversion completes>

UPDATE therm2_val:
REQUEST therm2_val -> wait -> GET therm1_val  (previous conversion, wrong but never used)
<therm2_val conversion completes>
REQUEST therm2_val -> wait -> GET therm2_val   (previous conversion, correct and used)
<therm2_val conversion completes>
```

## changes ##
This PR fixes these bugs by switching to a better library and removing the max() function

Please note that changing the sampling rate is out of scope for this PR. It will be a part of the next PR which will aim for reducing thermistor update duration while maintaining its reliability and implementing a timer interrupt.


modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
```diff
@@ -15,6 +15,8 @@ Adafruit_NeoPixel_ZeroDMA strip(NUM_PIXELS, NEO_PIN, NEO_RGBW);
 Adafruit_NeoPixel_ZeroDMA strip(NUM_PIXELS, NEO_PIN, NEO_GRB + NEO_KHZ800);
 #endif
 
+unsigned long timeStamp = 0;
```
sfoster1, line 18: Should this be in here?

modules/thermo-cycler/thermo-cycler-arduino/thermistorsadc.cpp
```diff
@@ -97,12 +97,20 @@ float ThermistorsADC::heat_sink_temperature() {
 
 
 int ThermistorsADC::_read_adc(int index) {
+  int res = 0;
```
sfoster1, line 100: This, and the return type of `_read_adc`, should be `uint16_t` or some other unsigned type, and the result of `readADC_SingleEnded` should be immediately cast to unsigned - the ADS1115 in single-ended mode only reads voltages above ground reference. We may not have experienced this based on what we're setting the pga to, but if the read voltage was high enough that it had the MSB set we'd get a negative value.

sfoster1 (changes requested):

modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
sanni-t: I want to keep it until we finish changing the sample rate and adding timer interrupt. I want to monitor loop() duration for a little while

modules/thermo-cycler/thermo-cycler-arduino/thermistorsadc.cpp
sanni-t: I don't think we will get a -ve value. In single ended mode,  the max value we get is 7FFF.. an input beyond this value is clipped to 7FFF.
Also, even though the single ended mode reads voltages above ground ref, the conversion may result in -ve value as per this note in the datasheet:
"Single-ended signal measurements, where VAINN = 0 V and VAINP = 0 V to +FS, only use
the positive code range from 0000h to 7FFFh. However, because of device offset, the
ADS111x can still output negative codes in case VAINP is close to 0 V."

(I can vouch for that note 'cause I just spent an insane amount of time trying to figure out how I was getting occasional -ve values for a pin I had connected to GND 😑)

sanni-t:
^ Only changed _read_adc to uint16. Explanation in the discussion above about signed-ness of the converted value

modules/thermo-cycler/thermo-cycler-arduino/thermistorsadc.cpp
sfoster1: Oof. That’s kinda terrible! But it all makes sense now. I’ll resolve.

sfoster1 (approved):

PR #68, opened 2019-06-21
tests(thermocycler): add thermistor life test, NGS prep test

sanni-t:
## overview ##

**NGS Prep test**: A python script runs an NGS prep sequence of 29 temperature cycles while recording the device status to a csv file.

**Thermistor life test:** A python script runs the above NGS protocol continuously 250 times. This is to check how much the thermistors drift during its lifetime. This test will last ~10 days and will be repeated thrice.

The two tests are very similar but are saved separately since they could each have future changes that the other would not want.

## changes ##

1. Main change in the thermocycler firmware is the addition of temperature safety check and making the manual fan control mode a semi-manual one. Previously, when a user entered manual fan control mode, there were no checks being made or warnings given if the system was heating up beyond the safety limits. This led scenarios where a thermocycler that's in manual fan control mode would have the fan turned off or set to a very low value and then being issued a command to go to a value (usually lower than room temperature) that required a lot of heat to be displaced out of the device. But since the fan was set to an insufficient speed, and the code was not controlling the fan anymore, that heat would keep building inside the device, making it rise to levels high enough to damage the board. 

The semi manual mode allows the user to set the fan speed manually as long as it is sufficient to keep the system temperatures within the safety limit.

2. added a safety check that will deactivate the thermocycler if the system crosses safe temperature levels.
3. removed cover fan control since we are getting rid of those fans. They aren't needed as the cover doesn't heat up much.
4. added a continuous status print mode and reformatted the status parameters.

NOTE: the fan control thresholds and speed values will be revisited in [#3643](https://github.com/Opentrons/opentrons/issues/3643)

### DIFF
--- a/modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
+++ b/modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
@@ -243,6 +243,17 @@ void update_fans_from_state()
 
   if (millis() - last_checked > 100)
   {
+    if (!auto_fan)
+    { // Heatsink safety threshold overrides manual operation
+      if (temp_probes.heat_sink_temperature() > 55)
+      {
+        // Fan speed proportional to temperature
+        float pwr = temp_probes.heat_sink_temperature() / 100;
+        pwr = max(pwr, heatsink_fan.manual_power);
+        heatsink_fan.set_percentage(pwr);
+      }
+      return;
+    }
     last_checked = millis();
     if (master_set_a_target)
     {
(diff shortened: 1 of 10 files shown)


modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
```diff
@@ -242,6 +243,17 @@ void update_fans_from_state()
 
   if (millis() - last_checked > 100)
   {
+    if (!auto_fan)
+    { // Heatsink safety threshold overrides manual operation
+      if (temp_probes.heat_sink_temperature() > 55)
+      {
+        // Fan speed proportional to temperature
+        float pwr = temp_probes.heat_sink_temperature() / 100;
```
sfoster1: Worried about types here and on 283, maybe make this a 100.0? Also maybe have a HEATSINK_P_CONSTANT defined somewhere that can be 1.0 for now?

sfoster1 (approved):
Looks all functional to me, just a quibble about some types

modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
sfoster1: Yeah in general there's some random thermal cutoffs especially in the fan stuff that I'd love to see as #defines somewhere.

sanni-t: heat_sink_temperature() is a float so we're actually good here but yea, I see how it's easy to have a bad type conversion. I'll make the changes

PR #70, opened 2019-07-09
fix(thermocycler): add pause on bottom switch during lid open

sanni-t:
Previously, the thermocycler lid had a frequent issue of not being able to open properly. This was because in closed position, the lid was staying hooked onto the latch. When made to open, the lid was supposed to move down a little in order to free the latch so it could open, but this downward movement distance was different for different thermocyclers and also changed depending on the labware used and whether the thermocycer had a lid seal.

This PR changes the lid open movement so the lid moves down enough and precisely, by moving down until it hits the bottom switch, to let the solenoid open the latch easily. This eliminates the variability of downward distance and allows the lid to open more reliably.

### DIFF
--- a/modules/thermo-cycler/thermo-cycler-arduino/lid.cpp
+++ b/modules/thermo-cycler/thermo-cycler-arduino/lid.cpp
@@ -240,15 +240,20 @@ bool Lid::open_cover()
      return true;
   }
   motor_on();
-  if(move_angle(-5))  // move down a bit
+  bool res;
+  if(move_angle(LID_OPEN_SWITCH_PROBE_ANGLE))  // move down to reach bottom switch
   {
     // lid hit bottom switch
     solenoid_on();
     delay(400);
     move_angle(10);     // move up a bit
     solenoid_off();
+    res = move_angle(LID_MOTOR_RANGE_DEG);
   }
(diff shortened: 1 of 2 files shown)


modules/thermo-cycler/thermo-cycler-arduino/lid.cpp
```diff
@@ -246,11 +240,14 @@ bool Lid::open_cover()
      return true;
   }
   motor_on();
-  move_angle(-1);    // move down a bit
-  solenoid_on();
-  delay(100);
-  move_angle(5);     // move up a bit
-  solenoid_off();
+  if(move_angle(-5))  // move down a bit
```
sfoster1: Just to be absolutely sure, should we move `-LID_MOTOR_RANGE_DEG` here?

sfoster1 (approved):
Looks fine, one small comment

modules/thermo-cycler/thermo-cycler-arduino/lid.cpp
sanni-t: Good call. Hardware also requested to abort lid open if bottom switch is not pressed after that step. I'll make those changes

PR #71, opened 2019-07-09
perf(thermocycler): use timer interrupt for thermistor read and PID compute

### LINKED ISSUE #3610: Thermocycler: firmware: switch to timer interrupt for PID computation
## issue ##
Right now the actual delta T in PID computation varies, even though we compute with a constant value. This is because the loop, esp thermistor reads, takes a very long time. 
This varying value causes changes in temperature profile of the thermistor everytime anything in the firmware is changed 

## proposed solution ##
- Make the loop shorter by making thermistor reads faster and split over multiple loops
- use a timer interrupt for PID compute

sanni-t:
## overview ##
This PR adds a 25ms timer interrupt that updates a pair of thermistors on each interrupt. The loop cycles through the next pair on each interrupt and the PID corresponding to that pair is computed right
after. This ensures that the compute occurs every 100ms.

##changes ##
- changed PID library to add a sampling mode selection: 
      auto- uses the inbuilt sampling rate to perform PID compute()
      manual- compute() needs to be called by the main program
- peltier PIDs changed to manual sampling mode (lid PID remains on auto)
- added a 25ms timer interrupt
- added the thermistor read & PID compute flow using timer interrupts (described above)
- switched ADC to 250 SPS (a compromise between high sampling rate and low noise susceptibility)
- removed all non gcode code

Closes [#3610](https://github.com/Opentrons/opentrons/issues/3610)


sfoster1 (approved):
Looks like it'll work and definitely good to have it on a fixed timescale!

sanni-t:
Merge is blocked pending review from Hardware

andy-g-calderon (approved):
Tested this with Sanniti earlier, New PID terms and behavior look great.

PR #73, opened 2019-07-26
feat(arduino-board-files): change samd bootloader and bump board version

### LINKED ISSUE #3773: Thermocycler: firmware: Change bootloader to modify fan pin levels during fw update
## overview

Currently the fan turns on while uploading firmware because the bootloader has those pins pulled high. Change these pin definitions.
Also check if any other pin statuses need to be changed.
This might involve switching to a completely new bootloader since the source code of the current bootloader doesn't seem to be available.

sanni-t:
In order to have the fans not turn on upon system reset or when updating firmware, the fan pin
assignment needed to be changed in the bootloader. 
This bootloader also assigns the thermocycler front button LED as an activity indicator while in bootloader mode and disables the peltiers (as a safety measure).

The PR for updated bootloader source code is referenced.

Closes [#3773](https://github.com/Opentrons/opentrons/issues/3773)


sfoster1 (approved):
LGTM

sanni-t:
Just realized I didn't change the VID number in board definitions

PR #75, opened 2019-07-31
feat(thermocycler): detect and report thermistor failure

### LINKED ISSUE #3772: Thermocycler: firmware: detect electrical failures and deactivate w/ serial message
## overview

Compliance requires the device to shut off/deactivate if a component failure is detected. For example, if a thermistor connection is broken. It would be a good UX to send a serial message once a failure is detected

sanni-t:
## overview

Closes [#3772](https://github.com/Opentrons/opentrons/issues/3772)
This PR adds a safety feature that would allow the module to deactivate if any of the thermistors gets disconnected since it would otherwise cause the PID computation to go haywire (This is also a compliance requirement) 
It also sends error warning through serial. The warning will be repeated every 2 seconds until the thermistor connection is repaired and system rebooted.

### DIFF
--- a/modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
+++ b/modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
@@ -342,6 +342,8 @@ void debug_status_prints()
   gcode.add_debug_response("Fan auto?", auto_fan);
   // Cover temperature:
   gcode.add_debug_response("T.Lid", temp_probes.cover_temperature());
+  // Thermistor status:
+  gcode.add_debug_response("T_error", int(thermistor_failure));
   // Motor status:
 #if HW_VERSION >= 3
   gcode.add_debug_response("Motor_fault", int(lid.is_driver_faulted()));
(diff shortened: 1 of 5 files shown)


modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
```diff
@@ -824,15 +826,28 @@ void therm_pid_peltier_update()
 void loop()
 {
   timeStamp = micros();
-  temp_safety_check();
   therm_pid_peltier_update();
+  if (temp_probes.detected_invalid_val)
+  {
+    if (millis() - last_error_print > ERROR_PRINT_INTERVAL)
```
sfoster1, line 832: If this triggers in or around `millis` overflow time at ~50 days, you're gonna see either
- this firing every single iteration through `loop()` based on underflow because `last_error_print` is near `UNSIGNED_LONG_MAX` and `millis()` is returning a small value
- or it just won't trigger until another 50 days pass

We probably want to do something more like
```c++
unsigned long now = millis();
if (now < last_error_print) {
  last_error_print = now;
}
if (UNSIGNED_LONG_MAX - last_error_print < ERROR_PRINT_INTERVAL) {
   last_error_print = UNSIGNED_LONG_MAX - ERROR_PRINT_INTERVAL;
}
if (now > (last_error_print + ERROR_PRINT_INTERVAL)) {
   gcode.response....
}
```

sfoster1: (`UNSIGNED_LONG_MAX` might not be the name of the constant but it or something like it will be in `<limits>` or `<stdint.h>` or something)

sfoster1 (changes requested):
I think this has some behavior around `millis()` wraparound that we wouldn't like, but other than that looks good. I assume the thermistor error bounds are based on testing?

sfoster1 (approved):

PR #76, opened 2019-08-05
 fix(tempdeck): fix model v4 fan control, LED pins & gcode args check

### LINKED ISSUE #72: Fix tempdeck overheating issue
## overview

A customer called in about their tempdeck heating when supposed to cool and does not cool down soon.

## current behavior

When a v1.2 tempdeck is set to 15C or if it is set to a temperature > 14C right after a higher temperature then the tempdeck goes to the target temperature, stays there for about 2-3mins and then starts heating. If the this temperature is allowed to rise upto 55C before deactivating then the temperature keeps rising until 120C (at which point the tempdeck would get damaged and stop working) or until power is shut off. This rise in temperature happens because when the tempdeck is deactivated but is > 55C, the module actively tries to cool the surface by turning on the peltiers, which makes it worse since the fans still don't have enough power. 

## fix
1. In v1.2, fan low speed is too low to remove any heat from the tempdeck chamber. Increase this speed. What makes this not too straight forward is that different models use different fans (some are PWM some aren't). Make sure to set the same perceived speed for all models

2. Include 15C in the cold temperature zone (it isn't included in any zone right now which prevents the fan from turning ON)

sanni-t:
Closes #72 

This PR fixes the overheating problem caused because of reasons stated in #72 

## Changes
- Include corner temperatures in respective temp zones (TEMPERATURE_FAN_CUTOFF_COLD & TEMPERATURE_FAN_CUTOFF_HOT)
- Make v4 fan behavior same as v3 fans- i.e, use fan on/off instead of pwm. PWM-ing the fan causes unexpected behavior in the control circuitry that uses capacitors and an inductor which probably create some charge/discharge pattern that makes the fan turn off unexpectedly. The result is that the fan isn't ON long enough to push the heat out. 
- Latest v4 tempdecks have a pwm controlled fan which also happens to have a much faster acceleration hence the previous on/off timings were not suiting well for it. So increased the overall on/off period and decreased the power. Note that the code is changed a bit so that the same control works for all v3 and v4 fans and all fans behave similarly.
- Add a version check for proper LED pin assignment (the LED connections changed mid v3 production)
- fixed a gcode set_temperature bug where an `M104` without any temperature specified would count as a valid gcode and result in setting the target as either 0C or the previously set target.

### DIFF
--- a/modules/temp-deck/temp-deck-arduino/lights.cpp
+++ b/modules/temp-deck/temp-deck-arduino/lights.cpp
@@ -168,7 +168,17 @@ void Lights::startup_animation(int target_number, int transition_time) {
   set_color_bar(0, 0, 0, 1);
 }
 
-void Lights::setup_lights() {
+void Lights::setup_lights(bool is_blue_pin_5) {
+  if (is_blue_pin_5)
+  {
+    blue_led = 5;
+    red_led = 6;
+  }
+  else
+  {
+    blue_led = 6;
+    red_led = 5;
+  }
   pinMode(red_led, OUTPUT);
   pinMode(blue_led, OUTPUT);
   Wire.setClock(400000);
--- a/modules/temp-deck/temp-deck-arduino/lights.cpp
+++ b/modules/temp-deck/temp-deck-arduino/lights.cpp
@@ -7,6 +7,10 @@ Lights::Lights(){
   https://github.com/adafruit/Adafruit-PWM-Servo-Driver-Library
 */
   pwm = Adafruit_PWMServoDriver();
+
+  // Default led pins
+  blue_led = 6;
+  red_led = 5;
 }
 
 void Lights::set_pwm_pin(int pin, float val) {
--- a/modules/temp-deck/temp-deck-arduino/temp-deck-arduino.ino
+++ b/modules/temp-deck/temp-deck-arduino/temp-deck-arduino.ino
@@ -26,6 +26,10 @@
 #include "gcode.h"
 
 #define device_version "v2.0.0"
+#define MODEL_VER_TEMPLATE "temp_deck_v"
+#define MODEL_VER_TEMPLATE_LEN sizeof(MODEL_VER_TEMPLATE) - 1
+#define SERIAL_VER_TEMPLATE "TDV03P2018"
+#define SERIAL_VER_TEMPLATE_LEN sizeof(SERIAL_VER_TEMPLATE) - 1
 
 #define PIN_BUZZER 11  // a piezo buzzer we can use tone() with
 #define PIN_FAN 9      // blower-fan controlled by simple PWM analogWrite()
@@ -532,9 +536,7 @@ void setup() {
 
   memory.read_serial(device_serial);
   memory.read_model(device_model);
-
-  const String model_ver_template= "temp_deck_v";
-  String ver = device_model.substring(model_ver_template.length());
+  String ver = device_model.substring(MODEL_VER_TEMPLATE_LEN);
   model_version = ver.toInt();
 
   if (model_version == 3 || model_version == 4)
@@ -543,12 +545,12 @@ void setup() {
     if (model_version == 3)
     {
       // V3 tempdecks produced after Oct 15 have different LED pin configuration
-      const String serial_ver_template = "TDV03P2018"; // example serial-TDV03P20181008A01
-      const uint8_t date_length = 4;
+      // serial number has the production date. eg. TDV03P*20181008*A01
+      const uint8_t date_length = 4;  // MMDD
       String serial_date = device_serial.substring(
-        serial_ver_template.length(), serial_ver_template.length() + date_length);
+        SERIAL_VER_TEMPLATE_LEN, SERIAL_VER_TEMPLATE_LEN + date_length);
       int v3_date = serial_date.toInt();
-      if (v3_date > 1019)
+      if (v3_date > 1015)
       {
         is_blue_pin_5 = true;
       }
--- a/modules/temp-deck/temp-deck-arduino/temp-deck-arduino.ino
+++ b/modules/temp-deck/temp-deck-arduino/temp-deck-arduino.ino
@@ -71,8 +71,8 @@ unsigned long fan_timestamp = 0;
 bool is_fan_on = false;
 bool is_v3_v4_fan = false;
 
-// LED pins for model versions 3 & 4: red = 6, blue = 5
-// LED pins for model versions < 3  : red = 5, blue = 6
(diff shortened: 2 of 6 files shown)


modules/temp-deck/temp-deck-arduino/lights.cpp
```diff
@@ -168,7 +168,17 @@ void Lights::startup_animation(int target_number, int transition_time) {
   set_color_bar(0, 0, 0, 1);
 }
 
-void Lights::setup_lights() {
+void Lights::setup_lights(bool is_blue_pin_5) {
```
sfoster1, line 175: arg name doesn't match the declaration, minor but still let's fix it

sfoster1: Also, a bit more major - let's either put this logic in the constructor or set some reasonable defaults there, otherwise these class variables are uninitialized. Also if they're in the ctor they can be declared `const`, and initialized like

```c++
Lights::Lights(bool blue_is_pin_5): blue_led(blue_is_pin_5 ? 5 : 6), red_led(blue_is_pin_5 ? 6 : 5) {}
```

Or just specify the pin numbers in the ctor and hoist that logic to the callsite.

modules/temp-deck/temp-deck-arduino/temp-deck-arduino.ino
```diff
@@ -25,7 +25,11 @@
 #include "thermistor.h"
 #include "gcode.h"
 
-#define device_version "v1.2.0"
+#define device_version "v2.0.0"
```
sfoster1: Are there hardware changes or is this for the software update?

modules/temp-deck/temp-deck-arduino/temp-deck-arduino.ino
```diff
@@ -530,11 +530,31 @@ void setup() {
...
   memory.read_model(device_model);
-
-  if (device_model.indexOf("v3.0") > 0) is_v3_0_fan = true;
-  else if (device_model.indexOf("v4.0") > 0) is_v4_0_fan = true;
-
-  lights.setup_lights();
+  String ver = device_model.substring(MODEL_VER_TEMPLATE_LEN);
+  model_version = ver.toInt();
+
+  if (model_version == 3 || model_version == 4)
```
sfoster1, line 425: Maybe we should have an `else` here for if something corrupted this?

modules/temp-deck/temp-deck-arduino/temp-deck-arduino.ino
```diff
@@ -55,21 +59,21 @@ float _offset_temp_diff = 0.0;
...
 bool is_fan_on = false;
-bool is_v3_0_fan = false;
+bool is_v3_v4_fan = false;
 
-// model version 3.1 has same fans as v3.0, but pcb uses mosfet to PWM fan power
-#define FAN_V4_0_LOW 0.1
-bool is_v4_0_fan = false;
+// LED pins for model versions 3 (post 2018.10.15) & 4: red = 6, blue = 5
+// LED pins for model versions < 3 & 3.0 (pre- 2018.10.15) : red = 5, blue = 6
+bool is_blue_pin_5 = false;
```
sfoster1: It looks like `is_blue_pin_5` is only used in `setup`, it can be a local variable there

modules/temp-deck/temp-deck-arduino/temp-deck-arduino.ino
```diff
@@ -134,6 +138,7 @@ PID myPID(&CURRENT_TEMPERATURE, &TEMPERATURE_SWING, &TARGET_TEMPERATURE, DOWN_PI
 
 String device_serial = "";  // leave empty, this value is read from eeprom during setup()
 String device_model = "";   // leave empty, this value is read from eeprom during setup()
+int model_version;          // value is read from device_model during setup()
```
sfoster1: It looks like `model_version` is only used in `setup`, it can be a local variable there

modules/temp-deck/temp-deck-arduino/temp-deck-arduino.ino
```diff
@@ -134,6 +138,7 @@ PID myPID(&CURRENT_TEMPERATURE, &TEMPERATURE_SWING, &TARGET_TEMPERATURE, DOWN_PI
 
 String device_serial = "";  // leave empty, this value is read from eeprom during setup()
```
sfoster1: We should consider making these local to `setup` if possible, not sure how much I like this combination of the free store and the heap

sfoster1 (changes requested):
The fixes look good and I'll test them, but a lot of style quibbles (acknowledging that a lot of it is just previous code).

One question though - What other versions of the hardware are out there? What makes the v4 fan different from the v1 and v2 fan? Can we just drop the `is_v3_v4_fan` thing and manually control _all_ of them?

modules/temp-deck/temp-deck-arduino/temp-deck-arduino.ino
sanni-t: So, the hardware change was actually made 2 versions ago (v1.1.0 was the firmware version for the tempdecks with changed fan) but the device version was only bumped from v1.0.1 to v1.1.0. It probably should have been bumped to v2 instead. So I have kinda retroactively bumped it now. What do you think?

sanni-t:
Tempdeck models v3 and v4 have the same fan but have some changes in the circuitry and components. Models v3 and v4 have fans that are different from previous models in that they are not PWM controllable (unlike previous models which *are* pwm controlled). v3 & v4 are instead pulsed ON-OFF at regular intervals (the fan is on for a couple of seconds and off for a couple of seconds)

modules/temp-deck/temp-deck-arduino/lights.cpp
sanni-t: We can't add the logic to ctor since we won't know whether `blue_is_pin_5` or not until we check the tempdeck's version # in setup().
I'll assign them default values in the ctor and then update them in setup()

modules/temp-deck/temp-deck-arduino/temp-deck-arduino.ino
sanni-t: We use `device_serial` & `device_model` in device_info gcode response as well. And it's very much possible we might use it in other contexts in the future esp with the Gen2 tempdecks if they have any changes that might need us to add conditions based on the device_model. So, I think keeping them global makes sense

sanni-t:
^ above changes are for latest v4 tempdecks (Aug+ production) that actually have PWM controlled fans that have a much higher ramp-to-speed rate than the older, non-pwm v4 fans.
Updated PR description

sfoster1 (approved):
code lgtm

PR #78, opened 2019-08-16
feat(thermocycler): enable lid operation using optical switch

### LINKED ISSUE #3876: Thermocycler: firmware: make lid operation changes for optical switch
## overview

- make the change for using optical switch as bottom switch.
- make the movement change for new latch design

sanni-t:
Closes [#3876](https://github.com/Opentrons/opentrons/issues/3876)

The DVT units use an optical switch as the bottom switch rather than a microswitch that was used in EVTs. 

* NOTE: To use, make HW_VERSION=4 in preferences.local.txt file *


modules/thermo-cycler/thermo-cycler-arduino/lid.cpp
```diff
@@ -241,19 +247,17 @@ bool Lid::open_cover()
   }
   motor_on();
   bool res;
-  if(move_angle(LID_OPEN_SWITCH_PROBE_ANGLE))  // move down to reach bottom switch
+  // move down a bit to release latch
+  unsigned long push_steps = abs(LID_OPEN_DOWN_MOTION_ANGLE) * STEPS_PER_ANGLE;
```
sfoster1: Why do this manually instead of `move_angle`?

modules/thermo-cycler/thermo-cycler-arduino/lid.cpp
```diff
@@ -241,19 +247,17 @@ bool Lid::open_cover()
...
-    res = move_angle(LID_MOTOR_RANGE_DEG);
-  }
-  else
-  {
-    res = false;
+    _motor_step(DIRECTION_DOWN);
   }
+  solenoid_on();
+  delay(400);
+  move_angle(10);     // move up a bit
```
sfoster1: Let's make the comments a little more about the why rather than the what - why are we moving up a bit? to clear the solenoid?

modules/thermo-cycler/thermo-cycler-arduino/lid.cpp
```diff
@@ -266,9 +270,17 @@ bool Lid::close_cover()
   }
   motor_on();
   bool res = move_angle(-LID_MOTOR_RANGE_DEG);
-  delay(500);
   if (res)
   {
+#if HW_VERSION >=4
+    // move down a bit
```
sfoster1: Same thing about why rather than what - is this to make sure we hit the optical switch?

modules/thermo-cycler/thermo-cycler-arduino/lid.cpp
```diff
@@ -336,9 +348,15 @@ bool Lid::setup()
   pinMode(PIN_COVER_SWITCH, INPUT);
   pinMode(PIN_BOTTOM_SWITCH, INPUT);
 #endif
-  // Both switches are NORMALLY CLOSED
+  // Cover switch is NORMALLY CLOSED
```
sfoster1: Rather than duplicating the #ifs, can we put this in an `inline` function?

modules/thermo-cycler/thermo-cycler-arduino/lid.h
```diff
@@ -74,7 +74,9 @@
 #define PULSE_HIGH_MICROSECONDS 2
 #define MOTOR_STEP_DELAY 60   // microseconds
 #define LID_CLOSE_BACKTRACK_ANGLE 3.0
```
sfoster1: Maybe some comments about what these angles are for and when they're used?

modules/thermo-cycler/thermo-cycler-arduino/lid.cpp
```diff
@@ -266,9 +270,17 @@ bool Lid::close_cover()
...
   {
+#if HW_VERSION >=4
+    // move down a bit
+    unsigned long final_steps = abs(LID_CLOSE_LAST_STEP_ANGLE) * STEPS_PER_ANGLE;
+    for (unsigned long i=0;i<final_steps;i++)
+    {
+      _motor_step(DIRECTION_DOWN);
+    }
+#endif
+  delay(500);
```
sfoster1: What are we waiting for here?

sfoster1 (changes requested):
It seems like it'll all work but I think we should have some better comments around why we're doing things, because whenever we release this we probably won't look at it again for like 6 months and forget stuff

modules/thermo-cycler/thermo-cycler-arduino/lid.cpp
sanni-t: Because move angle exits when a switch press is detected. With the optical switch, we need to move down a bit more while the switch is 'ON' to let the latch actually lock.

modules/thermo-cycler/thermo-cycler-arduino/lid.h
sanni-t: Pushed new commit and added the comments in the open & close functions since explaining what the angles are will need explanation of the whole open & close process

modules/thermo-cycler/thermo-cycler-arduino/lid.cpp
sanni-t: Changed `move_angle` a bit so that it can be used here

modules/thermo-cycler/thermo-cycler-arduino/lid.cpp
```diff
@@ -300,6 +331,17 @@ void _motor_fault_callback()
 #endif
 }
 
+inline bool Lid::_bottom_switch_check()
```
sfoster1, line 334: If this is defined as inline you may need to put the implementation in the header, I forget

sfoster1 (approved):
LGTM if tests pass and it works!

modules/thermo-cycler/thermo-cycler-arduino/lid.cpp
sanni-t: Oh! This doesn't give any errors but you are probably right. I merged the PR before noticing this comment. I'll put up a PR with the correction

PR #79, opened 2019-08-19
feat(thermocycler): open/close lid on button press. Do not deactivate

### LINKED ISSUE #3808: Thermocycler: firmware: Open/ Close lid on button 'toggle'
## overview

Use the front button to open the lid if it is closed and close lid if it's open. If it's in between, it will open.
Do not deactivate.

sanni-t:
Closes [#3808](https://github.com/Opentrons/opentrons/issues/3808)

This PR enables the front button on thermocycler to open the lid when the lid is closed and to close it when it is open.
This operation will not deactivate the thermocycler.

### DIFF
--- a/modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
+++ b/modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
@@ -576,6 +576,7 @@ void front_button_callback()
 
 bool is_front_button_pressed()
 {
+#if HW_VERSION >=3
   if (front_button_pressed && millis() - front_button_pressed_at >= 200)
   {
     if(digitalRead(PIN_FRONT_BUTTON_SW) == LOW)
@@ -584,6 +585,7 @@ bool is_front_button_pressed()
       return true;
     }
   }
+#endif
   return false;
 }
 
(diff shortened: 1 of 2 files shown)


modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
```diff
@@ -862,8 +862,20 @@ void loop()
 #if HW_VERSION >=3
```
sfoster1: If this only works in 4 shouldn't this be `HW_VERSION > 3`

sanni-t: Sorry, I was kinda wrong. The V3 thermocycler lid is also able to alternately close/open because the condition for opening is that the lid should be either 'closed' or 'in_between'. Because of the 'faulty' switch position, in V3 thermocyclers, the lid will read 'in_between' when closed so it will still execute `open_cover()`

sanni-t: I just moved the `#if` to  `is_front_button_pressed()` since it could be misleading here (Also, it really should have been in that function anyway)

sfoster1 (approved):
LGTM

PR #80, opened 2019-08-21
fix(thermocycler): fix DVT lid jump-back & latch opening issues

sanni-t:
## overview 
LID_CLOSE_LAST_STEP_ANGLE was larger than the actual travel space available to the lid
which was causing the lid to jump back upon hitting the end; so, decreased this distance. 

This jump-back was also happening occasionally (on some units) while opening the lid, which was causing the latch to not open on time. Hence the change to engage the solenoid before starting to move the lid.

## changes
- changes described above to fix the jump-back
- added motor speed control (`MOTOR_RPM`)
- added lid open check & reopen upon failure to open
- updated lid test to include the actual lid status since we can now detect when it is closed


sfoster1 (approved):
If it works it lgtm

PR #81, opened 2019-08-29
fix(thermocycler): fix lid not closing without labware

sanni-t:
Some DVT thermocyclers were unable to close their lids when there was no labware in them. This was caused due to gear backlash. 

This PR adds a check after normal closing movement and if the lid isn't closed, it'll try to close it again. Detailed explanation of the fix is in the code.


TaylorGoodman (commented):
new code functioned great.  Nice work!

sfoster1 (approved):
LGTM

PR #82, opened 2019-09-05
feat(thermocycler, arduino-board-files): add serial number storage capability

### LINKED ISSUE #3492: Thermocycler: save device model and serial in "eeprom"
## overview

The ATSAMD21 doesn't have a dedicated eeprom space like the atmega32. We might have to assign some amount of program memory space as eeprom storage and prevent the bootloader from writing over it so we can store the thermocycler's model & serial numbers.

(Will flush this out better as I start working on it)

### LINKED ISSUE #4007: Thermocycler: firmware: add eepprom writer
## overview
Make an EEPROM writer firmware that takes a serial number from barcode scanner as input and writes to EEPROM.

sanni-t:
Closes [#3942](https://github.com/Opentrons/opentrons/issues/3492)
Closes [#4007](https://github.com/Opentrons/opentrons/issues/4007)
Closes [#3941](https://github.com/Opentrons/opentrons/issues/3941)

## overview
This PR adds serial and model number read & write feature. Uses an I2C EEPROM to store the two numbers ([eeprom datasheet](http://ww1.microchip.com/downloads/en/DeviceDoc/AT24C01C-AT24C02C-I2C-Compatible-Two-Wire-Serial-EEPROM-1Kbit-2Kbit-20006111A.pdf))

## changes
1. Added eepromWriter firmware to write serial & model numbers to eeprom (barcode scanner compatible)
[ NOTE: After discussion with @andy-g-calderon @TaylorGoodman & @yunhodavidkim it was decided that model number will be a part of serial number (TC*V01*..) and each hardware change will result in a version bump. So, DVTs are V01 and PVTs, since they have a h/w change, would be V02.]
2. Made the main firmware now read serial and model numbers from eeprom instead of having dummy values
3. Added production script to automate the process of writing serial number & uploading application firmware
4. Updated flags in makefile
5. Added a shared data header file to thermocycler board defs, bumped up the arduino board def version


modules/thermo-cycler/production/eepromWriter/eepromWriter.ino
```diff
@@ -0,0 +1,215 @@
...
+    }
+  }
+  return '~';
+}
+
+bool write_to_eeprom(String serial_num)
+{
+  char the_number[MAX_IN_NUM_LEN];
+  uint8_t number_len;
+  serial_num.toCharArray(the_number, serial_num.length()+1);
```
sfoster1:
```suggestion
  serial_num.toCharArray(the_number, MAX_IN_NUM_LEN);
```

Per [the arduino docs](https://www.arduino.cc/reference/en/language/variables/data-types/string/functions/tochararray/) that should be the size of the dest buffer.

modules/thermo-cycler/production/eepromWriter/eepromWriter.ino
```diff
@@ -0,0 +1,215 @@
...
+      Serial.println("\nfailed");
+      return 0;
+    }
+  }
+  /********** Update Model no. **********/
+  erase_eeprom_section(MODEL_LOC);
+  Serial.println("\nupdating model number..");
+  char model_num[VER_NUM_LEN+2];
+  model_num[0] = 'v';
+  model_num[VER_NUM_LEN-1] = '\0';
```
sfoster1:
```suggestion
  model_num[VER_NUM_LEN+1] = '\0';
```

this is just nulling out the last byte right? shouldn't this be the last element of `model_num`? Could we instead do `char model_num[VER_NUM_LEN+2] = {'v', 0, 0, 0};` and then assign into the middle

modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
```diff
@@ -120,7 +120,7 @@ bool is_target_hot()
 
 bool is_target_cold()
 {
-  return target_temperature_plate < TEMPERATURE_ROOM;
+  return target_temperature_plate <= TEMPERATURE_ROOM;
```
sfoster1: This should not be in this PR, it will confuse us if we have to browse the history to see when stuff changed.

modules/thermo-cycler/thermo-cycler-arduino/eeprom.cpp
```diff
@@ -0,0 +1,59 @@
...
+  {
+    the_number += _read_char(addr++);
+    if (the_number.length() > MAX_IN_NUM_LEN+1)
+    { // Either the serial/model didn't finish writing correctly/ is invalid
+      // or there is no serial/model stored
+      return "~";
+    }
+  }
+  return the_number;
+#endif
```
sfoster1, line 49: If `HW_VERSION` is less than 4, this will preprocess to

```c++
String Eeprom::read(MemOption option) {}
```

which is either a compiler error (it better be) or happily-compiled undefined behavior because it doesn't return anything. We should return `""` or something if that #if doesn't evaluate.

modules/thermo-cycler/thermo-cycler-arduino/eeprom.cpp
```diff
@@ -0,0 +1,59 @@
...
+  }
+  else if (option == MemOption::model)
+  {
+    addr = MODEL_LOC;
+  }
+  else
+  {
+    return "\0";
+  }
+  String the_number;
```
sfoster1: Can we do `the_number.reserve(MAX_IN_NUM_LEN)` so we don't expand it every time we append a character?

modules/thermo-cycler/thermo-cycler-arduino/eeprom.cpp
```diff
@@ -0,0 +1,59 @@
...
+  while (_read_char(addr) != '\0')
```
sfoster1: Looks like we're reading everything twice this way, can we just read up to the max for the specific case (serial vs model) and send it upstream to deal with? maybe detect if we only read 0xff?

modules/thermo-cycler/thermo-cycler-arduino/eeprom.cpp
```diff
@@ -0,0 +1,59 @@
...
+  {
+    Wire.requestFrom(EEPROM_ADDR, 1); // request 1 byte from eeprom
+    if (Wire.available())    // slave may send less than requested
+    {
+      char c = Wire.read(); // receive a byte as character
+      return c;         // print the character
+    }
+  }
+  return '~';
+#endif
```
sfoster1, line 66: This similarly needs to return something if the #if isn't true otherwise it's undefined behavior

modules/thermo-cycler/thermo-cycler-arduino/eeprom.cpp
```diff
@@ -0,0 +1,59 @@
...
+  Wire.beginTransmission(EEPROM_ADDR);
+  Wire.write(word_address);
+  uint8_t error = Wire.endTransmission();
+  if (!error)
+  {
+    Wire.requestFrom(EEPROM_ADDR, 1); // request 1 byte from eeprom
+    if (Wire.available())    // slave may send less than requested
+    {
+      char c = Wire.read(); // receive a byte as character
+      return c;         // print the character
```
sfoster1: `return Wire.read();`?

modules/thermo-cycler/thermo-cycler-arduino/eeprom.h
```diff
@@ -0,0 +1,32 @@
...
+#define eeprom_h
+
+#include "Arduino.h"
+#include <Wire.h>
+
+#define EEPROM_ADDR     0x52  // TODO: make this 0x52 for thermocycler
+#define MAX_IN_NUM_LEN  17  // TCV0120190903Annn
+#define BUFFER_BYTES    10
+#define SERIAL_LOC      10
+#define MODEL_LOC       SERIAL_LOC + MAX_IN_NUM_LEN + BUFFER_BYTES
```
sfoster1: We should define the `MODEL_LOC` as an absolute address rather than as a derivation of the `SERIAL_LOC`, for a couple reasons:

1. If we define it as an absolute location, you can see what it is at a glance by looking at this file
2. When we define it as a derived location, we can accidentally change it by changing one of the priors, and break back compatibility for reading serials. 

In fact, I'd recommend something like this:

```c++
#define SERIAL_LOC 10
#define MAX_IN_NUM_LEN 17
#define MODEL_LOC  30

static_assert(MODEL_LOC + MAX_IN_NUM_LEN < SERIAL_LOC);
```

so you'll get a compiler error if it fails. And this should ideally be in a header file shared with the eeprom writer class.

modules/thermo-cycler/production/firmware/readme.md
```diff
@@ -0,0 +1 @@
+# Contains the latest released firmware binary
```
sfoster1, line 1: Why not use github releases for this?

modules/thermo-cycler/production/eepromWriter/eepromWriter.ino
```diff
@@ -0,0 +1,215 @@
+#include <Wire.h>
+
+#define EEPROM_ADDR       0x52
+
+#define MAX_IN_NUM_LEN    17  // max serial = 17 (TCV01yyyymmddAnnn) | max model = 3 //vnn
```
sfoster1: See the similar comment on eeprom.h - this stuff should be in a shared data file or something

sfoster1 (changes requested):

sfoster1 (changes requested):
There's a couple individual issues (in the inline comments), and some larger structural ones:

- If we're going to run this from a python script on the host computer anyway, let's pull as much functionality out of the eeprom writer as we can. Let's pull the model number and do string parsing on the Python side since it's a lot less painful, let's make the serial interface a bit less interactive, that sort of thing.
- If we're going to keep having a separate eeprom writer and normal firmware, we should put the eeprom layout definitions in a shared header file so there's a single source of truth.
- EEPROM sector layouts should be absolute, so they're easier to understand at a glance; and we should probably have some static asserts to make sure they don't step on each other.
- We should consider making the driving script loop rather than one-shot: have it stay resident and wait for another thermocycler to be connected on its own, so the factory doesn't have to start it over and over
- IMO (and feel free to disagree) we shouldn't rely on null termination for eeproms, we should have fixed-length formats (where if the data is shorter than that it can be nul but doesn't have to be).

modules/thermo-cycler/production/eepromWriter/eepromWriter.ino
```diff
@@ -0,0 +1,188 @@
...
+  }
+  return '~';
+}
+
+bool write_to_eeprom()
+{
+  uint8_t serial_len = serial_num.length();
+  uint8_t model_len = model_num.length();
+  char serial_arr[serial_len];
+  serial_num.toCharArray(serial_arr, serial_len+1);
```
sfoster1: Why `serial_len+1`? won't that overrun the buffer? if you want room for the null byte make the buffer bigger too

modules/thermo-cycler/thermo-cycler-arduino/eeprom.cpp
sfoster1: still same issue after this commit

sfoster1 (changes requested):
Couple more little changes but I like the shared data!

sfoster1 (approved):
LGTM!

PR #83, opened 2019-09-09
fix(thermocycler): room temp not included in either hot or cold

sanni-t:
This PR fixes a bug where room temperature of 23C wasn't being included in either hot or cold temp range, which would make the fans & LEDs behave incorrectly


sfoster1 (approved):
LGTM

PR #85, opened 2019-09-20
fix(thermocycler): decrease light strip update delays

sanni-t:
## overview
The light strip was taking about 15milliseconds to update. This is not desired as it blocks the loop
for that long.

## changes
- reduced per LED update delay to 10us
- bumped the FW version & uploaded updated binary to /production
- added a firmware uploader script to make it more reliable to upload on thermocyclers with uf2 bootloader (this bootloader doesn't play well with Arduino)

## review requests
1. Check for code sanity
2. Test on a thermocycler to check that the perceived light behavior (like, blinking speed) hasn't changed significantly


sfoster1 (approved):
LGTM

b-cooper (approved):
tested script on TC with new bootloader 👢

PR #86, opened 2019-09-24
feat(thermocycler): edit file path and file types in firmware uploaders

sanni-t:
## overview
SZ team needs to use the `firmware_uploader` to upload the thermocycler firmware to do preliminary tests on each unit's unassembled boards. Hence, making the script use the production firmware as default rather than the dev firmware.

## changes
- added an arg to provide custom file path
- the script now allows usage of both .bin and .uf2 files. This is being added for future use since we will most likely be building firmware artifacts with .uf2 files instead of .bin ones

### DIFF
--- a/modules/thermo-cycler/production/firmware_uploader.py
+++ b/modules/thermo-cycler/production/firmware_uploader.py
@@ -16,14 +16,16 @@ from serial.tools.list_ports import comports
 from argparse import ArgumentParser
 
 THIS_DIR = os.path.dirname(os.path.realpath(__file__))
-DEFUALT_FW_FILE_PATH = os.path.join(THIS_DIR, "firmware", "thermo-cycler-arduino.ino.bin")
+DEFAULT_FW_FILE_PATH = os.path.join(THIS_DIR, "firmware", "thermo-cycler-arduino.ino.bin")
 OPENTRONS_VID = 1240
 MAX_SERIAL_LEN = 16
 
 def build_arg_parser():
     arg_parser = ArgumentParser(
         description="Thermocycler firmware uploader")
-    arg_parser.add_argument("-F", "--fw_file", required=False)
+    arg_parser.add_argument("-F", "--fw_file", required=False,
+                            default=DEFAULT_FW_FILE_PATH,
+                            help='Firmware file (default: ..production/firmware/thermo-cycler-arduino.ino.bin)')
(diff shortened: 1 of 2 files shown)


modules/thermo-cycler/production/firmware_uploader.py
```diff
@@ -1,20 +1,31 @@
...
-FIRMWARE_FILE_PATH = "../../../build/tmp/thermo-cycler-arduino.ino.bin"
+THIS_DIR = os.path.dirname(os.path.realpath(__file__))
+DEFUALT_FW_FILE_PATH = os.path.join(THIS_DIR, "firmware", "thermo-cycler-arduino.ino.bin")
 OPENTRONS_VID = 1240
 MAX_SERIAL_LEN = 16
 
+def build_arg_parser():
+    arg_parser = ArgumentParser(
+        description="Thermocycler firmware uploader")
+    arg_parser.add_argument("-F", "--fw_file", required=False)
```
sfoster1: if you put the default above into here as `default=DEFAULT_FW_FILE_PATH` it will display when you do `--help` but this works fine too

modules/thermo-cycler/production/firmware_uploader.py
```diff
@@ -1,20 +1,31 @@
...
+import os
 import uf2conv
 import serial
 import time
 from serial.tools.list_ports import comports
+from argparse import ArgumentParser
 
-FIRMWARE_FILE_PATH = "../../../build/tmp/thermo-cycler-arduino.ino.bin"
+THIS_DIR = os.path.dirname(os.path.realpath(__file__))
+DEFUALT_FW_FILE_PATH = os.path.join(THIS_DIR, "firmware", "thermo-cycler-arduino.ino.bin")
```
sfoster1: typo in the name?

sfoster1 (approved):
Couple small suggestions but it looks like it will work

modules/thermo-cycler/production/firmware_uploader.py
sanni-t: oh! cool. I'll make that change

modules/thermo-cycler/production/firmware_uploader.py
sanni-t: oops

PR #88, opened 2019-09-30
fix(thermocycler): make fw upload scripts windows compatible

sanni-t:
Addresses [#4113](https://github.com/Opentrons/opentrons/issues/4113)

## overview

The uf2conv.py script had a bug where the object containing all drives info was not decoded before doing string manipulation. Also, on windows, drives/ volumes aren't referred to with names but by assigned drive letter (D:/ ..); so instead of looking for a volume called 'TCBOOT', we have to verify whether it's a thermocycler bootloader by verifying INFO_UF2.TXT file contents.

Refer to the issue link above to read about why this change was necessary

### DIFF
--- a/modules/thermo-cycler/production/firmware_uploader.py
+++ b/modules/thermo-cycler/production/firmware_uploader.py
@@ -56,11 +56,13 @@ def find_bootloader_drive():
     # takes around 4 seconds for TCBOOT to register in /Volumes
     retries = 7
     while retries:
-        for d in uf2conv.get_drives():
-            if "TCBOOT" in d:
-                print("Bootloader Volume found")
-                return d
         print("Searching bootloader...")
+        for d in uf2conv.get_drives():
+            with open(os.path.join(d + '/INFO_UF2.TXT'), mode='r') as f:
+                drive_info = f.read()
+                if 'Opentrons Thermocycler M0' in drive_info:
+                    print("Bootloader Volume found")
+                    return d
         retries -= 1
         time.sleep(1)
     raise Exception ('Bootloader volume not found')
--- a/modules/thermo-cycler/production/firmware_uploader.py
+++ b/modules/thermo-cycler/production/firmware_uploader.py
@@ -57,12 +57,10 @@ def find_bootloader_drive():
     retries = 7
     while retries:
         print("Searching bootloader...")
-        for d in uf2conv.get_drives():
-            with open(os.path.join(d + '/INFO_UF2.TXT'), mode='r') as f:
-                drive_info = f.read()
-                if 'Opentrons Thermocycler M0' in drive_info:
-                    print("Bootloader Volume found")
-                    return d
+        for d in uf2conv.get_tc_drives():  # returns only valid TCBOOT volumes
+            # The first detected thermocycler volume is picked
+            print("Bootloader Volume found")
+            return d
         retries -= 1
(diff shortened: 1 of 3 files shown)


modules/thermo-cycler/production/firmware_uploader.py
```diff
@@ -56,11 +56,13 @@ def find_bootloader_drive():
     # takes around 4 seconds for TCBOOT to register in /Volumes
     retries = 7
     while retries:
-        for d in uf2conv.get_drives():
-            if "TCBOOT" in d:
-                print("Bootloader Volume found")
-                return d
         print("Searching bootloader...")
+        for d in uf2conv.get_drives():
+            with open(os.path.join(d + '/INFO_UF2.TXT'), mode='r') as f:
```
sfoster1: You're probably going to get a FileNotFound error here if the file doesn't exist in that root. Also, `os.path.join()` takes multiple arguments (like `os.path.join(d, 'INFO_UF2.TXT')`. You may want to consider using pathlib here.

sfoster1: Also maybe move this into `uf2conv`? it's used both here and in the serial and firmware uploader, so might as well centralize it

modules/thermo-cycler/production/serial_and_firmware_uploader.py
```diff
@@ -58,9 +59,11 @@ def find_bootloader_drive():
     retries = 7
     while retries:
         for d in uf2conv.get_drives():
-            if "TCBOOT" in d:
-                print("Bootloader Volume found")
-                return d
+            with open(os.path.join(d + '/INFO_UF2.TXT'), mode='r') as f:
```
sfoster1: Same comments as the version in firmware_uploader

modules/thermo-cycler/production/uf2conv.py
```diff
@@ -172,7 +172,7 @@ def get_drives():
     if sys.platform == "win32":
         r = subprocess.check_output(["wmic", "PATH", "Win32_LogicalDisk",
                                      "get", "DeviceID,", "VolumeName,",
-                                     "FileSystem,", "DriveType"])
+                                     "FileSystem,", "DriveType"]).decode()
```
sfoster1, line 175: Like 99% sure this is fine but there's a chance this will fail if someone has a drive with a volume name that isn't ascii (for instance, hanzi) and crash the script, depending on what decode decides is the system locale and what the volume label is encoded as. Might want to give it a try. If it's a problem, you can probably just change the subsequent code to operate on bytes rather than string.

sfoster1 (changes requested):
Couple inline changes.

modules/thermo-cycler/production/uf2conv.py
sanni-t: Tested by making a volume label on Windows to 'Win(你好)' and no error was thrown while parsing drives. The letters were decoded as 'Win(??)'.
There's no decoding on Mac; `os.listdir(rootpath):` (line 188) returns the chinese charaters of the volume label correctly with no errors or ??

sfoster1: 👍

sfoster1 (approved):

sanni-t:
No longer required since PR #89 removes uf2 bootloader

PR #89, opened 2019-10-08
feat(thermocycler, arduino-board-files): switch to SAM-BA bootloader- OT mod

### LINKED ISSUE #4097: Thermocycler: bootloader: load into application firmware on reset
## overview

Have the controller load into the application space by default, always. 

## current behavior

If the device was reset while in bootloader mode, it just resumes the bootloader instead of running the application firmware. This causes the device to stay inside the bootloader until a new firmware file is written to it.

## expected behavior

Have the device come out of bootloader and run the application sketch on reset

## Possible problems in making this happen

If loading bootloader either erases the application space or somehow tampers with it, then there is no way to reset into the application.

## Notes
Brian- "the bootloader doesn’t seem to hotplug either
if my robot puts it in bootloader and I unplug it and plug it into my computer, it’s still in bootloader, but the drive doesn’t appear until I powercycle the module"

sanni-t:
Closes [#4097](https://github.com/Opentrons/opentrons/issues/4097)
Addresses [#4113](https://github.com/Opentrons/opentrons/issues/4113)
## overview

The uf2 bootloader had been giving a lot of trouble on windows machines. It also had little annoying features like starting up in bootloader if the device was reset while in bootloader & waiting for firmware file. So we decided to switch it with the SAM-BA bootloader base used in the Arduino SAMD core boards. This bootloader needs a dedicated tool for flashing firmware to the board. We will be using the BOSSA commandline tool- bossac, again used by Arduino. With these things in mind, following changes were made to enable proper use of this bootloader for thermocycler and thermocycler production:

- made changes to the SAM-BA bootloader to disable fans and peltiers
- added the SAM-BA bootloader mod with source files to ThermocyclerM0 board defs, bumped arduino board version
- changed thermocycler firmware_uploader.py and serial_and_firmware_uploader.py to use bossac for firmware upload
- updated makefile to use 1.6.21 Arduino SAMD version since it has been found that the bossac configurations used in the latest 2 versions results in errors while uploading from Arduino

## requests

- [ ] Check if the bootloader funstions fine- fans don't turn on. Other weird things don't happen
- [ ] Serial & f/w upload scripts work well on both Mac & Windows


modules/thermo-cycler/production/firmware_uploader.py
```diff
@@ -32,11 +33,14 @@ def build_arg_parser():
...
     return arg_parser
 
-def find_opentrons_port():
+def find_opentrons_port(bootloader=False):
     retries = 5
     while retries:
         for p in comports():
             if p.vid == OPENTRONS_VID:
+                if bootloader:
+                    if p.pid != TC_BOOTLOADER_PID:
```
sfoster1, line 42: Shouldn't we also skip this if !bootloader and p.pid == TC_BOOTLOADER_PID

sfoster1 (commented):
Just a quick question about the port finder, overall looks fine if it works

modules/thermo-cycler/production/firmware_uploader.py
sanni-t: Not really. Case in point- right after flashing the bootloader, the board has bootloader PID only. When we have to upload firmware on it, the script starts by looking for an opentrons port so that it can perform the 1200 baud touch and put it in bootloader mode. Right now, it performs the 1200baud touch on bootloader port itself (which has no effect) and then sees the bootloader port (which was already there) to upload firmware to. If we made it so that if !bootloader it should only look for a non bootloader port then it will never be able to find such a port after a bootloader upload.
If we want to make the conditions tighter so that this script doesn't work on anything other than a thermocycler, then we can have it look for ports that are either tc_bootloader or tc port only

b-cooper (approved):
functions as expected, scripts and all ☯️

PR #90, opened 2019-10-24
feat(thermocycler): add thermistor offset, PID control fans, overshoots for volume

### LINKED ISSUE #3643: Thermocycler: firmware: update fan speeds and calculate fan induced temp offsets
## overview

Update the the fan speeds during different states- cooling/ heating/ ramping up/ ramping down/ heatsink too hot

Also, the fan has a consistent effect on the temperature of peltiers (for example, for temperatures > 70C, fan speed of 20% shifts the peltiers by about +0.4C). So we need to record this effect for strategic temperature points and fan speeds and create an offset lookup table to compensate for this drift in temperature.

Tasks:
1. Change safety temp limit to 68C or 5C less than target for temps between 50C-70C
2. Add the linear equation found from tests and check performance
3. If performance not on par across different temperatures, use exponential equation and check performance

- [ ] Add offsets disable gcode

### LINKED ISSUE #4144: Thermocycler: firmware: read the offset equation constants from EEPROM
## overview

- possibly create a new gcode to write n read constants from eeprom & use these in calculation
- add to serial_and_fw_upload script: writing the constants to EEPROM

### LINKED ISSUE #4395: Thermocycler: firmware: temperature adjustments according to volume
## overview

Different volumes of liquid take different lengths of time to heat/cool to the same target temperature. Since we can't afford this change in time-to-temp for each volume, we will need to give some 'push' in the right direction in heating/cooling. This would most probably be in the form of overshoots in our temperature profile. The overshoot will vary according to the liquid volume. e.g, more volume would have a bigger overshoot to provide it a bigger push. 
These overshoots will live in a lookup table or be generated dynamically on the firmware based on a regression model. `set_temperature` gcode will have a `V` (volume) parameter to specify the volume for each target temp. If no volume specified, default of 25uL overshoot will be implemented.

## Tasks

- [ ] Add volume param to `set_temperature`
- [ ] Gather overshoot data for different volumes
- [ ] Add overshoots data and adjust target temperatures accordingly

sanni-t:
This is the PR overarching all temperature tuning issues..
Closes [#3643](https://github.com/Opentrons/opentrons/issues/3643): Adds offsets to thermistors based on the regression model formed with parameters of current heatsink temperature and target temperature. Changed the fan behavior to suit regression model conditions.

Closes [#4144](https://github.com/Opentrons/opentrons/issues/4144): Offset constants are saved to & read from EEPROM. Added gcodes to read and write the constants.

Closes [#4395](https://github.com/Opentrons/opentrons/issues/4395): Added temperature adjustments according to liquid volume. Used a regression model to dynamically calculate the adjustment/ 'overshoot' values for different volumes. Volume amount is passed in `set_temperature` gcode with the param `V`. eg., `M104 S70 V50` sets target 70C for 50uL liquid volume

## Additional changes
- Introduced a 'plate deactivate' gcode which disables just the plate. Required not just right now while testing offset & tuning outcomes but also was mentioned as a long term requirement for protocols.
- Added the ability to edit & read PID constants using gcodes
- Added PID control for fans so we can continuously monitor and control fans based on heatsink temperatures
- Adjusted start time of hold timer to accommodate for the overshoots

## Review requests
- [x] The fan behaves as expected
- [x] The offsets give us results within our specs of +/-0.2 degree C
- [x] Offsets are saved to & read from & implemented correctly
- [x] The new gcodes work as expected
- [x] We get expected overshoots for different liquid volumes
- [x] The code looks good

### DIFF
--- a/modules/thermo-cycler/thermo-cycler-arduino/eeprom.cpp
+++ b/modules/thermo-cycler/thermo-cycler-arduino/eeprom.cpp
@@ -116,6 +116,7 @@ bool Eeprom::set_offset_constant(OffsetConst constant, float val)
     {
       Serial.print("\nError: Failed to write to address:");
       Serial.println(byte(addr+i));
+      pinMode(WP_PIN, INPUT);
       return false; // failed
     }
   }
--- a/modules/thermo-cycler/thermo-cycler-arduino/eeprom.cpp
+++ b/modules/thermo-cycler/thermo-cycler-arduino/eeprom.cpp
@@ -105,12 +105,17 @@ bool Eeprom::set_offset_constant(OffsetConst constant, float val)
     case OffsetConst::C:
       addr = C_LOC;
       break;
+    default:
+      Serial.println("\nError: No constant name provided");
+      return false;
   }
   _erase_eeprom_section(addr, uint8_t(sizeof(this_const.f_val)));
   for (uint8_t i = 0; i < sizeof(this_const.f_val); i++)
   {
     if(!_write_char(byte(addr+i), byte(this_const.b_val[i])))
     {
+      Serial.print("\nError: Failed to write to address:");
+      Serial.println(byte(addr+i));
       return false; // failed
     }
   }
@@ -123,7 +128,8 @@ bool Eeprom::_erase_eeprom_section(uint8_t addr, uint8_t sec_len)
   {
     if(!_write_char(byte(addr+i), byte(0xff)))  // erasing = writing 0xffh
     {
-      Serial.println("\nfailed to erase section");
+      Serial.print("\nError: Failed to erase at address:");
+      Serial.println(byte(addr+i));
       return false;
     }
   }
@@ -145,6 +151,9 @@ float Eeprom::get_offset_constant(OffsetConst constant)
     case OffsetConst::C:
       addr = C_LOC;
       break;
+    default:
+      Serial.println("\nError: No constant name provided");
+      return false;
   }
   for (uint8_t i = 0; i < sizeof(this_const.f_val); i++)
   {
--- a/modules/thermo-cycler/thermo-cycler-arduino/eeprom.cpp
+++ b/modules/thermo-cycler/thermo-cycler-arduino/eeprom.cpp
@@ -1,6 +1,12 @@
 #include "eeprom.h"
 #include <ot_shared_data.h>
 
+union ConstValue
+{ // To access the float value byte by byte
+  float f_val;
+  byte b_val[sizeof(f_val)];
+};
+
 Eeprom::Eeprom()
 {}
 
@@ -49,6 +55,20 @@ String Eeprom::read(MemOption option)
 #endif
 }
 
+bool Eeprom::_write_char(uint8_t word_address, uint8_t value)
+{
+  Wire.beginTransmission(OT_EEPROM_ADDR);
+  Wire.write(word_address);
+  Wire.write(value);
+  byte error = Wire.endTransmission();
+  delay(OT_WR_TIME);
+  if (error)
+  {
+    return false;
+  }
+  return true;
+}
+
 char Eeprom::_read_char(uint8_t word_address)
 {
 #if HW_VERSION >= 4
@@ -65,6 +85,72 @@ char Eeprom::_read_char(uint8_t word_address)
   }
 #endif
   return '~';
+}
+
+bool Eeprom::set_offset_constant(OffsetConst constant, float val)
+{
+  pinMode(WP_PIN, OUTPUT);
+  digitalWrite(WP_PIN, LOW); // Disables write protect
+  ConstValue this_const;
+  this_const.f_val = val;
+  uint8_t addr;
+  switch(constant)
+  {
+    case OffsetConst::A:
+      addr = A_LOC;
+      break;
+    case OffsetConst::B:
+      addr = B_LOC;
+      break;
+    case OffsetConst::C:
+      addr = C_LOC;
+      break;
+  }
+  _erase_eeprom_section(addr, uint8_t(sizeof(this_const.f_val)));
+  for (uint8_t i = 0; i < sizeof(this_const.f_val); i++)
+  {
+    if(!_write_char(byte(addr+i), byte(this_const.b_val[i])))
+    {
+      return false; // failed
+    }
+  }
+  pinMode(WP_PIN, INPUT);
+}
+
+bool Eeprom::_erase_eeprom_section(uint8_t addr, uint8_t sec_len)
+{
+  for (uint8_t i = 0; i < sec_len; i++)
+  {
+    if(!_write_char(byte(addr+i), byte(0xff)))  // erasing = writing 0xffh
+    {
+      Serial.println("\nfailed to erase section");
+      return false;
+    }
+  }
+  return true;
+}
+
+float Eeprom::get_offset_constant(OffsetConst constant)
+{
+  ConstValue this_const;
+  uint8_t addr;
+  switch(constant)
+  {
+    case OffsetConst::A:
+      addr = A_LOC;
+      break;
+    case OffsetConst::B:
+      addr = B_LOC;
+      break;
+    case OffsetConst::C:
+      addr = C_LOC;
+      break;
+  }
+  for (uint8_t i = 0; i < sizeof(this_const.f_val); i++)
+  {
+    this_const.b_val[i] = _read_char(addr+i);
+  }
+  return this_const.f_val;
 }
 
 void Eeprom::setup()
--- a/modules/thermo-cycler/thermo-cycler-arduino/thermistorsadc.h
+++ b/modules/thermo-cycler/thermo-cycler-arduino/thermistorsadc.h
@@ -84,6 +84,7 @@ class ThermistorsADC{
             float back_right_temperature();
             float cover_temperature();
             float heat_sink_temperature();
+            float plate_temp_offset;
             bool detected_invalid_val;
 
       private:
--- a/modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
+++ b/modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
@@ -47,6 +47,15 @@ PID PID_Cover(
   &current_temperature_cover, &temperature_swing_cover, &target_temperature_cover,
   PID_KP_COVER, PID_KI_COVER, PID_KD_COVER, P_ON_M, DIRECT);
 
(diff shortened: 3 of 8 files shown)


andy-g-calderon:
The fan behavior and offsets are working well.

andy-g-calderon (approved):
Everything worked well across all temp ranges

modules/thermo-cycler/thermo-cycler-arduino/eeprom.cpp
```diff
@@ -67,6 +87,72 @@ char Eeprom::_read_char(uint8_t word_address)
   return '~';
 }
 
+bool Eeprom::set_offset_constant(OffsetConst constant, float val)
+{
+  pinMode(WP_PIN, OUTPUT);
+  digitalWrite(WP_PIN, LOW); // Disables write protect
+  ConstValue this_const;
+  this_const.f_val = val;
+  uint8_t addr;
```
sfoster1, line 96: technically has an uninitialized path if the enum isn't valid? maybe have a `default` in the case that returns `false`

modules/thermo-cycler/thermo-cycler-arduino/eeprom.cpp
```diff
@@ -67,6 +87,72 @@ char Eeprom::_read_char(uint8_t word_address)
...
+    case OffsetConst::C:
+      addr = C_LOC;
+      break;
+  }
+  _erase_eeprom_section(addr, uint8_t(sizeof(this_const.f_val)));
+  for (uint8_t i = 0; i < sizeof(this_const.f_val); i++)
+  {
+    if(!_write_char(byte(addr+i), byte(this_const.b_val[i])))
+    {
+      return false; // failed
```
sfoster1, line 120: if we're ok with calling `Serial.println` from internal code, maybe print an error here?

modules/thermo-cycler/thermo-cycler-arduino/eeprom.cpp
```diff
@@ -67,6 +87,72 @@ char Eeprom::_read_char(uint8_t word_address)
...
+  pinMode(WP_PIN, INPUT);
+}
+
+bool Eeprom::_erase_eeprom_section(uint8_t addr, uint8_t sec_len)
+{
+  for (uint8_t i = 0; i < sec_len; i++)
+  {
+    if(!_write_char(byte(addr+i), byte(0xff)))  // erasing = writing 0xffh
+    {
+      Serial.println("\nfailed to erase section");
```
sfoster1: does this need an error keyword for the upstream parsing? like `"\nerror: failed to erase section"`? Either way it should probably print the address that failed

modules/thermo-cycler/thermo-cycler-arduino/eeprom.cpp
```diff
@@ -67,6 +87,72 @@ char Eeprom::_read_char(uint8_t word_address)
...
+  {
+    case OffsetConst::A:
+      addr = A_LOC;
+      break;
+    case OffsetConst::B:
+      addr = B_LOC;
+      break;
+    case OffsetConst::C:
+      addr = C_LOC;
+      break;
```
sfoster1, line 154: Same thing about a possible uninitialized path

modules/thermo-cycler/thermo-cycler-arduino/thermistorsadc.h
```diff
@@ -84,6 +84,7 @@ class ThermistorsADC{
             float back_right_temperature();
             float cover_temperature();
             float heat_sink_temperature();
+            float plate_temp_offset;
```
sfoster1, line 87: This definitely needs a default value initializer in the ctor

modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
```diff
@@ -264,32 +304,107 @@ void update_fans_from_state()
...
+          else
+          {
+            if (PID_fan.Compute())
+            {
+              heatsink_fan.set_percentage(fan_pid_out);
+            }
+          }
+        }
       }
       else if (is_ramping_down())
```
sfoster1: can we add `// && !is_target_cold()` or something just to help keep track of this

modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
```diff
@@ -250,11 +289,12 @@ void update_fans_from_state()
   {
```
sfoster1, line 292: I know this is a pain, but the number of if/else cascades in this function is getting pretty insane. Could we maybe refactor this to
- just bail early if `millis() - last_checked <= 100)
- have a different function to call if `master_set_a_target` or not or something
- if those functions are still long enough that you can get lost in them, either add more or add some comments indicating where they are in the chain

modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
```diff
@@ -527,9 +683,56 @@ void read_gcode()
...
+            if (!eeprom.set_offset_constant(OffsetConst::C, gcode.popped_arg()))
+            {
+              gcode.response("EEPROM ERROR","offset C not saved");
+            }
+          }
+          update_offset_constants();
+          break;
+        case Gcode::get_offset_constants:
+          /**** Get current constants ****/
+          gcode.response("a", String(const_a, 4));
```
sfoster1: looks like we require upper case in the request and then provide lower case here, we should keep it consisent

modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
```diff
@@ -696,6 +899,28 @@ void TC4_Handler()
...
+  }
+  if (isnan(eeprom.get_offset_constant(OffsetConst::C)))
+  {
+    eeprom.set_offset_constant(OffsetConst::C, CONST_C_DEFAULT);
+  }
+}
+
+void update_offset_constants()
+{ // Should be called every time after updating EEPROM values
+  const_a = eeprom.get_offset_constant(OffsetConst::A);
```
sfoster1, line 950: maybe explicit cast to void for the return here like `(void)eeprom.get_offset_constant(OffsetConst::A)`

modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
```diff
@@ -846,18 +1096,81 @@ void therm_pid_peltier_update()
...
+  {
+    return false;
+  }
+}
+
+double plate_overshoot()
+{
+  if (this_step_target_temp >= current_temperature_plate)
+  {
+    return (0.0105 * current_volume + 1.0869);
```
sfoster1: where do these numbers come from? should they be in eeprom? should they be `#define`s?

modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
```diff
@@ -846,18 +1096,81 @@ void therm_pid_peltier_update()
...
+  }
+}
+
+void adjust_target_for_volume()
+{
+  if (just_changed_temp)
+  {
+    if (!timer_started && crossed_true_target())
+    {
+      // Serial.println("started timer");
```
sfoster1: Commented code

modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
```diff
@@ -696,6 +899,28 @@ void TC4_Handler()
   }
 }
 
+void check_saved_offsets()
+{ // If erroneous value or no offsets saved in EEPROM, then save the defaults to EEPROM
+  if (isnan(eeprom.get_offset_constant(OffsetConst::A)))
```
sfoster1: this will work but it makes me feel EXTREMELY bad to do it this way. this is a check on byte values that should really be done _at the level of the byte values_, not by what they happen to represent when interpreted as a float. could we have a function that checks if the value is set?

modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
```diff
@@ -846,18 +1096,81 @@ void therm_pid_peltier_update()
...
+      { // stayed at overshoot temp long enough. Go to real target temp
+        // Serial.println("Done with overshoot. Going back to real target");
+        just_changed_temp = false;
+        overshoot_start_timestamp = 0;
+        timer_started = 0;
+        target_temperature_plate = this_step_target_temp;
+      }
+      else
+      { // need to stay at overshoot temp for longer
+        // Serial.println("staying at overshoot");
```
sfoster1: commented out code?

sfoster1 (changes requested):
If it works it works, but I have some change requests inline.

andy-g-calderon:
Quick summary of 50 uL science testing.

Annika ran a reaction of 50 uL three times on a Themocylcer and the %CV was less than our specified 10% for a reaction

andy-g-calderon (approved):
Science was sucessfully  completed at 50 uL three times. The %CV was less than our 10% product specification.

b-cooper (approved):
Looks to me pending Seth's comments :octocat:

modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
sfoster1: miread this, my bad

modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
sanni-t: This check is mainly used to update eeprom when there were no values saved previously (Happens when a new thermocycler has firmware uploaded for the first time).
We do check if the value is set or not in `set_offset_constant`. If we error while writing a byte to eeprom then an error is raised and then `check_saved_offsets` checks whether the error resulted in a garbage/nan value at the address. It then overwrites the garbage value with default values. We don't do anything further if these default values don't get written though. That's a very rare edge case and it'd probably mean the eeprom's damaged anyway so there's really not much that can be repaired by firmware. The errors should take care of informing the user not to proceed further

sfoster1: OK, that sounds good. I'd still love to check byte values rather than float representation but that's ok.

sfoster1 (approved):
LGTM!

PR #91, opened 2019-10-29
chore(CI, thermocycler): deploy thermocycler builds to s3

sanni-t:
Closes [#4045](https://github.com/Opentrons/opentrons/issues/4045)

This  PR adds deploy to travis CI. It's configured to deploy all modules binaries to opentrons-modules-builds bucket on all branches. The deployed artifact name is appended with tag#, number of commits since last tag (for untagged commits) and SHA short (all from `git describe --tag`) and branch name.

This PR also removes the binaries from thermocycler production folder and keeps them only in ./build. The makefile copies each binary's uploader on build. 

The firmware version is now based on `git describe --tag` of the commit. If no tag description was provided while building then the firmware version is `unknown`


modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler.h
```diff
@@ -12,7 +12,13 @@
 #include "eeprom.h"
 
 /********* Version **********/
-#define FW_VERSION "v1.0.2"
+/********* Versions **********/
+/* Version guidelines: */
+#ifdef TC_FW_VERSION
+  #define FW_VERSION String(TC_FW_VERSION)
+#else
+  #define FW_VERSION "unknown"
```
sfoster1: Let's do an `#error` or something here, I think it's ok to either require you to make a code change to set it or use the makefile

sfoster1 (approved):
LGTM other than the error note

PR #92, opened 2019-10-30
Temp probe data collector

Carlos-fernandez:
These test scripts are used for thermocycling testing

Eutechnics Driver:
The Eutechnics serial driver was copied from Mechanical-Test Repo to this repo locally. The driver establishes a connection to the 4500 Eutechnics temperature probe. Allowing a user to extract the necessary data from the measuring device. 

TC_mini_driver:
The TC_mini_driver.py is a Thermocycler driver i created to write simple Gcode commands to thermocycler.  

offset_test:
The Offset test runs the thermocycler in a thread, while running separate thread to the Eutechnics thermometer. This helps us sync data collection from the thermocycler and the temperature probe measured from the PCR top plate. This test script was used in the first round of offset tesing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,6 +69,7 @@ coverage.xml
 
 # Data from QC tests
 *.csv
+*.xlsx
 
 # Translations
 *.mo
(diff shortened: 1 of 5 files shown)


sanni-t:
Replaced by #107

PR #94, opened 2019-11-13
ci(travis): make deployed file readable

sanni-t:
This PR makes the deployed firmware binaries' zip publicly readable.


sfoster1 (approved):
You could also just leave the bucket open (even if it yells at you).

mcous (approved):
I get scary emails from AWS about public access buckets, so I prefer this where possible

PR #95, opened 2019-11-21
feat(thermocycler): make fans start gradually, don't overshoot for ~same temp, cleanup

### LINKED ISSUE #3609: Thermocycler: firmware: make fan start gradually
## overview ##

Right now the fan goes from 0 to 70% in an instant, which isn't a great user experience

Change this behavior so the fan always starts gradually

sanni-t:
Closes [#3609](https://github.com/Opentrons/opentrons/issues/3609)
## overview

Has following changes:
1. For non-PID fan settings that set a high fan speed, the fans will now ramp up to the high speed instead of jumping to high speed straight away. It's done by implementing a low pass filter. This was needed for a better user experience; we don't want to startle users every time the TC is cooling down.
2. Don't overshoot if the temp we want to go to is very close to the current temperature.
3. As part of safety check, also check for thermistor uniformity. If any thermistors are damaged they will give a much different reading than rest of the thermistors.
4. Removed some flags that we are no longer using.

### DIFF
--- a/modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
+++ b/modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
@@ -582,10 +582,10 @@ void read_gcode()
           }
           // `target_temperature_plate`: The target temperature the plate will
           //  go to in order to compensate for thermal lag. This will basically
-          //  amount to inflated targets (overshoots) for a Kt amount of time after
-          //  the target is set.
-          //  It will return to `this_step_target_temp` after Kt seconds or
-          //  once the well temperatures catch up with the peltiers
+          //  amount to inflated targets (overshoots) for a OVERSHOOT_DURATION
+          //  amount of time after the target is set.
+          //  It will return to `this_step_target_temp` after the above set time
+          //  or once the well temperatures catch up with the peltiers
           target_temperature_plate = this_step_target_temp + plate_overshoot();
         #else
           target_temperature_plate = gcode.popped_arg();
@@ -1034,6 +1034,7 @@ void setup()
 
 void temp_safety_check()
 {
+
   if (!master_set_a_target)
   {
     return;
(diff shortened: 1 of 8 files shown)


modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
```diff
@@ -1051,6 +1052,15 @@ void temp_safety_check()
     gcode.response("System too hot! Deactivating.");
     deactivate_all();
   }
+  // When peltiers are stable, check if all thermistors give approximately
+  // the same reading. If they are not same, then there might be a problem with
+  // some/ one of the thermistors and might need to be replaced.
+  if (!just_changed_temp &&
+      temp_probes.hottest_plate_therm_temp() - temp_probes.coolest_plate_therm_temp() > ACCEPTABLE_THERM_DIFF)
+  {
+    gcode.response("Plate temperature not uniform. Deactivating.");
```
sfoster1: Does this need an error keyword

modules/thermo-cycler/thermo-cycler-arduino/thermo-cycler-arduino.ino
```diff
@@ -1154,14 +1164,18 @@ bool crossed_true_target()
 
 double plate_overshoot()
 {
-  if (this_step_target_temp >= current_temperature_plate)
+  if (this_step_target_temp - current_temperature_plate >= 0.5)
   {
     return (POS_OVERSHOOT_M * current_volume + POS_OVERSHOOT_C);
   }
-  else
+  else if (this_step_target_temp - current_temperature_plate <= 0.5)
```
sfoster1: this should be `<= -0.5` right?

sfoster1 (changes requested):
Couple inline changes, generically looks good. Nice filter for the fan.

sfoster1 (approved):
Code LGTM

PR #96, opened 2019-12-03
feat(thermocycler): add error indication with leds and serial messages

sanni-t:
## overview

This PR adds user-facing error indication of system level errors generated by the thermocycler. This would be useful as a fallback until we put together some structure for detecting and reporting error on the app.
Right now the thermocycler detects 3 system level errors:
`system_too_hot`, `invalid_thermistor_value`, `plate_temperature_not_uniform`

## changes
- organized all (3) systems errors such that they get set as flags in a `system_errors` byte.
- added an LED flashing function that gets implemented if any flag in system_errors is set.
- added a function that just reports all errors indicated by `system_errors` at a regular interval.
- all the 3 errors will now prevent a user from setting a temperature until the issue is fixed *and* the module is power-cycled.
- made `ACCEPTABLE_THERM_DIFF` 4 degC (for now) after discussion with Taylor & Andy

### DIFF
--- a/modules/thermo-cycler/thermo-cycler-arduino/lights.cpp
+++ b/modules/thermo-cycler/thermo-cycler-arduino/lights.cpp
@@ -40,6 +40,9 @@ void Lights::update()
     case Light_action::wipe:
       _color_wipe(_color);
       break;
+    case Light_action::flashing:
+      _flash_on_off(_color);
+      break;
     case Light_action::all_off:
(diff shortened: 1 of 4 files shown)


sfoster1 (approved):
LGTM, should handle the problem nicely

b-cooper (approved):
Looks good 🍡

PR #97, opened 2019-12-04
revert(thermocycler): revert to not using fan ramping when ramping down plate temp

sanni-t:
HW team pointed out that we should test if ramping the fans has any effect on the performance of the thermocycler. Since the fan ramping while going from a higher temp to lower temp (targets > room temp) has some effect on the temperature offsets, it's best to wait to implement this feature until we test it.

The effect of fan when going to targets < room temp isn't detrimental to performance as far as we know, so, ramping during `fans_for_cold_target()` can be left as is.


sfoster1 (approved):
LGTM

PR #100, opened 2019-12-11
feat(thermocycler): use offsets for all temps

sanni-t:
## overview

Turns out we need the offsets at lower temperatures too. Luckily, the same existing equation works for lower temperatures too.

## changes
- enable offset calculation for lower temps
- Makefile: use only the version number part of TC_FW_VERSION passed by travis (so it'll be `v1.1.0` instead of `thermocycler@v1.1.0`)


sfoster1 (approved):
LGTM

PR #101, opened 2019-12-19
chore(CI, modules): filter tags and use separate fw version for each module

sanni-t:
## overview

This PR adds versioning for magdeck & tempdeck through git tags. Similar to the thermocycler, the two modules will extract version number from appropriately named tags eg. 'tempdeck@v2.1.0' for tempdeck or 'magdeck@v1.1.0' for magdeck. 

## changes
- pass magdeck & tempdeck firmware versions from git tags to firmware during build
- deploy release firmware with the 'release' suffix

## review requests
- Does the deploy scheme make sense? 
~- The deployed bundle still consists of all 3 modules' firmware bundled together in a zip~ (Changed as per requested to deploy each module firmware in its own zip), except now the firmwares will contain their own version number. So each module will have its own, independent release but the bundle will also contain non-release versions for the other modules. Would this work for the update api?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,23 +7,39 @@ cache:
   directories:
     - ~/arduino_ide
 
-env:
-  - THIS_BUILD_TAG=`git describe --tags`
-
 install:
   - make setup
 
+env:
+  global:
+    - THIS_BUILD_TAG=`git describe --tags`
+    - MD_TAG=`git describe --tags --match "magdeck*"`
+    - TD_TAG=`git describe --tags --match "tempdeck*"`
+    - TC_TAG=`git describe --tags --match "thermocycler*"`
+
+jobs:
+  include:
+    - stage: deploy1
+      name: Branch build
+      env: BRANCH_NAME=$TRAVIS_BRANCH
+      if: tag IS blank
+
+    - stage: deploy2
+      name: Release build
+      env: BRANCH_NAME=release
+      if: tag IS present
+
 script:
-  - make build TC_FW_VERSION=$THIS_BUILD_TAG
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -7,23 +7,39 @@ cache:
...
 
 before_deploy:
   - make clean
-  - mkdir modules-$THIS_BUILD_TAG-$TRAVIS_BRANCH
-  - cp -r ./build modules-$THIS_BUILD_TAG-$TRAVIS_BRANCH
-  - zip -r modules-$THIS_BUILD_TAG-$TRAVIS_BRANCH.zip modules-$THIS_BUILD_TAG-$TRAVIS_BRANCH
-  - rm -rf modules-$THIS_BUILD_TAG-$TRAVIS_BRANCH
+  - mkdir modules-$THIS_BUILD_TAG-$BRANCH_NAME
+  - cp -r ./build modules-$THIS_BUILD_TAG-$BRANCH_NAME
+  - zip -r modules-$THIS_BUILD_TAG-$BRANCH_NAME.zip modules-$THIS_BUILD_TAG-$BRANCH_NAME
```
sfoster1: Maybe instead of doing an explicit `zip` here just upload the entire output directory? that way things can choose whether they want to download a single zip with everything or an individual module firmware

sfoster1 (changes requested):
Yeah I think we probably should separate the tags a bit better. Instead of uploading one zip, just upload the folder.

sfoster1 (approved):
The kind of annoying thing is that the s3 prefix where things are uploaded just depends on the last tag, without any filtering; that's a little annoying but we can work with it and I'm not sure what a better way would be.

PR #102, opened 2019-12-22
feat(tempdeck): add v15 changes, use s3 artifact in production

sanni-t:
## overview

This PR adds small changes to the tempdeck firmware in order to accommodate for h/w changes in v15 tempdecks. This PR also reorganizes eepromWriter for tempdeck and bundles relevant files into the s3 build artifact so we can start using this versioned-tracked firmware in production

## changes
- F/W: use pwm fan settings for v15, add safety check to stabilize_to_room_temp to account for peltier 'runaway' from overheating. If the tempdeck has overheated, the peltiers would be disabled and fan turned on on high. The tempdeck will also send an error over serial. To get out of the error state, the user will have to power cycle the module
- Makefile: bundle the tempdeck uploader & supporting files for production
- Remove tempdeck hex file since we'll be using the CI built one
- Moved eepromWriter.hex out of /firmware and updated write_module_memory.py script so that the script will work for s3 deployed tempdeck firmware as well as for magdeck hex file in the repo (we will make changes to magdeck files later)

## review requests
- Code-wise: if code looks fine and deployment method works
- @TaylorGoodman, @andy-g-calderon  : Please verify if the v15 tempdeck as well as previous versions work fine with this firmware version


sfoster1 (approved):
LGTM if it all works

PR #103, opened 2020-01-15
chore(CI): deploy process per module, one travis job 

b-cooper:
## Overview

This branch incorporates a couple of changes to the release build deployment process for module firmware.  The branch build artifact generation should remain largely unchanged.

The release process is now as follows:
- Push a tag, e.g. "thermocycler@v3.2.1"
- just the "thermocycler" artifacts will be built in travis
- those artifacts will then be dropped (not compressed) in the modules builds s3 bucket within './thermocycler/v3.2.1/' sub-directory

This will give us a steady location for accessing released firmware binaries for the opentrons modules. Especially from the OT-2 software build process in the forthcoming addition at https://github.com/Opentrons/buildroot/pull/67.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,41 +13,48 @@ install:
 env:
   global:
     - THIS_BUILD_TAG=`git describe --tags`
-    - MD_TAG=`git describe --tags --match "magdeck*"`
-    - TD_TAG=`git describe --tags --match "tempdeck*"`
-    - TC_TAG=`git describe --tags --match "thermocycler*"`
+
+# base configs
+_deploy_s3: &deploy_s3
+  provider: s3
+  skip_cleanup: true
+  access_key_id: $AWS_ACCESS_KEY_ID
+  secret_access_key: $AWS_SECRET_ACCESS_KEY
+  region: us-east-2
+  acl: public_read
+  bucket: opentrons-modules-builds
+  dot_match: true
+  on:
+    all_branches: true
 
 jobs:
   include:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -13,41 +13,48 @@ install:
...
+      script:
+        - make build
+        - make zip-all
+      before_deploy:
+        - make clean
+      deploy:
+        - # upload all branch artifacts to S3
+          <<: *deploy_s3
+          local_dir: ./build
+          upload_dir: modules-$THIS_BUILD_TAG-$TRAVIS_BRANCH
```
sanni-t: $TRAVIS_BRANCH is not defined

mcous (approved):
LGTM 🍭

sanni-t (commented):
Everything looks good but there is one not-issue but more of an inconvenience this change in structure creates- Since the release artifact is not a zip anymore but separate files, there won't be a (single) link of updated firmware and related files bundled together that we can send to production. One way we could achieve that is by manually bundling every release in a single zip and uploading it as a release asset on github. 
Unless it can be automated too.

sanni-t:
So, I guess what I'm trying to say is we can go ahead with this structure and just make sure to incorporate the bundling of all files in our firmware release process. Or, if it is possible, we could just have a release upload a binary file AND a zip of all files (including the binaries & firmware uploaders) so the binary can be used by buildroot/ api and the zip would go to production team

sanni-t (approved):

PR #106, opened 2020-02-06
fix(magdeck): return false when no number read

### LINKED ISSUE #105: Magdeck: Build of current magdeck code results in magdeck 'freezing' on move cmd
## overview

If the current source files of magdeck are compiled using Arduino 1.8.10 and uploaded to a magdeck, then the module will freeze while executing any 'move' type commands like `Move to position` or `probe plate`

sanni-t:
Closes #105 

The code was interpreting absence of a gcode parameter as 'parameter present' (in `gcode.read_number()`). The only gcodes that have optional parameters are the movement type gcodes which have current and speed values as optional. Because of this, the gcode interpreter was reading empty values and setting motor current and speed to zero. This was causing the motor to not move, which was making the firmware to freeze and not respond to any serial input.

Adding the else condition to `read_number` makes sure that the gcode interpreter doesn't read from a non existent parameter & fixes the issue.

Note:
It is possible that Arduino1.8.5 (which was used to build the original firmware 2 years ago) had a different (compiler?) behavior that resulted in default return value of 0 because of which this bug wasn't coming to light. I will be testing this

Update:
The above note is correct.

### DIFF
--- a/modules/mag-deck/mag-deck-arduino/gcodemagdeck.cpp
+++ b/modules/mag-deck/mag-deck-arduino/gcodemagdeck.cpp
@@ -84,6 +84,10 @@ bool GcodeMagDeck::read_number(char key) {
       return false;
     }
   }
+  else
+  {
+    return false;
+  }
 }
 
 void GcodeMagDeck::print_device_info(String serial, String model, String version) {


sfoster1 (approved):
Code change LGTM, will cycle back with testing

b-cooper (approved):
📖 #️⃣ looks good. Will test on a device

b-cooper:
Tested on magnetic module. ⚙️ behaves as expected

PR #110, opened 2020-02-13
fix(tempdeck): correct model check to handle version 3

sfoster1:
This model_version check would be incorrect for version 3, and it would
act as if it was a v1 or 2 tempdeck


b-cooper (approved):
🌡

PR #111, opened 2020-02-13
feat(write_module_memory): support gen2 magdecks and tempdecks

sfoster1:
These will save their model version as 20 to fit with how we do the rest
of the models (e.g. 1.5) and avoid firmware refactors


b-cooper (approved):
🚶

PR #114, opened 2020-03-18
fix(magdeck): set steps/mm to 100 for gen2 and later models

### LINKED ISSUE #113: Magdeck: on gen2 and later, set steps/mm to 100
Magnetic modules have a physical steps/mm value of 100 steps/mm. Unfortunately, this value is in the firmware as 50 and all extent checking is based on that, so when you put in numbers for motion targets they're effectively in mm/2. This has to change.

If the module detects it is a gen2, it should use a value of 100 steps/mm and possibly tweak other motion parameters to behave similarly to the gen1s but with the correct units.

ahiuchingau:
This PR changes the `steps_per_mm` value for gen2 or later magdeck models to 100. 

Note: It now checks the model version first and set steps/mm before the rest of the setup. Will this affect anything that I'm not aware of?

closes #113

### DIFF
--- a/modules/mag-deck/mag-deck-arduino/mag-deck-arduino.ino
+++ b/modules/mag-deck/mag-deck-arduino/mag-deck-arduino.ino
@@ -24,6 +24,9 @@
 String device_serial = "";  // leave empty, this value is read from eeprom during setup()
 String device_model = "";   // leave empty, this value is read from eeprom during setup()
 
+#define MODEL_VER_TEMPLATE "mag_deck_v"
+#define MODEL_VER_TEMPLATE_LEN sizeof(MODEL_VER_TEMPLATE) - 1
+
 GcodeMagDeck gcode = GcodeMagDeck();  // reads in serial data to parse command and issue reponses
 Memory memory = Memory();  // reads from EEPROM to find device's unique serial, and model number
 
@@ -45,8 +48,11 @@ Memory memory = Memory();  // reads from EEPROM to find device's unique serial, 
 
 #define CURRENT_TO_BYTES_FACTOR 114
 
-#define STEPS_PER_MM 50  // full-stepping
-unsigned long STEP_DELAY_MICROSECONDS = 1000000 / (STEPS_PER_MM * 10);  // default 10mm/sec
+// #define steps_per_mm 50  // full-stepping
+// unsigned long step_delay_microseconds = 1000000 / (steps_per_mm * 10);  // default 10mm/sec
+
+int steps_per_mm = 50;  // full-stepping for GEN1
+unsigned long step_delay_microseconds = 1000000 / (steps_per_mm * 10);  // default 10mm/sec
 
 #define MAX_TRAVEL_DISTANCE_MM 40
 float FOUND_HEIGHT = MAX_TRAVEL_DISTANCE_MM - 15;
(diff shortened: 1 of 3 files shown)


modules/mag-deck/mag-deck-arduino/mag-deck-arduino.ino
```diff
@@ -45,8 +48,11 @@ Memory memory = Memory();  // reads from EEPROM to find device's unique serial,
 
 #define CURRENT_TO_BYTES_FACTOR 114
 
-#define STEPS_PER_MM 50  // full-stepping
-unsigned long STEP_DELAY_MICROSECONDS = 1000000 / (STEPS_PER_MM * 10);  // default 10mm/sec
+// #define steps_per_mm 50  // full-stepping
```
sfoster1: remove commented-out code

modules/mag-deck/mag-deck-arduino/mag-deck-arduino.ino
```diff
@@ -45,8 +48,11 @@ Memory memory = Memory();  // reads from EEPROM to find device's unique serial,
...
+// unsigned long step_delay_microseconds = 1000000 / (steps_per_mm * 10);  // default 10mm/sec
+
+int steps_per_mm = 50;  // full-stepping for GEN1
```
sfoster1: this can be unsigned

sfoster1 (changes requested):
Couple little changes and we need to test motion params on the gen2s.

sfoster1 (approved):
lgtm once tested

modules/mag-deck/mag-deck-arduino/mag-deck-arduino.ino
```diff
@@ -305,15 +308,22 @@ void activate_bootloader(){
...
+  int model_version = device_model.substring(MODEL_VER_TEMPLATE_LEN).toInt();
+  if (model_version >= 20) {  // Update steps_per_mm for GEN2 or later
+    steps_per_mm = 100;
+    step_delay_microseconds = 100000 / (steps_per_mm * 10);
+  }
+
   setup_pins();
   setup_digipot();
   set_current(CURRENT_HIGH);
   set_speed(SPEED_LOW);
```
sanni-t: I think we'll need to adjust the speeds (SPEED_HIGH, SPEED_LOW, SPEED_PROBE) too. The current speed values will make the motor move much faster. The speed settings should be half of GEN1 to keep the actual speeds same in both models.

sanni-t (changes requested):
Let's test the speeds

modules/mag-deck/mag-deck-arduino/mag-deck-arduino.ino
```diff
@@ -45,7 +49,7 @@ Memory memory = Memory();  // reads from EEPROM to find device's unique serial,
 
 #define CURRENT_TO_BYTES_FACTOR 114
 
-#define STEPS_PER_MM 50  // full-stepping
+#define STEPS_PER_MM ((model_version < 20) ? 50 : 100)  // full-stepping
```
sfoster1: Not a big fan of this approach, though it would work fine. It checks the value every single time it's called, and putting a ternary in a non-parameterized macro like this hides that there's logic involved.

Instead, let's make a new struct that pulls together all the motion params that we adjust here; let's instantiate one on the free store; and let's adjust its values depending on the model and check them there. This will also pull it all together nicely. Call it like MotionParams or something.

sfoster1 (changes requested):
A bit of code cleanup in the comment.

sfoster1 (approved):
This is looking good! runs well on a gen2

ahiuchingau:
Tested the new changes with @sanni-t on both Gen2 and "Gen1" module, worked well on both

sanni-t (approved):
✨🤩✨

Laura-Danielle (approved):
Updated firmware on my gen1 magnetic module, and ran a protocol.

Everything worked as expected.

PR #115, opened 2020-04-24
chore(eepromWriter): Enable the py script to upload just application f/w

sanni-t:
Refer [#5483](https://github.com/Opentrons/opentrons/issues/5483)
## overview

This PR adds arguments `write_serial` & `module` to specify whether to only upload application f/w or to upload serial number as well. This is necessary for updating our production SOP in a way that doesn't need the production team to use Arduino to upload firmware from a dev branch. 

## changes
- Added args to allow just app f/w upload
- Removed a duplicate key in models
- Updated file fetch so that the script can find the firmware files without having to run from inside the directory
- Mention requirement of avrdude 6.3+ . This is because the config file we're using is for v6.3+ and older versions don't work on MacOS 10.15+

### DIFF
--- a/eepromWriter/write_module_memory.py
+++ b/eepromWriter/write_module_memory.py
@@ -1,10 +1,10 @@
 '''
 This script takes in 2 args: `--write_serial` & `--module`
-    i) if `write_serial` is true: Upload Serial & model of a temp/mag-deck &
+    i) if `write_serial` is True: Upload Serial & model of a temp/mag-deck &
                                   upload the application firmware.
     ii) if `write_serial` is false: only upload the application firmware
 `--module`: specifies whether we're uploading to a 'tempdeck' or 'magdeck'.
-            Is needed only when `write_serial` is false.
+            Is needed only when `write_serial` is False or not specified.
 
 What the script does:
     1) discover port name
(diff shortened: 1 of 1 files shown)


eepromWriter/write_module_memory.py
```diff
@@ -180,37 +190,55 @@ def _parse_model_from_barcode(barcode):
     raise Exception(BAD_BARCODE_MESSAGE.format(barcode))
 
 
+def build_arg_parser():
+    arg_parser = ArgumentParser(
+        description="Firmware & serial uploader for Opentrons modules")
+    arg_parser.add_argument("--write_serial", required=True)
```
sfoster1: if you use `action='store_true'` you can later test for `args.write_serial` and it will be `True` if specified and `False` otherwise. Then the user does `--write_serial` (without an argument) when passing it on the commandline.

In addition, if you specify `help='some string'` then when you do `--help` it will print all the arguments with their help strings. Good for discoverability

sfoster1 (approved):
Couple things you could fix up in the argument parsing but it'll work

eepromWriter/write_module_memory.py
sanni-t: Good call

PR #122, opened 2020-09-14
feat(thermocycler): improve temp status response & debug stat response

### LINKED ISSUE #116: Thermocycler: add thermistor values to error message
## current behavior

When thermistor values aren't uniform, the thermocycler goes into error mode, starts flashing orange light and outputs a message `Plate temperature is not uniform`. 
This isn't too helpful in checking how non-uniform it is and determining which thermistor is problematic

## change proposed

Add values of all plate thermistors to the error message-
`Error: Plate temperature is not uniform. T1: 40, T2: 40.2, T3: 40.5, T4: 40, T5: 40.3, T6: 55`

sanni-t:
Closes #116

## overview

This PR adds improvements to TC responses. These have been a long pending features and with more TCs out in the field, is becoming more and more crucial to implement these in order to debug issues faster. More work to follow in the epic [Thermocycler: Detailed status and errors ](https://app.zenhub.com/workspaces/opentrons-hmg-5e1cdf98e04060788ad94e32/issues/opentrons/opentrons/6541)

## changes

- added two fields to temperature status response: 
  
  1. total hold time. A follow-up to [#6479](https://github.com/Opentrons/opentrons/pull/6479) 
  2. temperature is at target. This is a good-to-have field since it will give the driver/ API a good idea about whether temperature overshoot is over and the plate temperatures are stable/stabilizing near target. Right now, the only way to know that is by monitoring the plate temperatures over a specific time window and checking if they are in an acceptable temperature range. Though not a bad technique, it makes the api & firmware give different results for 'when' is the plate at target. 
  So, the response to `M105` will now look like `T:50 C:47 H:20 T_Hold:50 At_target?:0`
- moved the time gap check out of debug 'stat' response (which was there primarily for continuous
debug prints) so that a 'stat' command gets an immediate response.
- cleaned up redundant error messages
- added thermistor values to error messages. So an example error message would look like:
`Error:Plate temperature is not uniform. T1: 36.6511	T2: 41.5688	T3: 41.0032	T4: 41.5809	T5: 41.1606	T6: 40.8699	T.Lid: 22.9323	T.sink: 38.3661	T_error: 0.0000`

## questions

~Will the new error message with thermistor values be read & logged fully by the TCPoller or will it only read the first line and issue a pause?~ 
Answered in comment below. Made changes accordingly to make the message a one-liner

### DIFF
--- a/modules/thermo-cycler/thermo-cycler-arduino/gcode.cpp
+++ b/modules/thermo-cycler/thermo-cycler-arduino/gcode.cpp
@@ -161,15 +161,20 @@ void GcodeHandler::device_info_response(String serial, String model, String vers
   Serial.println();
 }
 
-void GcodeHandler::targetting_temperature_response(float target_temp,
-                                    float current_temp, float time_remaining)
+void GcodeHandler::targetting_temperature_response(float target_temp, float current_temp,
+                                                   float time_remaining,  float total_hold_time,
+                                                   bool at_target)
 {
   Serial.print(F("T:"));
   Serial.print(target_temp, SERIAL_DIGITS_IN_RESPONSE);
   Serial.print(F(" C:"));
   Serial.print(current_temp, SERIAL_DIGITS_IN_RESPONSE);
   Serial.print(F(" H:"));
-  Serial.println((unsigned int)time_remaining);
+  Serial.print((unsigned int)time_remaining);
+  Serial.print(F(" Total_H:"));
+  Serial.print((unsigned int)total_hold_time);
+  Serial.print(F(" At_target?:"));
+  Serial.println(at_target);
 }
 
 void GcodeHandler::targetting_temperature_response(float target_temp,
@@ -195,6 +200,7 @@ void GcodeHandler::idle_lid_temperature_response(float current_temp)
   Serial.print(F(" C:"));
   Serial.println(current_temp, SERIAL_DIGITS_IN_RESPONSE);
 }
+
 void GcodeHandler::response(String param, String msg)
 {
   Serial.print(param);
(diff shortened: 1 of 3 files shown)


sfoster1:
> Closes #116
> ## overview
> 
> This PR adds improvements to TC responses. These have been a long pending features and with more TCs out in the field, is becoming more and more crucial to implement these in order to debug issues faster. API work to follow: [#6538](https://github.com/Opentrons/opentrons/issues/6538), [#6537](https://github.com/Opentrons/opentrons/issues/6537)
> ## changes
> 
>     * added two fields to temperature status response:
>       
>       1. total hold time
>       2. temperature is at target.
> 
>     * moved the time gap check out of debug 'stat' response (which was there primarily for continuous
>       debug prints) so that a 'stat' command gets an immediate response.
> 
>     * added thermistor values to error messages
> 
>     * cleaned up redundant error messages
> 
> 
> ## questions
> 
> Will the new error message with thermistor values be read & logged fully by the TCPoller or will it only read the first line and issue a pause?

It will probably just read the first message. Those pollers really aren't set up to read multiline responses and i'd recommend against it.

We're also getting to the point where we really need a more complex message passing protocol than plain text, or ideally than gcode. That's not something we could necessarily do now but for now, putting things on one line and establishing precedent for labeling fields could help.

modules/thermo-cycler/thermo-cycler-arduino/gcode.cpp
```diff
@@ -211,7 +217,7 @@ void GcodeHandler::add_debug_response(String param, float val)
 {
   Serial.print(param);
   Serial.print(F(": "));
-  Serial.print(val);
+  Serial.print(val, 4);
```
b-cooper: Is there a reason that this isn't `SERIAL_DIGITS_IN_RESPONSE` (which I believe equals 4), instead of this magic number ?

b-cooper (approved):
Code looks good, still have to test on TC. 🏯

sfoster1 (approved):
LGTM pending test

sanni-t:
Tested on a thermocycler & api 3.20.1 robot. The responses all updated well and the existing driver works with new response fields.

PR #125, opened 2021-02-03
chore(build): Switch to cmake as a configuration system

sfoster1:
This PR switches away from raw makefiles as a configuration and build system and uses cmake instead.

# Why?

We're about to enter territory that would be remarkably painful to tread upon with `make`: for upcoming module projects, we'll compile both for the target (probably with gcc) and the host (probably with clang); we'll also use other tools like clang sanitizers. We'll build in testing, which means - for native code projects - compiling different executables. All of this is hard to do in make, and the ways that you work around it typically mean abandoning the places where Make can help you the most.

Raw makefiles are pretty good at replacing shell scripts as a build system. They give you dependency management, they give you some built-ins for managing compiler options for compiler options. The farther you go from using them as a replacement for shell scripts, the worse they get. In particular, they really start to drag when the _configuration_ of a build becomes more complex, and especially when there are multiple complex configurations. The typical way to avoid this in raw Make is to start avoiding using Make's builtin rules (so you don't have to, e.g., switch around $CC) or have increasingly complex switches around environment variables. 

The other route that people take is to use a config system. There are a lot of these: [kconfig](https://www.kernel.org/doc/html/latest/kbuild/kconfig-language.html), buildroot's [config](https://buildroot.org/downloads/manual/manual.html#configure) which somehow also runs under make in bizarre ways, IDE project configuration (which therefore ties you to an IDE), others.

But there's also [cmake](https://cmake.org/). Cmake is an ugly (just check out the code!) configuration system to look at, and its overeager attempts to help you out can occasionally get in your way, but it is fundamentally a solid, well-designed, [extremely well-documented](https://cmake.org/cmake/help/v3.19/) opinionated multi-platform native code configuration system that is widely used, integrates well with IDEs, and fixes the issues I noted above.

## OK but why change what works?

Well, unfortunately, cmake's core desire is to compute a bunch of configuration and emit... Makefiles. Which means it's very hard to use alongside a different Makefile based system. Also I needed the practice, quite honestly.

# So what changed?

`opentrons-modules` in this pr uses cmake as a configuration system, generating (currently) makefiles.

That means that the logic that used to be in the top-level `Makefile` is now spread out in the `CMakeLists.txt` files that exist in the source tree. Each file is responsible for building what is below it, in the parallel builds directory.

# How do I use this?
(Check out the README)

The biggest usability downside of cmake is that it has two steps: configure, and build. You can do them both with the `cmake` command, which is your friend. The specific cmake you want to use is CMake 3.19, which you can find [here](https://cmake.org/download/). We want 3.19, which is the latest release, because of the new "presets" concept which allows much easier (MUCH easier) configuration of things like cross-compilers.

So first you configure: `cmake .`
And then you build: `cmake --build ./builds`

You can also use the generated Makefiles to build - `cd builds && make`.

To bundle up all the modules into a specific package, run `cmake --install ./builds`.

When you run the configuration step, cmake will automatically download the arduino IDE and configure it for you.

# What else do we need to do with this?
- [ ] Probably a good idea to upload the build results onto actual modules
- [x] Test on OSX (works great!)


ahiuchingau (approved):
uploaded firmware on a magdeck successfully 🔥

Laura-Danielle (approved):
Tested on Mac OS and successfully updated firmware on a temperature module after using cmake to build everything.

Something we may want to consider removing in the future is the `.python-version` file in the repo. I'm not sure why it is there, and it gave me trouble when I was trying to run the `write_module_memory` script.

README.md
```diff
@@ -9,77 +9,41 @@ Magnetic Module (Magdeck) and Thermocycler.
...
-Select the correct board file for your project from _Tools->board_ (Opentrons Thermocycler M0/ Opentrons Magdeck/ Opentrons Tempdeck)
-
-* NOTE: For Opentrons Thermocycler M0, you will have to set the appropriate flags in the `platform.local.txt` file: Find this file in your Arduino15 folder- a hidden folder usually located in C://Users/username/appData/Local on Windows. You will find the preferences file in `..Arduino15/packages/Opentrons/hardware/samd/1.0.1`. Paste this line in the file above: `compiler.cpp.extra_flags=-DLID_WARNING=false -DHFQ_PWM=false -DHW_VERSION=4 -DLID_TESTING=false -DRGBW_NEO=true -DVOLUME_DEPENDENCY=true -DTC_FW_VERSION="<YOUR_FW_VERSION_STRING>"` (set appropriate flag values)
-
-Then select _Sketch->Compile_.
-
-The build will be saved in a windows temp directory. You can check verbose compilation
-output for the temp directory path. To turn ON verbose mode, go to _File->Preferences_
-and select _compilation_ option for _'Show verbose output during'_
+If you run `cmake --install`, all the module firmwares will be built and zipped up with their supporting
```
sanni-t: Looks like this should be `cmake --install ./build`? 
Also, looks like --install only copies the zip files generated by `cmake --build ./build --target zip-all` into `dist` so you actually need to `zip-all` before doing a `--install`?

sanni-t (changes requested):

README.md
sfoster1: Oh, gross. I'll change that.

sanni-t (approved):
Looks great! Thanks for the review session @sfoster1.
Tested building thermocycler binary and uploading it to a thermocycler. All flags updated correctly.

PR #127, opened 2021-02-19
chore(stm32): add linters, checkers, tests

sfoster1:
Adds build infrastructure and minimal examples for
- Code formatting using [clang-format](https://clang.llvm.org/docs/ClangFormat.html), based on the [google c++ style guide](https://google.github.io/styleguide/cppguide.html) with some little modifications like 4-space stops instead of 2 because wtf. Available as `cmake --build ./build-stm32-cross --target heater-shaker-format`; checked in ci. Also builds in `all`.
- Code linting using [clang-tidy](https://clang.llvm.org/extra/clang-tidy/), which checks for a fun variety of common bugs and issues, available in `all` and as `heater-shaker-lint`. Checked in CI.
- Tests! And finally the first host build for these modules. Uses [catch2](https://github.com/catchorg/Catch2/tree/v2.13.4), a modern header-only C++ testing library that uses BDD-style test definitions (Given, Then, etc). Since these need a different compiler, they're build and run with `cmake --preset=stm32-host .` and then `cmake --build ./build-stm32-host --target check`. Don't use the `test` target because that doesn't actually build the tests.
- Also sets your normal sort of compiler options (wall, werror, so on)

As with the cross-compilation setup, all the dependencies are downloaded automatically with 2 exceptions:
- clang: since many systems (including OSX) come with clang, and since host compilers are typically heavily integrated into the host system particularly on linux, there's an out provided: if you have a clang 11 on your path already, we'll just use that. If you don't, we try and download it, but especially on non-ubuntu linux setups this may not work so well. 
- host gcc: same deal. Systems have it, it doesn't really have official binary distributions because of this, so let's just find it.


sanni-t (approved):
Works as expected!

Laura-Danielle (approved):
Tested and lgtm

sanni-t (commented):
`--target check` works well! 
I think we should name it something else though, since we use `make check` for format checking in all our other codebases.

Something like `build-and-test` or `tests-check`?

sfoster1:
> `--target check` works well!
> I think we should name it something else though, since we use `make check` for format checking in all our other codebases.
> 
> Something like `build-and-test` or `tests-check`?

`build-and-test` sounds good. `check` is apparently the target generated by gnu autotools to do this task but we have never and will never use that so who cares.

ahiuchingau (approved):
Ran the `build-and-test` target and worked as expected.

README.md
```diff
@@ -51,6 +51,14 @@ Open the `/modules/.../<file_name>.ino` file in Arduino, select the appropriate
 
 The stm32 modules have debug harnessing built in; for instance, running `cmake --build ./build-stm32-cross --target heater-shaker-debug` will build the firmware and spin up the cross gdb and openocd. This target by default (because of commands in the gdbinit) will reset the board and upload the built firmware; if you want to use the firmware currently flashed to the board, run everything manually.
 
+## test
+
+This is not implemented for the arduino modules.
+
+For the STM32 modules, if you configure the host toolchain: `cmake --preset=stm32-host .`, you can then run the `check` target: `cmake --build ./build-stm32-host --target build-and-test`, which will build the tests and then run them. You can also use cmake's built-in `test` target: `cmake --build ./build-stm32-host --target test`, but this will not automatically actually build the tests.
```
ahiuchingau: Should this be the `build-and-test` target now that it has been renamed?
```suggestion
For the STM32 modules, if you configure the host toolchain: `cmake --preset=stm32-host .`, you can then run the `build-and-test` target: `cmake --build ./build-stm32-host --target build-and-test`, which will build the tests and then run them. You can also use cmake's built-in `test` target: `cmake --build ./build-stm32-host --target test`, but this will not automatically actually build the tests.
```

PR #129, opened 2021-03-11
feat(heater-shaker): FreeRTOS integration and tasks setup

sfoster1:
This PR adds [FreeRTOS](https://www.freertos.org), a widely available real-time operating system. It uses the version in the STM32 BSP.

The RTOS is integrated by changes in the new interrupts file, which segregates interrupt entry code into its own TU, and links the SysTick, PendSV, and SVC ISRs to the required entrypoints in the RTOS code, which give the RTOS the ability to schedule tasks.

Once that's done, we actually make some tasks and the first entries in the host/cross-agnostic compilation HAL. Specifically,

- We have four tasks
- Each task has a portable segment, in `include/heater-shaker`, that will include as much of the task logic as possible
- Tasks communicate with queues using well-defined messages (through `std::variant`). Since we don't want to host-compile FreeRTOS, the portable task segment is defined as a template class which can be provided with any queue implementation matching the queue concept in `include/heater-shaker/message_queue.hpp`.
- There is a `FreeRTOSQueue` implementation that lives in `include/firmware/freertos_message_queue.hpp`, and FreeRTOS task definitions in `include/firmware/freertos_xxx_task.hpp` and `firmware/freertos_xxx_task.cpp`. These are separate and important because in addition to defining the low-level task logic as it interfaces with FreeRTOS, they are also where the statically-defined storage for 
   - the freertos internals backing store
   - the task stacks
   - the message queue control structures
   - the message queue backing stores

Then, there's an implementation of the message queues for testing which just uses `std::deque` as a backing store, and a basic test that the tasks can be spun up in testing (there's no logic to test yet).

This PR also switches to github actions.


stm32-modules/heater-shaker/include/heater-shaker/task_concept.hpp
```diff
@@ -0,0 +1,4 @@
+#ifndef __TASK_CONCEPT_H_
+#define __TASK_CONCEPT_H_
+
+#endif  // __TASK_CONCEPT_H_
```
Laura-Danielle, line 4: Should this be commented out?

Laura-Danielle (approved):
Just one comment but otherwise it looks good to me

stm32-modules/heater-shaker/include/heater-shaker/task_concept.hpp
sfoster1: yeah, that's a note that says what the endif is ending. Honestly that file should probably not be checked in, whoops!

PR #130, opened 2021-04-02
feat(heater-shaker): add generic gcode parser

sfoster1:
Add a generic (limited, single-level) parser-combinator style gcode
parser using templates.

This parser (in gcode_parser.hpp) parses gcodes, one at a time, out of a
"complete" gcode command string - a command string that ends in a
newline delimiter and therefore represents upstream's intent that the
command is complete.

While the parser isn't yet instantiated in the application code, because
there's not really anything that can feed it data yet, it is fully
tested and ready for use.

### DIFF
--- a/stm32-modules/heater-shaker/include/heater-shaker/gcode_parser.hpp
+++ b/stm32-modules/heater-shaker/include/heater-shaker/gcode_parser.hpp
@@ -1,0 +1,228 @@
+/*
+** gcode_parser - a templated gcode parser for use anywhere
+*/
+#pragma once
+
+#include <stdio.h>
+
+#include <algorithm>
+#include <cctype>
+#include <charconv>
+#include <iterator>
+#include <optional>
+#include <string>
+#include <utility>
+#include <variant>
+
+namespace gcode {
+
+/*
+ * gcode::gobble_whitespace is a convenience function to strip out any
+ * whitespace from an input; the returned iterator is after any whitespace at
+ * the head of the input.
+ */
+template <typename Input, typename Limit>
+requires std::forward_iterator<Input>&&
+    std::sized_sentinel_for<Limit, Input> auto
+    gobble_whitespace(const Input& start_from, const Limit stop_at) -> Input {
+    // this has to be a lambda because std::isspace's prototype does not
+    // interact well with template argument deduction, sadly
+    return std::find_if_not(
+        start_from, stop_at,
+        [](const unsigned char c) -> bool { return std::isspace(c); });
+}
+
+/*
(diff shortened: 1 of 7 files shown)


stm32-modules/heater-shaker/include/heater-shaker/gcode_parser.hpp
```diff
@@ -0,0 +1,228 @@
...
+#include <algorithm>
+#include <cctype>
+#include <charconv>
+#include <iterator>
+#include <optional>
+#include <string>
+#include <utility>
+#include <variant>
+
+namespace gcode {
```
Laura-Danielle, line 17: Just so I have this straight in my head:
- This is just a template for generic gcode parsing. We will probably need to add more specific implementation for the heater shaker and that would live in a cpp file, right?

Laura-Danielle (approved):
I just had one question, but the code looks good.

stm32-modules/heater-shaker/include/heater-shaker/gcode_parser.hpp
sfoster1: No not really; what will happen is the host comms task core will use the individual gcode parsers implemented in gcode.hpp and instantiate the template there:

```
void HostCommsTask::handle_message(const Message& message) {
    ....
    auto parser = gcode::GroupParser<gcode::M3, gcode::M115, gcode::...>();

    auto result = std::make_pair(parser::ParseResult(std::monostate), message.cbegin());
    while (true) {
      result = parser.parse_available(result.second, message.cend());
      if (std::holds_alternative<std::monostate>(result.first()) {
          break;
      } else {
         build_and_dispatch(result.first);
      }
    } 
```

or something similar

stm32-modules/heater-shaker/tests/test_gcodes.cpp
```diff
@@ -0,0 +1,358 @@
+#include <array>
+
+#include "catch2/catch.hpp"
+#include "heater-shaker/gcodes.hpp"
+
+SCENARIO("SetRPM parser works") {
```
ahiuchingau, line 6: Wow these tests are very thorough!

Out of curiosity, I wonder if there's a way to combine some of these tests, like similar to the Pytest parametrize operator?

ahiuchingau (approved):
The comments are so helpful - code looks good to me!

stm32-modules/heater-shaker/tests/test_gcodes.cpp
sfoster1: There almost certainly is. Whenever we add more gcodes we should look into it.

PR #131, opened 2021-04-06
feat(heater-shaker): USB bringup

sfoster1:
This PR
- Slightly refactors the cmake BSP integration to have separate components
- Brings up the USB on the heater/shaker with a simple loopback application integrated into the system
- Adds some python utilities to examine peripheral control registers through gdb in a way other than doing the math for the address every time

## CMake/BSP

The STM32F303BSP package has drivers and also a bunch of middlewares, like USB support (which claims to be hal agnostic but actually isn't) and freertos (which actually is, more or less). While it's generally not going to be useful to have USB support or FreeRTOS without the rest of the drivers, it's certainly reasonable to want just the drivers, or the drivers and the rtos but not USB, or drivers and USB but not rtos, etc etc (ok those are the only possibilities right now but we might add more).

To that end, cmake packages can contain components. In true find-packages, where you're finding packages that already exist on the system, this is usually done by separating components with `::` like in C++ for some reason, but that's only for things that are actually pre-compiled (`INTERFACE` targets). Here, we just separate them with `_`. 

So you now do `find_package(STM32F303BSP COMPONENTS Drivers USB FreeRTOS)`, or `find_package(STM32F303BSP COMPONENTS Drivers)` and you by and large get what you ask for.

## GDB/register examination

The way peripheral registers work on embedded targets is that certain memory accesses actually touch the registers, and you have to know what registers map where. You look it up in the manual as a human; and as a programmer, you use the device header. But the device header provides preprocessor defines for those addresses, and by the time you give gdb an executable to run, they're gone. So you can't do `p NVIC->ITAR0` or whatever, because there's no symbol - that was a define, and the compiler ate it.

The way to handle this is with debugger extensions; this is what a lot of the open-source IDEs that are driven by GDB do. This is helped by ARM getting CMSIS implementers to publish parsable files of their register layouts, and those are available in python; and GDB can load python programs! So we can install some python dependencies like https://github.com/1udo6arre/svd-tools and modify the `gdbinit` a bit and get the ability to `svd get nvic itar0` and get a lovely table.

Downside: GDB running python is _extremely_ hostile to just about every single python management tool. It ignores virtualenvs, it ignores stuff like pyenv. It defaults to python2. So the support for this, which is in `FindGDBSVDTools.cmake` and in `stm32-modules/heater-shaker/firmware/gdbinit.template` is kind of hacky:
- we just download the modules from pypa or github or wherever and smash em in directories
- the gdbinit uses cmake templating capability to just pop those directories into `sys.path` of whichever random python interpreter gdb picks up on

That said... it works! You can run `heater-shaker-debug` on a nucleo or whatever and do `svd get nvic` and get a lovely pretty-printed table of the register statuses there.

## USB

Oh also USB works now. This uses the USB middlewares now available in the BSP to expose a CDC interface that looks like a terminal. Right now, it just echoes back whatever you type. You can test this on an actual board or by programming the nucleo once then swapping its usb headers over to give the application chip control of the USB (this will be very annoying though, might not be testable on a nucleo realistically).

There's an awful lot of code to deal with this, mostly copied from templates in the middleware and the example projects. Specifically,

- `stm32-modules/heater-shaker/firmware/usbd_conf.(h,c)` is modifications of a template from an example nucleo DFU project. This configures the link between the USB middleware and the specific stm32 hal it's designed for; it is not actually hal-agnostic.
- `stm32-modules/heater-shaker/usb_desc.(h,c)` is modifications of a template from the middleware itself and contains all the USB string descriptors and for some reason some routines to convert them to UTF-16 or whatever it is. This is also where the PID (which is ascii for HS because I'm soooo clever) lives.
- `stm32-modules/heater-shaker/firmware/freertos_comms_task.cpp` is where the actual integration of usb into the rest of the firmware lives. The callbacks here come out of the cdc middleware, and the two main things are 1) the freertos task polling `USBD_CDC_ReceivePacket` in its mainloop and 2) the callback `CDC_Receive` at the bottom of the file, which is where incoming data goes and where we implement the loopback. At some point this needs to turn into submitting the received data into a freertos queue for processing in the main task loop, so the main task loop can look like a sleep until there's data, then process the data, then set up to receive again and transmit if it needs to. Also maybe we want DMA because there's a suspicious amount of copying data in what may be an interrupt context.

# Review Requests
Look it over, test it if you have a board, ask questions.


Laura-Danielle (approved):
lgtm

PR #132, opened 2021-04-07
feat(heater-shaker): add tested PID calculator

sfoster1:
This PR adds a PID calculator integration. This is the core of a PID
controller. It computes a control output based on an error input,
certain historical data (for the integral and derivative term) and
pre-configured coefficients.

In addition, the heater-shaker src/ directory is now compiled as a
static library that is linked into the firmware and the tests,
respectively. Having the core in a separate static library continues to
enforce the need to keep code in the core separate from the firmware.


Laura-Danielle (approved):
lgtm

PR #133, opened 2021-04-09
chore(stm32): add CMake custom config example

ahiuchingau:
If you are using macOS and depending on the version, the Clang may or may not support Concepts. This can cause an error  when you run the stm32 module tests. To solve it, you need to (1) install `gcc@10` using homebrew, and then configure a `CMakeUserPresets.json` at the root of the repo to use the installed gcc as your compiler.

### Changelog
- update readme
- add an example of the preset file

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ If you are running on macOS, you might run into this error:
 error: 'concepts' file not found
 ```
 
-This means that the clang verison on your system does not support Concepts. You will need to do `brew install gcc@10` and copy the `examples/CMakeUserPresets.json` to the root directory. You can then specify this custom configration by using `cmake --preset=stm32-host-gcc`.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -59,6 +59,16 @@ For the STM32 modules, if you configure the host toolchain: `cmake --preset=stm3
...
 
+### Custom Configuration
+
+If you are running on macOS, you might run into this error: 
+
+```
+error: 'concepts' file not found
+```
+
+This means that the clang verison on your system does not support Concepts. You will need to do `brew install gcc@10` and copy the `examples/CMakeUserPresets.json` to the root directory. You can then specify this custom configration by using `cmake --preset=stm32-host-gcc`.
```
sfoster1:
```suggestion
This means that the clang version on your system does not support Concepts. You will need to do `brew install gcc@10` and copy the `examples/CMakeUserPresets.json` to the root directory. You can then specify this custom configration by using `cmake --preset=stm32-host-gcc`.
```

sfoster1 (approved):
Nice, thanks!

PR #134, opened 2021-04-15
feat(heater-shaker): bring up message interaction via USB serial

sfoster1:
This (big - should I split it up?) PR adds everything necessary to not just be a simple loopback on USB, but actually
- Listen to a host when connected
- Parse incoming messages for GCode
- Translate those messages to internal messages to send to other tasks, returning errors if necessary
- Send the messages to other tasks
- Have those other tasks response
- Eventually respond with the data or acknowledgements from the other tasks

This therefore required a lot of work in better defining the messages to be passed; handling the parsing, and the message handling in all of the relevant tasks; and one million little fixes on the USB side of things for problems that had been masked by the system actually trying to be a loopback.

The commits in the PR have been pretty closely janitored and have messages proportionate to their complexity, so I'd recommend browsing this PR by commit rather than looking at all the file changes at once.

## New Functionality
- Responds to M3 (set-spindle-speed, `M3 Snnnn\n`), M123 (get status, `M123\n` returns `M123 Cnnnn Tnnnn OK\n` for current and target rpm; these are hardcoded for now), M104 (set temperature, `M104 Snnn\n`), and M105 (get temperature, `M105\n` returns `M105 Cnnn Tnnn OK\n`, hardcoded for now like RPM)
- Has errors for if you spam it too much (a human will not be able to do this, a python script might)
- Has tests for all the above including a full roundtrip test from string to string implying the message roundtrips

## Testing
- Put it on an NFF probably, or hack together some nucleos to get a programmable micro that owns a USB 
- Plug it in to your computer
- Connect to it with your serial interface of choice; as a longtime serial terminal sommelier, I'd recommend `python -m serial.tools.miniterm /dev/ttywhatever 115200`
- _Marvel_ at the lack of kernel logs complaining about not being able to set flow control
- _ooh_ at being able to type `M105\n` and get a lovely response back,  
- _aahhh_ at being able to type a bunch of random stuff and hit enter and get a nicely formatted error with number back
- _wail_ at the code in `freertos_comms_task.cpp`. I apologize for nothing, there are no sins, laws, or morals in USB

### DIFF
--- a/stm32-modules/heater-shaker/firmware/freertos_comms_task.cpp
+++ b/stm32-modules/heater-shaker/firmware/freertos_comms_task.cpp
@@ -18,15 +18,19 @@
 #pragma GCC diagnostic pop
 
 #include "firmware/freertos_message_queue.hpp"
+#include "hal/double_buffer.hpp"
 #include "heater-shaker/host_comms_task.hpp"
+#include "heater-shaker/messages.hpp"
 #include "heater-shaker/tasks.hpp"
 
 struct CommsTaskFreeRTOS {
     USBD_CDC_ItfTypeDef cdc_class_fops;
     USBD_HandleTypeDef usb_handle;
     USBD_CDC_LineCodingTypeDef linecoding;
-    std::array<uint8_t, CDC_DATA_HS_MAX_PACKET_SIZE * 4> rx_buf;
-    std::array<uint8_t, CDC_DATA_HS_MAX_PACKET_SIZE * 4> tx_buf;
+
+    double_buffer::DoubleBuffer<char, CDC_DATA_HS_MAX_PACKET_SIZE * 4> rx_buf;
+    double_buffer::DoubleBuffer<char, CDC_DATA_HS_MAX_PACKET_SIZE * 4> tx_buf;
+    char *committed_rx_buf_ptr;
 };
 
 static auto CDC_Init() -> int8_t;
@@ -67,7 +71,7 @@ static CommsTaskFreeRTOS _local_task = {
          .datatype = 0x08},
     .rx_buf = {},
     .tx_buf = {},
-};
+    .committed_rx_buf_ptr = nullptr};
 
 // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
 static auto _top_task = host_comms_task::HostCommsTask(_comms_queue);
@@ -83,20 +87,50 @@ static std::array<StackType_t, stack_size> stack;
 static StaticTask_t
     data;  // NOLINT(cppcoreguidelines-avoid-non-const-global-variables)
 
+extern "C" {
+extern __ALIGN_BEGIN uint8_t
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
+    USBD_CDC_CfgHSDesc[USB_CDC_CONFIG_DESC_SIZ] __ALIGN_END;
+extern __ALIGN_BEGIN uint8_t
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
+    USBD_CDC_CfgFSDesc[USB_CDC_CONFIG_DESC_SIZ] __ALIGN_END;
+}
+
 // Actual function that runs in the task
 void run(void *param) {  // NOLINT(misc-unused-parameters)
-    static constexpr uint32_t delay_ticks = 1;
     auto *task_pair = static_cast<decltype(_tasks) *>(param);
     auto *local_task = task_pair->second;
+    auto *top_task = task_pair->first;
+    // This clears the capability bit that would be other sent upstream
+    // indicating we handle flow control line setting from host, which we don't,
+    // which leads to delays and annoying kernel messages. See
+    // stm32f303-bsp/Middlewares/ST/STM32_USB_Device_Library/Class/CDC/Src/usbd_cdc.c:174
+    // for annotated descriptor definitions
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-magic-numbers)
+    USBD_CDC_CfgHSDesc[30] = 0;
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-magic-numbers)
+    USBD_CDC_CfgFSDesc[30] = 0;
     USBD_Init(&local_task->usb_handle, &CDC_Desc, 0);
     USBD_RegisterClass(&local_task->usb_handle, USBD_CDC_CLASS);
     USBD_CDC_RegisterInterface(&local_task->usb_handle,
                                &local_task->cdc_class_fops);
     USBD_SetClassConfig(&local_task->usb_handle, 0);
     USBD_Start(&local_task->usb_handle);
+    local_task->committed_rx_buf_ptr = local_task->rx_buf.committed()->data();
     while (true) {
-        USBD_CDC_ReceivePacket(&local_task->usb_handle);
-        vTaskDelay(delay_ticks);
+        auto length_to_xmit =
+            top_task->run_once(local_task->tx_buf.accessible()->data(),
+                               local_task->tx_buf.accessible()->size());
+        if (length_to_xmit != 0) {
+            local_task->tx_buf.swap();
+            USBD_CDC_SetTxBuffer(
+                &_local_task.usb_handle,
+                // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
+                reinterpret_cast<uint8_t *>(
+                    local_task->tx_buf.committed()->data()),
+                length_to_xmit);
+            USBD_CDC_TransmitPacket(&_local_task.usb_handle);
+        }
     }
 }
 
@@ -114,7 +148,10 @@ auto start()
 
 static auto CDC_Init() -> int8_t {
     using namespace host_comms_control_task;
-    USBD_CDC_SetRxBuffer(&_local_task.usb_handle, _local_task.rx_buf.data());
+    USBD_CDC_SetRxBuffer(
+        &_local_task.usb_handle,
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
+        reinterpret_cast<uint8_t *>(_local_task.rx_buf.committed()->data()));
     return (0);
 }
 
@@ -122,6 +159,8 @@ static auto CDC_DeInit() -> int8_t {
     /*
        Add your deinitialization code here
     */
+    using namespace host_comms_control_task;
+    _local_task.committed_rx_buf_ptr = _local_task.rx_buf.committed()->data();
     return (0);
 }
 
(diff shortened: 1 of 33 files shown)


stm32-modules/heater-shaker/include/heater-shaker/errors.hpp
```diff
@@ -0,0 +1,56 @@
...
+    auto* wrote_to = std::copy(
+        ErrorType::errorstring(),
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic)
+        ErrorType::errorstring() +
+            std::min(strlen(ErrorType::errorstring()), available),
+        buf);
+    return &(*wrote_to) - buf;
+}
+struct USBTXBufOverrun {
+    static constexpr auto errorstring() -> const char* {
```
sfoster1: please don't make us remember to make the numebrs the same

stm32-modules/heater-shaker/include/heater-shaker/ack_cache.hpp
```diff
@@ -0,0 +1,69 @@
...
+    template <typename ContentElement>
+    auto add(const ContentElement& element) -> uint32_t {
+        for (auto cache_element = cache.begin(); cache_element != cache.end();
+             cache_element++) {
+            if (std::holds_alternative<std::monostate>(
+                    cache_element->contents)) {
+                cache_element->contents = Payload(element);
+                cache_element->id = next_id;
+                next_id++;
+                if (next_id == 0) {
```
sfoster1, line 31: consider using the `std::monostate` present to indicate "empty"

stm32-modules/heater-shaker/include/heater-shaker/host_comms_task.hpp
```diff
@@ -75,7 +75,11 @@ requires MessageQueue<QueueImpl<Message>, Message> class HostCommsTask {
      *
      * This function returns the amount of data it actually wrote into tx_into.
      **/
-    auto run_once(char* tx_into, size_t tx_length_available) -> size_t {
+
+    template <typename InputIt, typename InputLimit>
+    requires std::forward_iterator<InputIt>&&
+        std::sized_sentinel_for<InputLimit, InputIt> auto
+        run_once(InputIt tx_into, InputLimit tx_limit) -> InputLimit {
```
sfoster1, line 82: return type should be InputIterator

stm32-modules/heater-shaker/include/hal/message_queue.hpp
```diff
@@ -19,10 +19,17 @@ concept MessageQueue = requires(MQ mq, MessageType mt, const MQ cmq,
     ->std::same_as<bool>;
     { mq.try_send(cmt) }
     ->std::same_as<bool>;
+    // Queues must have a receive that takes a message instance by ptr
```
Laura-Danielle, line 22: I'm still a little confused as to why we have both a receive and a try/receive + what the different use cases would be. By instinct, I would assume the message queue would only need a try receive.

stm32-modules/heater-shaker/include/heater-shaker/errors.hpp
```diff
@@ -0,0 +1,56 @@
...
+    }
+    static constexpr auto code() -> uint32_t { return 1; }
+};
+struct InternalQueueFull {
+    static constexpr auto errorstring() -> const char* {
+        return "ERR002:internal queue full\n";
+    }
+    static constexpr auto code() -> uint32_t { return 2; }
+};
+struct UnhandledGCode {
```
Laura-Danielle: Would unhandled Gcode be returned for use-cases like sending an unsupported temperature, or will we have more specific errors in that case?

stm32-modules/heater-shaker/include/heater-shaker/ack_cache.hpp
Laura-Danielle: Will you be making this change in a later PR?

stm32-modules/heater-shaker/include/heater-shaker/ack_cache.hpp
```diff
@@ -0,0 +1,69 @@
...
+    auto clear() -> void {
+        for (auto& cache_element : cache) {
+            cache_element.contents = std::monostate();
+            cache_element.id = 0;
+        }
+    }
+
+  private:
+    // Present only for testing; do not use
+    friend class _AckCacheTestHook;
```
Laura-Danielle, line 62: love the `friend class` name, and also did not realize cpp had a concept like this.

stm32-modules/heater-shaker/include/heater-shaker/gcodes.hpp
```diff
@@ -59,6 +71,15 @@ struct SetTemperature {
     using ParseResult = std::optional<SetTemperature>;
     float temperature;
 
+    static auto write_response_into(char* buf, size_t available) -> size_t {
```
Laura-Danielle: I'm curious if there is a way (or whether it makes sense) to make `write_response_into` be more generic/shared among the structs. For any structs that require additional parameters in their response, maybe we could do a separate function which appends those parameters.

stm32-modules/heater-shaker/firmware/freertos_comms_task.cpp
```diff
@@ -83,20 +87,50 @@ static std::array<StackType_t, stack_size> stack;
...
+    local_task->committed_rx_buf_ptr = local_task->rx_buf.committed()->data();
     while (true) {
-        USBD_CDC_ReceivePacket(&local_task->usb_handle);
-        vTaskDelay(delay_ticks);
+        auto length_to_xmit =
+            top_task->run_once(local_task->tx_buf.accessible()->data(),
+                               local_task->tx_buf.accessible()->size());
+        if (length_to_xmit != 0) {
+            local_task->tx_buf.swap();
+            USBD_CDC_SetTxBuffer(
```
Laura-Danielle, line 126: I'm having trouble finding this function and `USBD_CDC_TrasmitPacket`

stm32-modules/heater-shaker/include/heater-shaker/host_comms_task.hpp
```diff
@@ -39,9 +60,306 @@ requires MessageQueue<QueueImpl<Message>, Message> class HostCommsTask {
...
+                message, TICKS_TO_WAIT_ON_SEND)) {
+            auto written = errors::write_into<errors::InternalQueueFull>(
+                tx_into, tx_length_available);
+            ack_only_cache.remove_if_present(id);
+            return std::make_pair(false, written);
+        }
+        return std::make_pair(true, 0);
+    }
+
+    auto visit_gcode(const gcode::GetTemperature& gcode, char* tx_into,
```
Laura-Danielle: cool use of overloads

stm32-modules/heater-shaker/include/heater-shaker/motor_task.hpp
```diff
@@ -32,7 +33,41 @@ requires MessageQueue<QueueImpl<Message>, Message> class MotorTask {
...
+            messages::AcknowledgePrevious{.responding_to_id = msg.id};
+        static_cast<void>(task_registry->comms.get_message_queue().try_send(
+            messages::HostCommsMessage(response)));
+    }
+
+    auto visit_message(const messages::GetRPMMessage& msg) -> void {
+        auto response = messages::GetRPMResponse{
+            .responding_to_id = msg.id,
+            .current_rpm =
+                1050,  // NOLINT(cppcoreguidelines-avoid-magic-numbers)
```
Laura-Danielle, line 65: Are these just dummy values?

stm32-modules/heater-shaker/include/heater-shaker/gcodes.hpp
```diff
@@ -116,48 +115,37 @@ struct GetTemperature {
     */
     using ParseResult = std::optional<GetTemperature>;
 
-    static auto write_response_into(char* buf, size_t available,
-                                    uint32_t current_temperature,
-                                    uint32_t setpoint_temperature) -> size_t {
+    template <typename InputIt, typename InLimit>
```
Laura-Danielle, line 118: Why are we changing these to templates?

Laura-Danielle (approved):
Looks good to me overall, just had some random comments.

stm32-modules/heater-shaker/include/heater-shaker/ack_cache.hpp
sfoster1: i would like to get this merged because it works and is tested, yeah

stm32-modules/heater-shaker/include/hal/message_queue.hpp
sfoster1: the difference is basically "do i want to do this task to do stuff other than wait for messages". for the usb queue, it has literally nothing to do other than wait for messages, so there's no point in it using `try_recv` and waking up every time the timeout pops - it will just loop back to the `try_recv` 100% of the time. but for something like the heater task, 
- it has to receive messages (and send them for that matter)
- it also has to run a control loop once every say 100ms
so there, doing a `try_recv` with a 100ms timeout is a better choice because the ultimate goal is that you fold in every single wake-up condition into a single syscall.

stm32-modules/heater-shaker/include/heater-shaker/errors.hpp
sfoster1: more specific error ideally

stm32-modules/heater-shaker/include/heater-shaker/ack_cache.hpp
sfoster1: yeah it can be pretty useful... although i consider it a pretty serious code smell in basically any other case than "this is for testing do not use it otherwise" since it indicates that somebody is either
a) just kind of giving up on finding appropriate usage restrictions
b) trying to implement extending classes or something bizarre

stm32-modules/heater-shaker/include/heater-shaker/gcodes.hpp
sfoster1: yeah actually this is the one case where going commit-by-commit kind of messes up. if you look at the second-to-last commit, https://github.com/Opentrons/opentrons-modules/pull/134/commits/5d220d61d17934503fd20d06e00d4209d2c9e5b8 (which switches to using iterator pairs) it boils down to a single call to `write_string_to_iterpair` which I think is ok.

the other way to do it would be free function template specializations i'm pretty sure:

```c++
// declare-only for the fully-unstructured template since we don't want to ever actually use it,
// only the specializations
template<typename InputIt, typename Limit, typename GCode, typename... Args>
auto write_response_into(Inputit input, Limit limit, GCode gcode, Args... args) -> InputIt;

template<typename InputIt, typename Limit, typename GCode>
requires std::is_same_v<GCode, M123>  // Limit this specialization to one of a couple gcodes.
             || std::is_same_v<GCode, M104>  // could also be done by giving gcodes a type tag,
                                                                    // a using-declaration called like no_arguments or
                                                                    // something
auto write_response_into(Inputit input, Limit limit, GCode gcode) -> InputIt {
   // do the no-args case; take advantage of some method on the GCode object that writes the
   // gcode-specific data
}

template< InputIt, typename Limit, typename GCode>
requires std::is_same_v<GCode, ...>  // Same deal with the limit but limit it to one of the other
             || ....                                         // ones
auto write_response_into(InputIt input, Limit limit, GCode gcode, uint32_t value) -> InputIt {
   // do the int response writing, taking advantage of some other GCode message
}
```

is this better? honestly not sure.

stm32-modules/heater-shaker/firmware/freertos_comms_task.cpp
sfoster1: they are in the usb driver: https://github.com/STMicroelectronics/STM32CubeF3/blob/v1.11.2/Middlewares/ST/STM32_USB_Device_Library/Core/Inc/usbd_core.h

in general, functions that kind of look like C functions (they have their "namespace" baked into their function call) are going to be from st's driver package

stm32-modules/heater-shaker/include/heater-shaker/host_comms_task.hpp
sfoster1: yeah. imo it's kind of clunky syntax wise, i'd much prefer a match-statement syntax like [rust](https://doc.rust-lang.org/rust-by-example/flow_control/match.html) but c++ is very conservative about that sort of thing

stm32-modules/heater-shaker/include/heater-shaker/motor_task.hpp
sfoster1: yeah exactly

stm32-modules/heater-shaker/include/heater-shaker/gcodes.hpp
sfoster1: it's important for everything in the call chain from `run_once` to have similar templated parameters because that way you can both call it with `char*` in the firmware and with `string.begin()` in the test harness

PR #157, opened 2021-05-04
feat(heater-shaker): main motor support

### LINKED ISSUE #135: heater/shaker: integrate STM32 motor control lib
The STM32 motor control library (https://www.st.com/content/st_com/en/ecosystems/stm32-motor-control-ecosystem.html) is how we will control the brushless DC motor at low RPMs. Like the BSP, it needs to be integrated and configured cleanly without using STM32CubeMX, so it should be
- Added as an external library downloaded by CMake
- Integrated into CMake so it can be built as a static lib and included in the project
- Have integration and configuration live inside the freertos motor control task

At the end of this ticket, the library should compile and not cause the micro to crash. It doesn't have to actually move a motor yet.

sfoster1:
This PR adds support for controlling the primary shake motor for the heater shaker, using the ST MCSDK.

There are a couple different phases to this:
- This PR drops a tremendous number of new files into the project, so I split up the project's firmware directory into subdirectories for each task.
- The actual MCSDK needs to be added. The MCSDK is shipped as a graphical configuration tool that's intended to work with STM CubeMX (https://www.st.com/en/embedded-software/x-cube-mcsdk.html). It has a little application that lets you configure some stuff about the motor parameters and generate a bunch of code, as well as copy over the internal code that doesn't change between configurations. That consists of a pre-compiled library for a variety of core types and some source code wrapping it and implementing a bunch of stuff. The copied-over stuff and the binaries are added in a subdirectory; the generated stuff is added to the motor task.
We're not going to be using the graphical configuration tool again because its configuration is in XML and unfriendly to source control; it requires java 1.8 and nobody wants to install that; and it's a UI driven tool so it would need to be automated with a clicker. Instead, we did an initial configuration pass and any further changes will be done manually in the generated source.
- Those configuration changes need to be made. There are some differences in the NFF from the initial configuration pass that EE made.
- We need actual solid handling for commands and errors and so on at the c++ level.

With all this done, we also need to make sure we compile with optimizations because the actual core process of the code races a timer to do a bunch of calculations and it won't win unless you're on O2 and also drop the PWM frequency some. We should consider putting this in CCMRAM.

This PR should be a sufficient basis for testing the motor. It won't be fully tuned, especially for an unloaded motor, but connecting one of the Faulhabers to the nff (and moving the headers around to configuration B as detailed [here](https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/339051206/Heater+Shaker+Non-Form+Factor+Board) should let the motor spin.

As with the USB PR, the best way to look at this is to go commit-by-commit. You can generally ignore a60c493 because it's just adding files and 50b05b7 because it's just changing line endings.

The manual for the motor control library is attached.

Some other notes about the particular way the mcsdk is integrated:

- It has this UI layer that drives a UART that connects to a host computer application called motor control cockpit for tuning. We don't support that, so most of the UI code and configuration is removed
- I tried really hard not to change any files in the MCSDK subdirectory and failed because one of them had straight up invalid syntax when compiled with a C++ compiler
- Aside from the pinmapping configurations there are a couple specific settings that needed to be changed to work:
   - in `drive_parameters.h`, `PWM_FREQUENCY` needs to be dropped to 10000 (from 30000). This is because the core work of the library in `FOC_CurrControllerM1` in `mc_tasks.c` does this thing where it triggers on a timer interrupt; does a bunch of work; and then at the end faults if there was another timer interrupt since it started, so it's racing a very high frequency timer. This isn't great! This function and the stuff it calls should probably go in CCMRAM at some point, especially if we need to increase the PWM frequency again.
   - in `pmsm_parameters.h`, `HALL_PHASE_SHIFT` needs to be 90 not 300. This is a confusing parameter based on my understanding. It seems like it should be the angular distance between when the first hall sensor triggers and when the back EMF on the first winding peaks, which is straightforward enough, but this motor has seven pole pairs so it should do seven electrical revolutions per full motor mechanical revolution. But there are physically 3 hall sensors statically placed on the motor, so it's unclear to me how that offset could be fixed - it might be 90 at the beginning of the first motor revolution, but it would be different on subsequent ones.

## Testing
It's going to be tough to test this if you don't have a motor. If you do, configure the board and plug in the motor appropriately and build and run the code. Make sure you're using the RelWithDebInfo cmake config; this can be done by rerunning `cmake --preset=stm32-cross -DCMAKE_BUILD_TYPE=RelWithDebInfo .` in `opentrons-modules`. From there, once the code's running, stopping it in the debugger will probably fault the motor controller, so don't do that. You can use `M3 Sxxxx` to set the target rpm to xxxx (including 0 to stop) and `M204 Syyyy` to set the ramp rate to yyyy.

Even then, it's quite likely the motor won't run very well - that's fine, we want to get it basically working so we can tune the controller parameters.

[dm00490980-stm32-motor-control-sdk-stmicroelectronics.pdf](https://github.com/Opentrons/opentrons-modules/files/6422751/dm00490980-stm32-motor-control-sdk-stmicroelectronics.pdf)


djgomez-opentrons (approved):
Tested on NFF and motor, Gcodes are registered and the motor spins. 

Currently, A and C hall effect sensors are swapped on the connector and corrected by pin definition. We should eventually fix this so as not to be confusing.

stm32-modules/heater-shaker/firmware/motor_task/mc_api.c
```diff
@@ -1,463 +1,463 @@
...
+  *
+  * - MC_ProgramSpeedRampMotor1()
+  * - MC_ProgramTorqueRampMotor1()
+  * - MC_SetCurrentReferenceMotor1()
+  *
+  * Failing to do so results in an unpredictable behaviour.
+  *
+  * @note The MC_StartMotor1() command only triggers the start-up procedure:
+  * It moves Motor 1's state machine from the #IDLE to the #IDLE_START state and then
+  * returns. It is not blocking the application until the motor is indeed running.
```
Laura-Danielle: By blocking the application here, we mean specifically only sending commands to the motor? And do we not want the ability to change speeds while the motor is running, or am I misunderstanding something here?

stm32-modules/heater-shaker/firmware/motor_task/mc_api.c
```diff
@@ -1,463 +1,463 @@
...
+  * Feedback error (#MC_SPEED_FDBK) will occur when the rotation speed is about to
+  * reach 0 rpm.
+  *
+  * @param  hFinalSpeed Mechanical rotor speed reference at the end of the ramp.
+  *                     Expressed in the unit defined by #SPEED_UNIT.
+  * @param  hDurationms Duration of the ramp expressed in milliseconds. It
+  *         is possible to set 0 to perform an instantaneous change in the speed
+  *         value.
+  */
+__weak void MC_ProgramSpeedRampMotor1( int16_t hFinalSpeed, uint16_t hDurationms )
```
Laura-Danielle: In reference to my question above, I'm assuming this is the function we would use to change the speed of the motor

stm32-modules/heater-shaker/firmware/CMakeLists.txt
```diff
@@ -52,7 +86,9 @@ set_target_properties(heater-shaker
              C_STANDARD 11
              C_STANDARD_REQUIRED TRUE)
 
-target_include_directories(heater-shaker PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/../include)
+  target_include_directories(heater-shaker
+    PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/../include
+    PRIVATE ${MOTOR_DIR})
```
Laura-Danielle, line 92: What benefit do we get out of explicitly making the motor directory private?

stm32-modules/heater-shaker/include/heater-shaker/gcodes.hpp
```diff
@@ -21,14 +21,15 @@
...
+
 struct SetRPM {
     /*
     ** Set RPM uses the spindle-speed code from standard gcode, M3 (CW)
     ** Format: M3 S<RPM>
     ** Example: M3 S500 sets target rpm to 500
     */
     using ParseResult = std::optional<SetRPM>;
-    uint32_t rpm;
+    int16_t rpm;
```
Laura-Danielle, line 31: Why did we change the type of rpm here?

stm32-modules/heater-shaker/firmware/motor_task/motor_policy.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+#include "motor_policy.hpp"
+
+MotorPolicy::MotorPolicy(MCI_Handle_t *handle): motor_handle(handle) {}
+
+auto MotorPolicy::set_rpm(int16_t rpm) -> void {
+    int16_t current_speed = get_current_rpm();
+    int16_t command_01hz = rpm * _01HZ / _RPM;
+    uint16_t ramp_time =
+        static_cast<int32_t>(std::abs(rpm - current_speed)) * RAMP_MS_PER_RPM;
```
Laura-Danielle: I'm confused about the 3 different types here. RAMPS_MS_PER_RPM is an unsigned 32 int, the result of casting `rpm-current_speed` is a signed 32 int (makes sense) but then we're saying that the total result should be an unsigned 16 int

stm32-modules/heater-shaker/firmware/motor_task/mc_tasks.c
```diff
@@ -191,7 +191,7 @@ void MCboot( MCI_Handle_t* pMCIList[NBR_OF_MOTORS],MCT_Handle_t* pMCTList[NBR_OF
   MCT[M1].pSpeedSensorVirtual = MC_NULL;
   MCT[M1].pSpeednTorqueCtrl = pSTC[M1];
   MCT[M1].pStateMachine = &STM[M1];
-  //MCT[M1].pTemperatureSensor = (NTC_Handle_t *) pTemperatureSensor[M1];
+  MCT[M1].pTemperatureSensor = MC_NULL; //(NTC_Handle_t *) pTemperatureSensor[M1];
```
Laura-Danielle, line 194: Would we not want to eventually monitor the temperature of the motor?

stm32-modules/heater-shaker/include/heater-shaker/errors.hpp
```diff
@@ -1,62 +1,56 @@
...
+    UNDER_VOLT = 2,
+    OVER_TEMP = 3,
+    START_UP = 4,
+    SPEED_FDBK = 5,
+    OVERCURRENT = 6,
+    SW_ERROR = 7,
+};
+
+template <typename IntType>
+auto operator<<(IntType to_shift, MotorErrorOffset offset) -> IntType {
```
Laura-Danielle, line 21: Can you explain the purpose of shifting the motor error offset?

stm32-modules/heater-shaker/firmware/motor_task/motor_policy.cpp
```diff
@@ -17,8 +18,10 @@ auto MotorPolicy::set_rpm(int16_t rpm) -> void {
     }
     int16_t current_speed = get_current_rpm();
     int16_t command_01hz = rpm * _01HZ / _RPM;
-    uint16_t ramp_time =
-        static_cast<int32_t>(std::abs(rpm - current_speed)) * RAMP_MS_PER_RPM;
+    uint16_t ramp_time = std::max(
```
Laura-Danielle, line 28: I see that you slightly changed the types here, but I'm still confused as to why it specifically needs to be unit16_t

Laura-Danielle (approved):
Code looks good, just some small comments again.

stm32-modules/heater-shaker/firmware/motor_task/mc_api.c
sfoster1: to be clear i didn't write this, this is generated code. but what it means is that the two ways stuff like this works is either
1. what they didn't do:
```
void MC_StartMotor1() {
   start(); // do some stuff to kick off the background controller tasks
   while(!started()) {}
   // by the time this function returns, the motor is spinning
}
```
2. what they did do
```
void MC_StartMotor1() {
   start(); // do some stuff to kick off the background controller tasks
  // it's on the higher levels of the application to do whatever scheduling they want and figure out when the motor is actually spinning, this just returns now
}
```

If they'd done 1, and this function didn't return, the thread that called it would be just spinning waiting for the motor, unable to do anything else like handle messages or anything. Since they've done 2, there could be a little more effort required, but you're at least able to do other stuff from the calling task.

stm32-modules/heater-shaker/firmware/motor_task/mc_api.c
sfoster1: Yeah

stm32-modules/heater-shaker/firmware/CMakeLists.txt
sfoster1: Not really that much honestly. It just is The Correct Thing because nothing outside of this application should ever need the stuff in there

stm32-modules/heater-shaker/include/heater-shaker/gcodes.hpp
sfoster1: To match the mc api's underlying data type. Since that's an int16_t, any extra data would just get thrown away.

stm32-modules/heater-shaker/firmware/motor_task/motor_policy.cpp
sfoster1: that probably could be tuned up a bit. it's meant to leave space for overflows before a subsequent divide, but we've baked the divide into the rpm command, so this doesn't really need to be there. I should fix that.

stm32-modules/heater-shaker/firmware/motor_task/mc_tasks.c
sfoster1: Eventually we probably do, but we don't have that thermistor right now

stm32-modules/heater-shaker/include/heater-shaker/errors.hpp
sfoster1: No because I should have deleted this, this was meant to be a nice typesafe bitfield thing but I went with a switch instead

stm32-modules/heater-shaker/firmware/motor_task/motor_policy.cpp
sfoster1: Because that's what the driver code's `ProgramSpeedRamp` function takes. That's pretty much it.

stm32-modules/heater-shaker/include/heater-shaker/errors.hpp
sfoster1: Oh actually it's kind of useful if you look at the tests. It lets you do
```
uint16_t value = 1 << MotorErrorOffset::FOC_DURATION;
```
rather than 
```
uint16_t value = 1 << static_cast<uint8_t>(MotorErrorOffset::FOC_DURATION);
```
everywhere

PR #159, opened 2021-05-05
chore(heater-shaker): add flash target

sfoster1:
We have a target for debugging, but it's actually kind of annoying to
use it to do a flash where you don't actually want to debug.

Add a cmake target that just calls openocd to flash the board and exits.

Also add a couple comments (which get printed in the console) to some
other cmake utilities.

## Testing
- Configure cmake `cmake --preset=stm32-cross .`
- Flash a board `cmake --build ./build-stm32-cross --target heater-shaker-flash`
- it should work


Laura-Danielle (approved):
nice, works for me

PR #161, opened 2021-05-12
feat(heater-shaker): read NTC thermistors

sfoster1:
The heater-shaker has 3 onboard NTC thermistors, all of which are https://www.digikey.com/en/products/detail/tdk-corporation/NTCG104ED104DT1X/8565524 with a 44.2k bias resistor. Two are (or will be) integrated into heater pads but in the NFF are populated on the board and selectable by jumpers; the final one is and always will be on the board.

Reading thermistors consists of setting up an ADC to actually read the voltage out of the circuit; doing some math with the bias resistance to figure out the current resistance of the thermistor; and then using a vendor-provided table to convert that resistance into a temperature (it's an exponential curve with a bunch of little calibration points in it, so it's usually better to use a table than write the equation).

This pull request does the above.
- It adds the table (in csv format) and some python scripting integrated into CMake to generate a c++ source and header pair that provides the table as a std::array. You can see the generated file in `build-stm32-(host|cross)/stm32-modules/heater-shaker/src/thermistor_lookups.(h|c)pp` once you build the tests or firmware; it's a big array. There's some lip service paid in the python generator and the generated code to supporting more than one kind of thermistor, but there's just one for now so it doesn't go that far.
- It adds code in `thermistor_conversion.(h|c)pp` to first convert ADC readings into resistances based on configured electrical values and then lookup the value in the table and interpolate between the closest two entries. It also handles out-of-range (technically out-of-domain, I suppose) ADC readings, and tests all the above
- It adds hardware configuration code to actually read the ADCs in `heater_hardware.(c|h)`. This is in C because the hal code isn't really friendly to modern C++; it plays games with `volatile` casts that are ok in c but not in c++. This is why we disable some warnings whenever we include (something that includes...) the hal code in c++, which lets us have those includes; but if we actually _use_ the macros that use this trick, which are things like the RCC enable macros, in c++ we can't compile it. So the bridge has to be in C. The code is kind of straightforward (populates some structs, calls some hal functions, simple as) but has an annoying quirk: rather than configure the ADC hardware or the HAL to scan all the channels, when we start a conversion we read just one channel and then kick off the next conversion in the subsequent ISR. This is because the ADC will only give us an interrupt after the entire conversion sequence is complete; that means by the time the interrupt fires, the first two conversions will be lost. So we have to get an interrupt for all three, and then reconfigure after reading them.
- It adds integrations throughout the system for reading these values. There's a new heater hardware task (freertos only, not shared) that starts the ADC conversions every 10ms or so. This is because this is a thermal system and therefore very slow. Then, the conversion interrupt calls back into the c++, and sends a message to the real hardware task, which runs the conversion and handles some errors.

There's some further work to be done just on the error handling around some edge cases (detailed lower down) but this is testable and works and I don't want another giant PR.

## Testing
- Flash this on an NFF
- Make sure the NTC select jumpers are set to the "PCB" side 
- Send some `M105`s and make sure the values are pretty OK
- Pull off each NTC jumper individually and make sure the board starts spamming you with messages about out of range. Try both (A connected, B disconnected) and (A disconnected, B disconnected) and make sure you get the spam in both cases; this is what the final commit addresses.
- Optionally, put your hand on the NTCs or something, or point a hair dryer or heat gun or heat source of some kind at the board and see if the temperature goes up. I'm pretty confident here, and also the real tests need some actual equipment, so only do this if it's convenient.

## Further work
- It probably shouldn't endlessly spam you if a thermistor is disconnected! There should be some latch-until-reset-by-host status that cuts off heat control if one of the thermistors stops working, and it should just say that error once, and also complain when it's in that state and you ask it to set a temperature or get a temperature
- We should have errors that are both more useful (in reality, you're not going to see a condition where the thermistor is working fine but it's below -40, because cmon; and you're not going to see a condition where the thermistor is working but it's over 125, because the thermal shutoff will happen - these are really sensing thermistor short circuits or disconnections) and more specific (thermistor A disconnected or thermistor B disconnected)


stm32-modules/heater-shaker/firmware/heater_task/freertos_heater_task.cpp
```diff
@@ -35,21 +42,75 @@ static FreeRTOSMessageQueue<heater_task::Message>
...
+        : hardware{.conversions_complete = conversion_handler,
+                   .hardware_internal = nullptr},
+          heater_main_task(queue),
+          hardware_task_handle(nullptr) {}
+};
+
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
+HeaterTasks _heater_tasks(_heater_queue, handle_conversion);
+
+static void handle_conversion(conversion_results *results) {
```
amitlissack: Should the input be `const conversion_results * results`? Does it matter?

stm32-modules/heater-shaker/firmware/heater_task/freertos_heater_task.cpp
```diff
@@ -35,21 +42,75 @@ static FreeRTOSMessageQueue<heater_task::Message>
...
 // Actual function that runs the task
 void run(void *param) {
-    auto *task_class = static_cast<decltype(_task) *>(param);
+    auto *local_tasks = static_cast<HeaterTasks *>(param);
+    while (true) {
+        local_tasks->heater_main_task.run_once();
+    }
+}
+
+void run_hardware_task(void *param) {
```
amitlissack, line 100: Can we have a comment describing of what this task is for?

stm32-modules/heater-shaker/src/generate_thermistor_table.py
```diff
@@ -0,0 +1,118 @@
...
+        next(self._csv) # part number header
+        next(self._csv) # R25 coefficient
+        next(self._csv) # B25 coefficient
+        next(self._csv) # empty line
+        headers = next(self._csv)
+        next(self._csv) # units
+        self._lines = [l for l in self._csv]
+        self._extractor = {'nominal': self.Nom, 'min': self.Min, 'max': self.Max}[which_resistance]
+        self._start_space = ' ' * at_space_depth
+        self._incremental_space = ' '*incremental_space_depth
```
amitlissack: nit: spaces around the `*`

amitlissack (commented):
Some questions and requests for comments.

Don't feel equipped to approve this.

stm32-modules/heater-shaker/firmware/heater_task/freertos_heater_task.cpp
sfoster1: Yeah it definitely should be, good catch. It doesn't  really matter that much but it's good practice.

djgomez-opentrons (approved):
Tested this on an NFF with NTCs in a temperature controlled water bath. 

electrical measurement claimed NTCs should have been reading ~25C and ~30C, the USB output was sort of switching around between these two. The NFF only had one NTC with the needed buffer. 

Basically, the core reading functionality appears to be working - there is still work needed in tuning the ADC settings and perhaps getting better debug outputs. Some better reworks on the NFF are also required before we can really test that stuff though so this is good for now.

Laura-Danielle (approved):
lgtm

PR #162, opened 2021-05-13
feat(heater-shaker): better thermistor errors

sfoster1:
This pr introduces
- Latching for thermistor errors: instead of constantly printing errors
for every ADC conversion that reveals a bad thermistor, just print it
once when it happens and then again in response to any heater command
- More specific thermistor errors for specific error conditions and
sensors: a thermistor a disconnect error, thermistor b short error, etc
etc.

## Testing
- Put it on a board
- Disconnect and reconnect thermistors. Note you only get one error every time you disconnect one.
- While at least one thermistor is disconnected, sending an M105 or an M104 should result in a disconnect error

## Review
- Does the centralization of sensor state I've done make sense?
- How about the centralization of heater state in general? This State struct will also be used to handle errors coming from things like the analog overtemperature latches.


stm32-modules/heater-shaker/include/heater-shaker/heater_task.hpp
```diff
@@ -82,72 +137,129 @@ requires MessageQueue<QueueImpl<Message>, Message> class HeaterTask {
...
     Queue& message_queue;
     tasks::Tasks<QueueImpl>* task_registry;
-    const thermistor_conversion::Conversion pad_a_conversion;
-    const thermistor_conversion::Conversion pad_b_conversion;
-    const thermistor_conversion::Conversion board_conversion;
-    int16_t pad_a_temperature;
-    int16_t pad_b_temperature;
-    int16_t board_temperature;
+    TemperatureSensor pad_a;
+    TemperatureSensor pad_b;
```
amitlissack, line 260: Are these members public?

sfoster1: no, the `private` section isn't in the diff but it's on line 134 of this file

stm32-modules/heater-shaker/include/heater-shaker/heater_task.hpp
```diff
@@ -29,21 +62,43 @@ requires MessageQueue<QueueImpl<Message>, Message> class HeaterTask {
     static constexpr const uint32_t CONTROL_PERIOD_TICKS = 100;
     static constexpr double THERMISTOR_CIRCUIT_BIAS_RESISTANCE_KOHM = 44.2;
     static constexpr uint8_t ADC_BIT_DEPTH = 12;
+    static constexpr double HEATER_PAD_OVERTEMP_SAFETY_LIMIT_C = 100;
```
Laura-Danielle, line 65: Why aren't these typed as int32_t?

stm32-modules/heater-shaker/include/heater-shaker/heater_task.hpp
```diff
@@ -82,72 +137,129 @@ requires MessageQueue<QueueImpl<Message>, Message> class HeaterTask {
...
+            auto response =
+                messages::AcknowledgePrevious{.responding_to_id = msg.id};
+            static_cast<void>(
+                task_registry->comms->get_message_queue().try_send(
+                    messages::HostCommsMessage(response)));
+        }
     }
 
     auto visit_message(const messages::GetTemperatureMessage& msg) -> void {
+        errors::ErrorCode code = pad_a.error != errors::ErrorCode::NO_ERROR
```
Laura-Danielle, line 159: I am having trouble understanding the logic on this line.

Laura-Danielle (approved):
Had two questions, but otherwise looks good.

stm32-modules/heater-shaker/include/heater-shaker/heater_task.hpp
sfoster1: Because of https://github.com/Opentrons/opentrons-modules/pull/163 mostly

stm32-modules/heater-shaker/include/heater-shaker/heater_task.hpp
sfoster1: Honestly I probably shouldn't do this with a ternary. The idea is to get the error held by either of pad a or pad b, and to end up with no error only if neither has an error code set.

PR #163, opened 2021-05-14
feat(heater-shaker): print subdegree temps

sfoster1:
We convert and hold all our temperatures in doubles, but weren't
printing them that way, because, well, look at this commit.

We now print them, which requires using snprintf, because nobody's
implemented std::to_chars for float or double.

## Testing
- See if you get nice numbers with decimal points when you send an M105
- No actual conversion code changed anywhere, just the printing code, so that's pretty much it

### DIFF
--- a/stm32-modules/heater-shaker/include/heater-shaker/gcodes.hpp
+++ b/stm32-modules/heater-shaker/include/heater-shaker/gcodes.hpp
@@ -122,7 +122,7 @@ struct GetTemperature {
         write_response_into(InputIt buf, InLimit limit,
                             double current_temperature,
                             double setpoint_temperature) -> InputIt {
-        auto res = snprintf(&*buf, (limit-buf), "M105 C%0.2f T%0.2f OK\n",
+        auto res = snprintf(&*buf, (limit - buf), "M105 C%0.2f T%0.2f OK\n",
                             static_cast<float>(current_temperature),
                             static_cast<float>(setpoint_temperature));
         if (res <= 0) {
--- a/stm32-modules/heater-shaker/include/heater-shaker/gcodes.hpp
+++ b/stm32-modules/heater-shaker/include/heater-shaker/gcodes.hpp
@@ -122,35 +122,13 @@ struct GetTemperature {
         write_response_into(InputIt buf, InLimit limit,
                             double current_temperature,
                             double setpoint_temperature) -> InputIt {
-        static constexpr const char* prefix = "M105 C";
-        char* char_next = &*buf;
-        char* char_limit = &*limit;
-        char_next = write_string_to_iterpair(char_next, char_limit, prefix);
-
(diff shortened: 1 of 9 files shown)


stm32-modules/heater-shaker/include/heater-shaker/gcodes.hpp
```diff
@@ -119,29 +120,33 @@ struct GetTemperature {
     requires std::forward_iterator<InputIt>&&
         std::sized_sentinel_for<InputIt, InLimit> static auto
         write_response_into(InputIt buf, InLimit limit,
-                            uint32_t current_temperature,
-                            uint32_t setpoint_temperature) -> InputIt {
+                            double current_temperature,
+                            double setpoint_temperature) -> InputIt {
         static constexpr const char* prefix = "M105 C";
```
amitlissack: Pardon the dumb questions....

If you're using snprintf, why not go full on with the format string `"M105 C%02.F T%02.F"`?

What the heck is all this code in the body of a `struct`?

sfoster1: that's... a good point about the full format string. i probably should do this, yeah.

as to the struct thing - c++ structs are (more or less) classes with default `public` visibility, so these are (static) member functions.

amitlissack: Got it. Thanks

amitlissack (approved):

stm32-modules/heater-shaker/include/heater-shaker/gcodes.hpp
```diff
@@ -69,7 +70,7 @@ struct SetTemperature {
     ** Example: M104 S25 sets target temperature to 25C
     */
     using ParseResult = std::optional<SetTemperature>;
-    float temperature;
+    double temperature;
```
Laura-Danielle, line 73: Does this have to be a double?

stm32-modules/heater-shaker/include/heater-shaker/messages.hpp
```diff
@@ -87,8 +87,8 @@ struct ErrorMessage {
 */
 struct GetTemperatureResponse {
     uint32_t responding_to_id;
-    int16_t current_temperature;
-    uint32_t setpoint_temperature;
+    double current_temperature;
```
Laura-Danielle, line 90: same here on the double, is it really necessary? I know this type is generally more expensive than others.

Laura-Danielle (approved):
Just a comment about doubles, otherwise looks good.

stm32-modules/heater-shaker/include/heater-shaker/messages.hpp
sfoster1: Doubles aren't really truly necessary, but it's nice to just not have to worry about it. And the timescales for the temperature stuff is so much slower than anything else we care about that it'll be fine.

PR #165, opened 2021-05-19
feat(heater-shaker): Handle the heater power error latch

sfoster1:
The heater/shaker has a hardware error detection mechanism which ANDs together in digital logic a whole bunch of error sources (fixed resistor thresholds feeding comparators for thermistor disconnect/short/overtemp) and feeds the result into an SR latch, with the micro having a line on the ~SR line. What all that means is

- We need to read the state of that latch, and feed it into our error detection mechanisms
- When we see that all the errors are gone, or when we want to heat, we should try and reset it, by holding the ~SR asserted (low) for a bit
- If it doesn't reset, we should stay in an error state
This PR basically does that. It adds an error for the hardware error latch, and reads it when it does ADC conversions.

When we go from pad sense error -> no pad sense error, we try and reset the latch. We also try and reset the latch when someone sends a set-temperature command.

Also we need to enable sscanf support for float separately, because of course we do.

## Testing
It's actually pretty hard to test this, because some of the comparators are configured incorrectly on the first-run NFFs and the latch is always locked on. But you can test that you get the appropriate error when you try and set a temperature, or when you disconnect and then reconnect a thermistor. That's about it.

### DIFF
--- a/stm32-modules/heater-shaker/firmware/heater_task/freertos_heater_task.cpp
+++ b/stm32-modules/heater-shaker/firmware/heater_task/freertos_heater_task.cpp
@@ -10,8 +10,8 @@
 #include "firmware/freertos_message_queue.hpp"
 #include "heater-shaker/heater_task.hpp"
 #include "heater-shaker/tasks.hpp"
+#include "heater_policy.hpp"
 #include "task.h"
-
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wvolatile"
 #include "heater_hardware.h"
@@ -56,12 +56,15 @@ struct HeaterTasks {
     heater_task::HeaterTask<FreeRTOSMessageQueue> heater_main_task;
     // NOLINTNEXTLINE(misc-non-private-member-variables-in-classes)
     TaskHandle_t hardware_task_handle;
+    // NOLINTNEXTLINE(misc-non-private-member-variables-in-classes)
+    HeaterPolicy policy;
     HeaterTasks(FreeRTOSMessageQueue<heater_task::Message> &queue,
                 void (*conversion_handler)(const conversion_results *))
         : hardware{.conversions_complete = conversion_handler,
                    .hardware_internal = nullptr},
           heater_main_task(queue),
-          hardware_task_handle(nullptr) {}
+          hardware_task_handle(nullptr),
+          policy(&hardware) {}
 };
 
 // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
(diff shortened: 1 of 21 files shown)


stm32-modules/heater-shaker/firmware/heater_task/freertos_heater_task.cpp
```diff
@@ -56,12 +56,15 @@ struct HeaterTasks {
     heater_task::HeaterTask<FreeRTOSMessageQueue> heater_main_task;
     // NOLINTNEXTLINE(misc-non-private-member-variables-in-classes)
     TaskHandle_t hardware_task_handle;
+    // NOLINTNEXTLINE(misc-non-private-member-variables-in-classes)
```
amitlissack, line 59: Is this directive for a linter? Is it really complaining about non-public members of a struct?  You don't have a choice, do you!

stm32-modules/heater-shaker/include/heater-shaker/heater_task.hpp
```diff
@@ -183,46 +224,83 @@ requires MessageQueue<QueueImpl<Message>, Message> class HeaterTask {
...
+                    state.system_status = State::ERROR;
+                }
+            } else {
+                state.system_status = State::ERROR;
+            }
+        } else if ((changes & State::POWER_GOOD_ERROR) != 0) {
+            auto error_message =
+                messages::HostCommsMessage(messages::ErrorMessage{
+                    .code = errors::ErrorCode::HEATER_HARDWARE_ERROR_LATCH});
+            static_cast<void>(
```
amitlissack, line 279: Why is `static_cast<void>` needed?

amitlissack (approved):
I don't get all of the details but code looks good.

stm32-modules/heater-shaker/firmware/heater_task/freertos_heater_task.cpp
sfoster1: Oh you can _make_ them private, they just default to public. I really wish I could turn  this off (it's clang-tidy specifically) but  I do in fact want this warning for _classes_.  I just like my bundle-of-data structs  :(

stm32-modules/heater-shaker/include/heater-shaker/heater_task.hpp
sfoster1: `try_send` returns a `bool` that is `true` if the message was sent, and `false`  otherwise (e.g. because the queue was full, it  has a limited size  since it's statically allocated). That method is annotated `[[nodiscard]]`, which means the compiler will reject calls to  this method that implicitly discard its return value;  you have to explicitly discard it, whichI do here because if this message doesn't get sent it fundamentally doesn't matter since  the state is still set.

PR #166, opened 2021-05-19
fix(heater-shaker): lower min ramp, neg rpm

sfoster1:
Need to spin the other way; need to lower min rpm ramp rate to 1


amitlissack (approved):

djgomez-opentrons (approved):
Tested on an NFF in the office, it works motor spins the correct way and the ramp rate can be reduced to 1. The units for ramp rate seem to be wrong but that's a separate issue.

PR #168, opened 2021-05-24
feat(heater-shaker): Heater support

sfoster1:
Links up the heater and heat control system.

- Brings up the heat pad  driver through a PWM channel
- Links up the control system and the new direct-power-set command to the hardware for heater driving

- New Gcode M104.D Sxx.xx allows direct set of relative heater power on a 0-1 scale; useful for testing. When running in this mode, the setpoint  temperature response from M105 is replaced with the power setpoint.

The control system itself isn't tuned at all yet, because we don't have the heater pads we're looking for, but the drive is verified on the bench using M104.D; the controller is tested and verified; and the thermistors and error  sense  are there.

There's one missing error sense condition, which is using the heater pad switch's current sense output to check for unplugged/short conditions; but we also don't  have the hardware for that on NFF rev A.

## Review Requests
- The whole policy-passed-in-to-run_once thing is still looking good right
- How badly do I need to split up the gcode tests into separate files?  Pretty bad, right?
- 
## Testing
Testing for this is mostly hardware focused. We should use a scope and M104.D to check that we're accurately commanding  heater power from 0-1, and honestly just trust that the controller is vaguely right until we get actual heater pads we can control on.

### DIFF
--- a/stm32-modules/heater-shaker/firmware/heater_task/heater_hardware.c
+++ b/stm32-modules/heater-shaker/firmware/heater_task/heater_hardware.c
@@ -16,6 +16,7 @@ static ntc_selection next_channel(ntc_selection from_which);
 typedef struct {
     ntc_selection reading_which;
     conversion_results results;
+    ADC_HandleTypeDef ntc_adc;
 } hw_internal;
 
 hw_internal _internals;
@@ -33,7 +34,8 @@ heater_hardware *HEATER_HW_HANDLE = NULL;
 #define HEATER_PGOOD_LATCH_PORT GPIOD
 #define HEATER_PGOOD_LATCH_PIN (1<<13)
 
-void gpio_setup(void) {
+
+static void gpio_setup(void) {
     // NTC sense pis all routed to the ADC
     GPIO_InitTypeDef gpio_init = {
     .Pin = (NTC_PAD_B_PIN | NTC_BOARD_PIN),
@@ -61,7 +63,7 @@ void gpio_setup(void) {
                       GPIO_PIN_SET);
 }
 
-void adc_setup(ADC_HandleTypeDef* adc) {
+static void adc_setup(ADC_HandleTypeDef* adc) {
     adc->Instance = ADC3;
     adc->Init.ClockPrescaler = ADC_CLOCK_SYNC_PCLK_DIV1;
     adc->Init.Resolution = ADC_RESOLUTION_12B;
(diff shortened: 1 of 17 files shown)


sfoster1:
Scope screenshots. Channel 1 (yellow) is the heater pad switch input,  channel 2 (green) is the heater power (the shots show the vertical scale for channel  2 at 5 v/div, so the top of the channel  2 is 36V; channel 1 tops  out around 3.3 at 0.5V/div)

5% duty cycle (M104.D S0.95)
fivepercent

50%  dutycycle (M104.D S0.5)
fifty

95% duty cycle (M104.D S0.95)
ninetyfive


There's no integration with the pad driver's current detection so it will run "fine" if you run it with nothing plugged in (but probably don't).

stm32-modules/heater-shaker/firmware/heater_task/heater_hardware.c
```diff
@@ -132,15 +191,36 @@ void heater_hardware_release_pg_latch() {
                       GPIO_PIN_SET);
 }
 
+void heater_hardware_power_disable(heater_hardware* hardware) {
+    hw_internal* internal = (hw_internal*)hardware->hardware_internal;
```
Laura-Danielle, line 195: Why do we need to cast `hardware->hardware_internal` here?

stm32-modules/heater-shaker/tests/test_gcodes.cpp
```diff
@@ -1,7 +1,10 @@
 #include <array>
 
 #include "catch2/catch.hpp"
+#pragma GCC diagnostic push
```
Laura-Danielle, line 4: It would be nice to split this file up, yes. But I don't think it needs to be in this PR.

stm32-modules/heater-shaker/tests/test_gcodes.cpp
```diff
@@ -687,3 +690,293 @@ SCENARIO("GetTemperatureDebug parser works") {
...
+                buffer.begin(), buffer.begin() + 5);
+            THEN("the response should be written only up to the size") {
+                REQUIRE_THAT(buffer, Catch::Matchers::Equals("M301 ccccc"));
+                REQUIRE(written == buffer.begin() + 5);
+            }
+        }
+    }
+}
+
+SCENARIO("SetHeaterPowerTest parser works") {
```
Laura-Danielle, line 881: Organizationally (I don't think that's a word..), it would be helpful for me to include the GCode in the scenario title.

Laura-Danielle (approved):
Looks good to me. I think you should eventually split up the test file for gcodes, and the policy being passed in still looks good to me.

stm32-modules/heater-shaker/tests/test_gcodes.cpp
sfoster1: Oh, that's a good idea. I'll do that in the pr where i split this file  up.

stm32-modules/heater-shaker/firmware/heater_task/heater_hardware.c
sfoster1: this is kind of the "private data" idiom in c, where you have a struct definition in your header (heater_hardware) with stuff you expect outside code  to initialize for you - like the conversion_complete callback - and a `void*`,  which erases types, to something you define only in your c file  and thus that only you know about, which is `hw_internal`. The downside is you have to make a local copy with a cast like this every time because the compiler needs to know what's on the other side of the void*.

PR #169, opened 2021-05-26
feat(heater-shaker): Add bootloader entry

### LINKED ISSUE #151: heater/shaker: bootloader entry sequencing
The STM32F3 has an integrated DFU bootloader implementation in its system memory, but it's hard to get into it since it requires an external pin. in theory, we can remap the interrupt table and jump to the system memory entry point; let's make sure that actually works.

sfoster1:
We would like to be able to update heater-shaker units from the field. There is a built-in bootloader on these parts ([AN2606](https://www.st.com/resource/en/application_note/cd00167594-stm32-microcontroller-system-memory-boot-mode-stmicroelectronics.pdf) is the relevant datasheet on the bootloaders). It exposes something not entirely unlike standard DFU. It also has some bizarre entry methods on the F303 part - you must change the state of an external pin, which would require extra circuitry.

However, the machinery that maps different parts of memory to 0x0 is user accessible in syscfg, and we know how cortex-m boot works, so we can jump to that built-in bootloader  from software after making sure we reset our clocking to something that it expects.

This has to be done in assembler because a c call messes around with the stack, which we don't want, so it's a little ugly, but it does work.

We also add the `dfu` command, which will enter this mode after stopping any ongoing heater or motor commands and nicely disconnecting from USB.

## Testing
- Connect a board via USB
- send it "dfu"
- see that  it goes away and comes back reporting itself as an stm32 bootloader
- use http://dfu-util.sourceforge.net/ or something similar for non-osx platforms to mess around with it. You can tell it to boot back to application mode using the `-s 0x0800000:leave` argument  to dfu-util. You can program  it from here but you have to put it in an annoying format first which I'll get to later

Sorry about the assembly.


stm32-modules/heater-shaker/include/heater-shaker/system_task.hpp
```diff
@@ -0,0 +1,156 @@
...
+        auto stop_message = messages::SetRPMMessage{
+            .id = 0, .target_rpm = 0, .from_system = true};
+        auto stop_id = prep_cache.add(stop_message);
+        stop_message.id = stop_id;
+        if (!task_registry->motor->get_message_queue().try_send(stop_message,
+                                                                1)) {
+            prep_cache.remove_if_present(stop_id);
+        }
+
+        auto cool_message = messages::SetTemperatureMessage{
```
Laura-Danielle, line 86: There aren't actually active cooling fans on the heater shaker right? Should we call this `deactivate_message` or something else?

stm32-modules/heater-shaker/include/heater-shaker/system_task.hpp
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        auto ack_message =
+            messages::AcknowledgePrevious{.responding_to_id = message.id};
+        static_cast<void>(
+            task_registry->comms->get_message_queue().try_send(ack_message, 1));
+
+        // Somehow we couldn't send any of the messages, maybe system deadlock?
+        // Enter bootloader regardless
+        if (prep_cache.empty()) {
```
Laura-Danielle, line 110: Would entering bootloader during a system deadlock potentially brick the hardware?

Laura-Danielle (approved):
Tested on the nff board and it seemed to work fine. I had two small comments, otherwise code looks good to me.

stm32-modules/heater-shaker/include/heater-shaker/system_task.hpp
sfoster1: Nope completely separate program

stm32-modules/heater-shaker/include/heater-shaker/system_task.hpp
sfoster1: I guess technically but it doesn't really matter IMO

PR #170, opened 2021-05-27
chore(heater-shaker): split and refactor gcode tests

sfoster1:
test_gcodes.cpp was pushing 1k lines which isn't really useful to
anybody. Instead,
- Take advantage of catch's template-parameterized tests to do the
common identical tests across all valid gcodes. Some of this can be
refined further but for now it's just the no-parameter cases (both for
parse and response generation) and the always-fail cases like
non-matching prefixes
- Split what remained up into different files
- Add labels by the actual gcode so it's easier to read.


stm32-modules/heater-shaker/include/heater-shaker/gcode_parser.hpp
```diff
@@ -43,7 +43,7 @@ requires std::forward_iterator<Input>&& std::sized_sentinel_for<Limit, Input>&&
                         typename PrefixArray::value_type> auto
     prefix_matches(const Input& start_from, Limit stop_at,
                    const PrefixArray& prefix) -> Input {
-    if (static_cast<size_t>(stop_at - start_from) <= prefix.size()) {
+    if (static_cast<size_t>(stop_at - start_from) < prefix.size()) {
```
Laura-Danielle, line 46: Did you mean to make this change?

Laura-Danielle (approved):
Awesome thanks! Just one comment.

amitlissack (approved):

stm32-modules/heater-shaker/include/heater-shaker/gcode_parser.hpp
sfoster1: Yeah the old way would incorrectly reject some very specific (and only seen in testing) cases, where you send exactly the prefix and nothing else, including a newline. That is an incorrect input - the newline is required - but it's an incorrect input that should happen at a higher layer.

PR #173, opened 2021-05-27
feat(heater-shaker): quick versioning

### LINKED ISSUE #171: heater/shaker: basic software versioning
As we go forward into wider testing we need to be able to keep track of what firmware is on what device. Store the git commit hash and best symbolic ref (e.g. branchname) as well as the version-from-tag information and make it accessible.

### LINKED ISSUE #172: heater/shaker: basic hardware capability/versioning
Just in case we want multiple kinds of heater/shaker, establish a precedent for a hardware versioning and capabilities system that will tell the upstream things like the shake radius.

sfoster1:
While we don't want full tag based versioning yet, it's nice to have
something. The heater-shaker now captures the current git
status (unfortunately, you'll need to do a make clean to update it -
this is mostly useful from CI) and put it in the build triggered by a
new gcode called version.

### DIFF
--- a/stm32-modules/heater-shaker/include/heater-shaker/gcodes.hpp
+++ b/stm32-modules/heater-shaker/include/heater-shaker/gcodes.hpp
@@ -453,12 +453,11 @@ struct EnterBootloader {
 
 struct GetVersion {
     /**
-     * GetVersion keys off the string "version" and returns hardware and
-     * software versions
+     * GetVersion keys off gcode M115 and returns hardware and
+     * software versions and eventually serial numbers
      * */
     using ParseResult = std::optional<GetVersion>;
-    static constexpr auto prefix =
-        std::array{'v', 'e', 'r', 's', 'i', 'o', 'n'};
+    static constexpr auto prefix = std::array{'M', '1', '1', '5'};
 
     template <typename InputIt, typename InLimit>
     requires std::forward_iterator<InputIt>&&
@@ -466,7 +465,7 @@ struct GetVersion {
         write_response_into(InputIt write_to_buf, InLimit write_to_limit,
                             const char* fw_version, const char* hw_version)
             -> InputIt {
-        static constexpr const char* prefix = "version FW:";
+        static constexpr const char* prefix = "M115 FW:";
         auto written =
             write_string_to_iterpair(write_to_buf, write_to_limit, prefix);
         if (written == write_to_limit) {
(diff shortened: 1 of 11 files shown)


stm32-modules/heater-shaker/include/heater-shaker/gcodes.hpp
```diff
@@ -423,4 +423,53 @@ struct SetHeaterPowerTest {
     }
 };
 
+struct GetVersion {
```
Laura-Danielle, line 454: Should this be in its own file? It feels kind of weird to have this in the gcodes header file.

stm32-modules/heater-shaker/src/CMakeLists.txt
```diff
@@ -3,6 +3,30 @@
...
+string(SUBSTRING ${PREPENDED_VERSION} 1 -l VERSION)
+if (NOT VERSION)
+  set(VERSION "(dev)")
+endif()
+execute_process(COMMAND ${GIT_EXECUTABLE} rev-parse --short HEAD
+  WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
+  OUTPUT_VARIABLE HASH
+  OUTPUT_STRIP_TRAILING_WHITESPACE)
+
+set(heater-shaker_VERSION "${VERSION}-${HASH}")
```
Laura-Danielle, line 27: Is there a reason to make this a random mix of capital/lower case + kebab and snake case?

Laura-Danielle (commented):
Two small comments about naming and location of `getVersion`

stm32-modules/heater-shaker/include/heater-shaker/gcodes.hpp
sfoster1: yeah.... but it needs to be parsed by the same thing... maybe I should get more creative about finding a gcode that matches it

stm32-modules/heater-shaker/src/CMakeLists.txt
sfoster1: to stay consistent with the project version utility, basically. and that's like that because appending `_VARIABLE` to some prefix is sort of the way (sadly) you do "return values" from cmake functions or macros

stm32-modules/heater-shaker/include/heater-shaker/gcodes.hpp
Laura-Danielle: I suppose it's not necessary for this PR, but it would be nice for the future.

Laura-Danielle (approved):

Laura-Danielle (approved):
lgtm. Thanks!

PR #175, opened 2021-05-28
feat(heater-shaker): Add simulator

sfoster1:
Adds a host-compiled simulator for the heater-shaker!

This needs boost (and adds the cmake infra for it, see Opentrons/cmake-utils#4) so definitely do a clean cmake configure: rm -rf ./build-stm32-host && cmake --preset=stm32-host[-gcc] . (i.e. --preset=stm32-host on linux, --preset=stm32-host-gcc on osx with the CMakeUserPresets.json active). This will take a while because it's downloading boost.

Then you can build the simulator: cmake --build ./build-stm32-host --target heater-shaker-simulator
Then you can run the simulator: ./build-stm32-host/stm32-modules/heater-shaker/simulator/heater-shaker-simulator. It will read from stdin, so send some gcodes, and print to stdout.

The simulator build replaces the rtos tasks with the new std::jthread and the freertos queues with the ever useful boost lockfree queue.

This needs a little more love; it should probably learn to use things other than stdin/stdout through the (mildly terrifying) boost::asio, maybe learn some command line flags about taking non-zero time to do things, maybe get a little more tested than me sending a couple M105s and grinning like a maniac, maybe try a compile on linux, maybe try a compile on windows (almost certainly will not work correctly).

We also just abort() (or, well, std::terminate(), this is c++ and we're not monsters) on "bootloader entry" which probably should change.


amitlissack (approved):
Love it.

PR #178, opened 2021-06-03
feat(heater-shaker): homing with solenoid

### LINKED ISSUE #147: heater/shaker: integrate home latch hardware
The heater/shaker has a latch that holds the plate in position when it is homed. Integrate the hardware that drives this and add the capability to the motor control task.

### LINKED ISSUE #148: heater/shaker: add home behavior
Once the homing driver is integrated, the heater/shaker needs to be able to actually get there. This consists of slowing to an appropriate speed, and then coasting into the home position while the homing hardware engages. This state machine needs to be integrated into the motor task.

sfoster1:
This PR adds the capability to home the plate, moving it slowly and engaging a solenoid to lock it in place.

That includes
- Bringing up the solenoid driver, which is an allegro h-bridge with current limit set by a DAC
- Adding a debug gcode, `G28.D Sxxx` where xxx is a current in integer milliamps (0 means disengage), to test the solenoid
- Adding a home gcode, `G28`, that starts the homing process
- Adding a home-process state machine to the motor task

The home-process state machine is a little involved because it wants to not block the motor task while it's going. This is done by having the motor task send a message back to itself every so often to check the homing status when it's in that state.

## Testing
- You'll need a prototype (I also need this, haven't tested the home sequence yet) to test the home sequence
- You'll need a solenoid to test the solenoid part standalone. This I've done - we do accurately set current and engage the solenoid with G28.D.
- Test that it both homes and also restarts ok 

Closes #148 
Closes #147


amitlissack (approved):
Can't test it, but looks good to me.

sfoster1:
With the latter two commits, it mostly works - it needs some more tweaking for consistent results, but the tweakable values are in the right place (and the tweaks may look more like motor system tuning than anything to do specifically with homing).

stm32-modules/heater-shaker/firmware/motor_task/motor_policy.hpp
```diff
@@ -3,22 +3,35 @@
 #include <cstdint>
 
 #include "heater-shaker/errors.hpp"
+
+#pragma GCC diagnostic push
```
Laura-Danielle, line 7: Just curious as to why the header file isn't in the include folder here.

stm32-modules/heater-shaker/simulator/motor_thread.cpp
```diff
@@ -59,6 +67,11 @@ auto run(std::stop_token st, std::shared_ptr<TaskControlBlock> tcb) -> void {
         } catch (const SimMotorTask::Queue::StopDuringMsgWait& sdmw) {
             return;
         }
+        if (tcb->task.get_state() ==
+            motor_task::State::HOMING_COASTING_TO_STOP) {
```
Laura-Danielle, line 71: I'm a bit confused, is coasting to stop bad? Why are we trying to send an error, or am I misunderstanding the code?

Laura-Danielle (approved):
Code looks good to me, just two questions

stm32-modules/heater-shaker/simulator/motor_thread.cpp
sfoster1: Well, the error message is kind of internal to the motor task. This is emulating the motor  driver saying "i had an overcurrent event", aka "i stalled because the solenoid is engaged". the message is sent  to the motor task, and therefore exercises the code on line 179 of motor_task.hpp (sadly tough to interlink) above. This is the way we tell the motor task that homing is done, on th simulation side.

PR #183, opened 2021-06-07
feat(heater-shaker): plate lock motor

sfoster1:
Add M240.D S(-)0.xxx to control the plate latch motor. No endstops or anything, just turn at a duty cycle.

This is controlled by the motor task and usse TIM3.

### DIFF
--- a/stm32-modules/heater-shaker/include/heater-shaker/host_comms_task.hpp
+++ b/stm32-modules/heater-shaker/include/heater-shaker/host_comms_task.hpp
@@ -41,12 +41,12 @@ requires MessageQueue<QueueImpl<Message>, Message> class HostCommsTask {
         gcode::GetTemperature, gcode::SetAcceleration,
         gcode::GetTemperatureDebug, gcode::SetHeaterPIDConstants,
         gcode::SetHeaterPowerTest, gcode::EnterBootloader, gcode::GetVersion,
-        gcode::Home, gcode::ActuateSolenoid>;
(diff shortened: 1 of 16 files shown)


stm32-modules/heater-shaker/include/heater-shaker/host_comms_task.hpp
```diff
@@ -579,6 +579,31 @@ requires MessageQueue<QueueImpl<Message>, Message> class HostCommsTask {
...
+            -> std::pair<bool, InputIt> {
+        auto id = ack_only_cache.add(gcode);
+        if (id == 0) {
+            return std::make_pair(
+                false, errors::write_into(tx_into, tx_limit,
+                                          errors::ErrorCode::GCODE_CACHE_FULL));
+        }
+        auto message =
+            messages::SetPlateLockPowerMessage{.id = id, .power = gcode.power};
+        if (!task_registry->motor->get_message_queue().try_send(
```
amitlissack, line 596: nit: there's quite a lot of duplicated  logic in the various `visit_gcode` functions.

amitlissack (approved):

stm32-modules/heater-shaker/include/heater-shaker/host_comms_task.hpp
sfoster1: There really is, I don't like it. I think I can fix it (probably in a future PR) but it's going to be pretty templatey.

PR #189, opened 2021-06-08
chore(heater-shaker): motor  tuning v1

sfoster1:
First-pass at motor tuning support, wrapping up a whole bunch of changes and new additions.

- adds the ability to change the motor speed control loop constants using M301, which now has a new target argument: `M301 T[H|P] Pxxx Ixxx Dxxx`.
- adds a scripts/ directory with a `requirements.txt` for making a venv, a readme, and a script for running a motor speed command and capturing the speed readbacks
- swaps some  parameters now
- fixes a mistake I guess I made at some point I can't remember that broke motor control library error detection; we now save errors before resetting them.
- moves the motor control library high-frequency control task, which runs the motor control system, to the [CCMRAM](https://www.st.com/content/ccc/resource/technical/document/application_note/bb/09/ca/83/14/e9/44/c5/DM00083249.pdf/files/DM00083249.pdf/jcr:content/translations/en.DM00083249.pdf). There's more detail at the link, but the short of it is it's a separate RAM area with 1-cycle fetch latencies which therefore makes code in it run faster. In practice, moving the motor control system from execute-from-flash to execute-from-CCMRAM makes the primary control loop task take 11us, whereas on flash it takes 14us. That's nice because the main failure mode of this control system is not finishing the control loop by the time its output is needed, and when it's running at 30KHz from CCMRAM, the standard timings give you about 5us of margin between a normal end-of-task and when the value is needed. So if it's in flash, that margin is gone. End of this post has some scope captures about this. The actual mechanics of the transition are a little fiddly. A different static library version of the motor controller builtins has to be linked in, and a compiler define has to be set, and the startup assembly needs to be changed a little. This is necessary because CCMRAM is ram, not flash, and therefore isn't persistent. During startup, the application has to put the code into CCMRAM every time. That means that it needs to know where the code lives in flash before the copy; this is done by telling the linker to put that code in a special section (which is why all those functions have that preprocessor-logic-gated ``__attribute__((section=".ccmram"))`` - we set the preproc define to make sure those apply), and then the startup assembly has to be changed to use the locations the linker knows about to copy all the data. The startup assembly is taken from [this community post](https://community.arm.com/developer/ip-products/system/f/embedded-forum/4605/how-to-use-ccm-sram-for-cortex-m4); the linker-script stuff was already there. The data is copied in the same way as the way we copy the initial values of initialized data into RAM at the appropriate address. Unlike initialized data, it would probably be fine to do this copy in C as long as we did it before trying to call any of the functions in question, but this keeps the nastiness localized, and prevents any upsetting surprises if we put more stuff in ccmram in the future.
- tweaks the interrupt priorities for pretty much everything that has an interrupt around to keep that 5us margin intact. Because we have an [NVIC](https://developer.arm.com/documentation/ddi0439/b/Nested-Vectored-Interrupt-Controller/NVIC-programmers-model), interrupts at higher priority (Lower priority number (sorry)) can preempt ISRs of lower priority (higher priority number). Interrupts of equal priority are serviced round-robin. In addition to the hardware-handling ISRs we define, for things like the heater ADC, the various timers and ADCs for the motor, and the USB, there's also the ones that FreeRTOS uses internally. FreeRTOSConfig.h [has a value](https://github.com/Opentrons/opentrons-modules/blob/652bbbc848feceba61801114bbc078f6cf341e66/stm32-modules/heater-shaker/firmware/system/FreeRTOSConfig.h#L144) that is a priority-not-to-exceed for internal library methods, for basically exactly this reason. If you make an ISR at a higher priority (lower priority number) than that config value, then it will preempt whatever the RTOS is doing internally. This means you can't use any FreeRTOS methods from ISRs at that priority, but that's fine, we don't need to - the [_other_ motor control task](https://github.com/Opentrons/opentrons-modules/blob/652bbbc848feceba61801114bbc078f6cf341e66/stm32-modules/heater-shaker/firmware/motor_task/freertos_motor_task.cpp#L78-L89), which runs as a normal task rather than in an ISR, handles all that stuff and is safe to do so.
- puts that high-frequency task regulation frequency back at 30KHz 

Running the controller at 20KHz, with the code in flash. Channel 1 goes high at the beginning of the control loop and low at the end. 
highfrek-20k-noccmram

Running at 20KHz, with the code in CCMRAM. You can see it's a lot faster.
highfreq-20k-ccmram

Running at 30KHz. This is a logic analyzer capture rather than a scope. Channel 0 is the same logic with the high-frequency task. Channel 2 is the timer interrupt that the controller is racing; if that goes high while channel 0 is still high (as it does around timing marker 1), there's an error. When an error is present, channel 1 is high. You can ignore Channel 3; I thought the USB interrupt was the main problem, but it's not really related. This capture shows the reason to tweak the interrupts - the variability in the duration and frequency of the peaks on channel 1.
Screen Shot 2021-06-10 at 9 29 18 AM

The logic analyzer capture from the above screenshot. This can be opened with [Saleae Logic 2](https://www.saleae.com/downloads/)
[30khz-failure.sal.zip](https://github.com/Opentrons/opentrons-modules/files/6631755/30khz-failure.sal.zip)

### DIFF
--- a/stm32-modules/heater-shaker/firmware/host_comms_task/usbd_conf.c
+++ b/stm32-modules/heater-shaker/firmware/host_comms_task/usbd_conf.c
@@ -64,7 +64,7 @@ void HAL_PCD_MspInit(PCD_HandleTypeDef *hpcd)
   __HAL_RCC_SYSCFG_CLK_ENABLE();
 
   /* Set USB Default FS Interrupt priority */
-  HAL_NVIC_SetPriority(USB_LP_CAN_RX0_IRQn, 5, 0);
+  HAL_NVIC_SetPriority(USB_LP_CAN_RX0_IRQn, 10, 0);
 
   /* Enable USB FS Interrupt */
   HAL_NVIC_EnableIRQ(USB_LP_CAN_RX0_IRQn);
--- a/stm32-modules/heater-shaker/firmware/motor_task/motor_policy.cpp
+++ b/stm32-modules/heater-shaker/firmware/motor_task/motor_policy.cpp
@@ -52,11 +52,11 @@ auto MotorPolicy::set_rpm(int16_t rpm) -> ErrorCode {
     }
     int16_t current_speed = get_current_rpm();
     int16_t command_01hz = rpm * _01HZ / _RPM;
-    uint16_t ramp_time = std::max(
+    uint16_t ramp_time_ms = std::max(
         static_cast<uint16_t>(
-            static_cast<int32_t>(std::abs(rpm - current_speed)) / ramp_rate),
+            static_cast<int32_t>(std::abs(rpm - current_speed)) / ramp_rate_rpm_per_ms),
         static_cast<uint16_t>(1));
-    MCI_ExecSpeedRamp(hw_handles->mci[0], -command_01hz, ramp_time);
+    MCI_ExecSpeedRamp(hw_handles->mci[0], -command_01hz, ramp_time_ms);
     if (MCI_GetSTMState(hw_handles->mci[0]) == IDLE) {
         MCI_StartMotor(hw_handles->mci[0]);
     }
@@ -84,7 +84,7 @@ auto MotorPolicy::set_ramp_rate(int32_t rpm_per_s) -> ErrorCode {
         rpm_per_s < MIN_RAMP_RATE_RPM_PER_S) {
         return ErrorCode::MOTOR_ILLEGAL_RAMP_RATE;
     }
-    ramp_rate = rpm_per_s;
+    ramp_rate_rpm_per_ms = std::max(static_cast<double>(rpm_per_s) / 1000.0, 1.0);
     return ErrorCode::NO_ERROR;
 }
 
--- a/stm32-modules/heater-shaker/firmware/motor_task/motor_policy.cpp
+++ b/stm32-modules/heater-shaker/firmware/motor_task/motor_policy.cpp
@@ -52,10 +52,11 @@ auto MotorPolicy::set_rpm(int16_t rpm) -> ErrorCode {
     }
     int16_t current_speed = get_current_rpm();
     int16_t command_01hz = rpm * _01HZ / _RPM;
-    uint16_t ramp_time_ms = std::max(
-        static_cast<uint16_t>(
-            static_cast<int32_t>(std::abs(rpm - current_speed)) / ramp_rate_rpm_per_ms),
-        static_cast<uint16_t>(1));
+    uint16_t ramp_time_ms =
+        std::max(static_cast<uint16_t>(
+                     static_cast<int32_t>(std::abs(rpm - current_speed)) /
+                     ramp_rate_rpm_per_ms),
+                 static_cast<uint16_t>(1));
     MCI_ExecSpeedRamp(hw_handles->mci[0], -command_01hz, ramp_time_ms);
     if (MCI_GetSTMState(hw_handles->mci[0]) == IDLE) {
         MCI_StartMotor(hw_handles->mci[0]);
@@ -84,7 +85,9 @@ auto MotorPolicy::set_ramp_rate(int32_t rpm_per_s) -> ErrorCode {
         rpm_per_s < MIN_RAMP_RATE_RPM_PER_S) {
         return ErrorCode::MOTOR_ILLEGAL_RAMP_RATE;
     }
-    ramp_rate_rpm_per_ms = std::max(static_cast<double>(rpm_per_s) / 1000.0, 1.0);
+    ramp_rate_rpm_per_ms =
+        // NOLINTNEXTLINE(cppcoreguidelines-avoid-magic-numbers)
+        std::max(static_cast<double>(rpm_per_s) / 1000.0, 1.0);
     return ErrorCode::NO_ERROR;
 }
 
@@ -98,4 +101,11 @@ auto MotorPolicy::plate_lock_set_power(float power) -> void {
 
 auto MotorPolicy::plate_lock_disable() -> void {
     motor_hardware_plate_lock_off(&hw_handles->tim3);
+}
+
+// NOLINTNEXTLINE(readability-convert-member-functions-to-static)
+auto MotorPolicy::set_pid_constants(double kp, double ki, double kd) -> void {
+    static_cast<void>(kp);
+    static_cast<void>(ki);
+    static_cast<void>(kd);
 }
--- a/stm32-modules/heater-shaker/firmware/motor_task/motor_policy.cpp
+++ b/stm32-modules/heater-shaker/firmware/motor_task/motor_policy.cpp
@@ -52,11 +52,10 @@ auto MotorPolicy::set_rpm(int16_t rpm) -> ErrorCode {
     }
     const int16_t current_speed = get_current_rpm();
     const int16_t command_01hz = rpm * _01HZ / _RPM;
-    const uint32_t speed_diff = std::abs(rpm-current_speed);
+    const uint32_t speed_diff = std::abs(rpm - current_speed);
     const uint32_t uncapped_ramp_time_ms = speed_diff / ramp_rate_rpm_per_ms;
-    const uint16_t ramp_time_ms =
-        std::max(static_cast<uint16_t>(uncapped_ramp_time_ms),
-                 static_cast<uint16_t>(1));
+    const uint16_t ramp_time_ms = std::max(
+        static_cast<uint16_t>(uncapped_ramp_time_ms), static_cast<uint16_t>(1));
     MCI_ExecSpeedRamp(hw_handles->mci[0], -command_01hz, ramp_time_ms);
     if (MCI_GetSTMState(hw_handles->mci[0]) == IDLE) {
         MCI_StartMotor(hw_handles->mci[0]);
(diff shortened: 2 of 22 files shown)


stm32-modules/heater-shaker/firmware/motor_task/motor_policy.cpp
```diff
@@ -99,3 +102,11 @@ auto MotorPolicy::plate_lock_set_power(float power) -> void {
 auto MotorPolicy::plate_lock_disable() -> void {
     motor_hardware_plate_lock_off(&hw_handles->tim3);
 }
+
+auto MotorPolicy::set_pid_constants(double kp, double ki, double kd) -> void {
+    // These conversions match those in drive_parameters.h and therefore let you just look
+    // at the numeric literals there
+    PID_SetKD(hw_handles->mct[0]->pPIDSpeed, static_cast<int16_t>(kd / (SPEED_UNIT / 10)));
```
amitlissack: Why is "/ 10" needed? What is 10? Could it be folded into SPEED_UNIT or another named constant?

stm32-modules/heater-shaker/firmware/motor_task/motor_policy.cpp
```diff
@@ -52,11 +52,12 @@ auto MotorPolicy::set_rpm(int16_t rpm) -> ErrorCode {
     }
     int16_t current_speed = get_current_rpm();
     int16_t command_01hz = rpm * _01HZ / _RPM;
-    uint16_t ramp_time = std::max(
-        static_cast<uint16_t>(
-            static_cast<int32_t>(std::abs(rpm - current_speed)) / ramp_rate),
-        static_cast<uint16_t>(1));
-    MCI_ExecSpeedRamp(hw_handles->mci[0], -command_01hz, ramp_time);
+    uint16_t ramp_time_ms =
```
amitlissack: Nit: That's a lot of stuff in one statement. Could we split it up to multiple statements?

stm32-modules/heater-shaker/firmware/motor_task/motor_policy.cpp
sfoster1: It matches the logic here: https://github.com/Opentrons/opentrons-modules/blob/652bbbc848feceba61801114bbc078f6cf341e66/stm32-modules/heater-shaker/firmware/motor_task/drive_parameters.h#L95-L100

That logic meanwhile is there because of autogenerated code that almost but does not quite adapt itself for changing the SPEED_UNIT define. It's supposed to be a thing you can change between RPM, Hz, and 0.1Hz and it changes all the math; but the autogenerated code that works with it in fact slaps those factors-of-10 in there by default which only works for 0.1Hz and I really don't want to deal with changing it.

Because this is a test gcode handler, the benefit of having the factor-of-10 magic is that the pid constants the caller sets now match numerically the values that are set by default in the above link.

stm32-modules/heater-shaker/firmware/motor_task/motor_policy.hpp
```diff
@@ -35,6 +36,8 @@ class MotorPolicy {
 
   private:
     static constexpr uint16_t MAX_SOLENOID_CURRENT_MA = 330;
-    int32_t ramp_rate = DEFAULT_RAMP_RATE_RPM_PER_S;
+    double ramp_rate_rpm_per_ms =
```
Laura-Danielle, line 39: why did we convert this to ms?

Laura-Danielle (approved):

stm32-modules/heater-shaker/firmware/motor_task/motor_policy.cpp
```diff
@@ -99,3 +101,16 @@ auto MotorPolicy::plate_lock_set_power(float power) -> void {
 auto MotorPolicy::plate_lock_disable() -> void {
     motor_hardware_plate_lock_off(&hw_handles->tim3);
 }
+
+auto MotorPolicy::set_pid_constants(double kp, double ki, double kd) -> void {
+    // These conversions match those in drive_parameters.h and therefore let you
+    // just look at the numeric literals there
+    static constexpr const double SPEED_UNIT_CONVERSION_SPC =
```
ahiuchingau, line 108: Why are both `constexpr` and `const` used here?

stm32-modules/heater-shaker/firmware/motor_task/motor_policy.cpp
amitlissack: Got it. Thanks for the explanation.

amitlissack (approved):

stm32-modules/heater-shaker/firmware/motor_task/motor_policy.cpp
sfoster1: they imply different things annoyingly. Honestly it's probably fine to leave either or both of them off since compilers are pretty good about recognizing it but might as well slot em in there

stm32-modules/heater-shaker/firmware/motor_task/motor_policy.hpp
sfoster1: Because it actually always has been whoops. The function in the MCLib we're calling, [MCI_ExecSpeedRamp](https://github.com/Opentrons/opentrons-modules/blob/652bbbc848feceba61801114bbc078f6cf341e66/stm32-modules/heater-shaker/firmware/motor_task/mc_interface.h#L85-L86), has its duration set in ms, and I just... never noticed until now.

stm32-modules/heater-shaker/firmware/motor_task/motor_policy.hpp
```diff
@@ -20,6 +20,7 @@ class MotorPolicy {
     MotorPolicy() = delete;
     explicit MotorPolicy(motor_hardware_handles* handles);
     auto set_rpm(int16_t rpm) -> errors::ErrorCode;
+    auto set_pid_constants(double kp, double ki, double kd) -> void;
```
amitlissack, line 23: Style question: Why use auto instead of using void return type?

stm32-modules/heater-shaker/firmware/system/startup_stm32f303xe.s
```diff
@@ -78,8 +78,25 @@ LoopCopyDataInit:
 	adds	r2, r0, r1
 	cmp	r2, r3
 	bcc	CopyDataInit
+	movs r1, #0
```
amitlissack, line 81: IS this generated?

stm32-modules/heater-shaker/scripts/requirements.txt
```diff
@@ -0,0 +1,9 @@
+cycler==0.10.0
```
amitlissack, line 1: Have you used pandas? I suggest we use it if we get more into plotting and data analysis. It's obscenely powerful.

amitlissack (approved):
Code looks good to me. Just some little questions.

stm32-modules/heater-shaker/firmware/motor_task/motor_policy.hpp
sfoster1: this form of the  function prototype: `auto name(args...) -> return_type;` is equivalent to `return_type name(args...)` but allows the return type to be derived from the args, or implied or something sometimes... most of the time it's pointless, but sometimes it's really useful, so the linter requires it, so I do  it to appease the linter.

stm32-modules/heater-shaker/firmware/system/startup_stm32f303xe.s
sfoster1: The file in general? Part of it (the interrupt vector table part) probably is. These parts are probably handwritten. Specifically my changes here are copied from the data init section and altered a little bit to work the ccm copy.

stm32-modules/heater-shaker/scripts/requirements.txt
sfoster1: I have not, I should check it out at some point. This stuff isn't really controlled at all and doesn't ship so I just kind of went with what I knew.

stm32-modules/heater-shaker/firmware/host_comms_task/usbd_conf.c
```diff
@@ -64,7 +64,7 @@ void HAL_PCD_MspInit(PCD_HandleTypeDef *hpcd)
   __HAL_RCC_SYSCFG_CLK_ENABLE();
 
   /* Set USB Default FS Interrupt priority */
-  HAL_NVIC_SetPriority(USB_LP_CAN_RX0_IRQn, 5, 0);
+  HAL_NVIC_SetPriority(USB_LP_CAN_RX0_IRQn, 10, 0);
```
Laura-Danielle, line 67: I thought you said in the description that the USB was not actually causing issues. Do we need to change the priority of this interrupt then?

Laura-Danielle (approved):
Code looks good, in general, just had a question about a USB interrupt priority.

stm32-modules/heater-shaker/firmware/host_comms_task/usbd_conf.c
sfoster1: well, we _probably_ don't. but in general it's nice to keep things deprioritized to avoid running out of priority space.

ahiuchingau (approved):
Still not proficient in c++ yet but the parts I do understand look good!

PR #194, opened 2021-07-21
chore: update cmake-utils

sfoster1:
Pull recent changes in cmake-utils into the subtree here and apply the new formatter changes.


stm32-modules/heater-shaker/include/heater-shaker/motor_task.hpp
```diff
@@ -38,15 +38,13 @@ namespace motor_task {
...
-    ->std::same_as<int16_t>;
-    { cp.get_target_rpm() }
-    ->std::same_as<int16_t>;
+    { p.set_rpm(static_cast<int16_t>(16)) } -> std::same_as<errors::ErrorCode>;
+    { cp.get_current_rpm() } -> std::same_as<int16_t>;
+    { cp.get_target_rpm() } -> std::same_as<int16_t>;
     {p.stop()};
-    { p.set_ramp_rate(static_cast<int32_t>(8)) }
-    ->std::same_as<errors::ErrorCode>;
+    {
```
pmoegenburg, line 45: make 2 lines?

pmoegenburg: lines 45-47

pmoegenburg (approved):
looks a lot nicer

stm32-modules/heater-shaker/include/heater-shaker/motor_task.hpp
sfoster1: that's the autoformatter doing that. if we've got some autoformatter config changes we should do them, but we shouldn't change back individual choices it makes.

PR #203, opened 2021-09-27
feat(heater_shaker): Add socket communication in simulator

DerekMaggio:
# Overview

- Add socket communication to heater shaker simulation
- Add ability to specify whether to use stdin or socket for G-Code input through cli options

# Changelog

- Added sim_driver.hpp/cpp
  - Contains `SimDriver` abstract base class
  - Contains `StdinSimDriver` and `SocketSimDriver` which inherit from `SimDriver`
  - Both `StdinSimDriver` and `SocketSimDriver` manage reading and writing to stdin and a socket respectively
  - `SocketSimDriver` handles connecting to socket
- Added cli_parser.hpp/cpp
  - Manages parsing passed cli options and returning the correct `SimDriver`

# Review requests

* This is my first venture into writing C++, hold my code to the same standards as everyone else's and tear it apart
* Note the following TODOs and let me know if it is okay to address these in a subsequent PR
    * [Refactor run function to use a driver.write method](https://github.com/Opentrons/opentrons-modules/pull/203/files#diff-170cba0464601e13288a39ae9871eb0887df83e71211194d443938fccfb31ea2R24)
    * [Change socket code to bind to a socket instead of connecting to it](https://github.com/Opentrons/opentrons-modules/pull/203/files#diff-07b6889b97919fe2f098a8eb9f0d19edc3394c672e25e7c60e5659c5785c174aR41)
    * [Socket read method starts dropping G-Codes if they are sent faster than one every 0.005 seconds](https://github.com/Opentrons/opentrons-modules/pull/203/files#diff-07b6889b97919fe2f098a8eb9f0d19edc3394c672e25e7c60e5659c5785c174aR58-R61)

# Risk assessment

None, stdin functionality was preserved and there is no automation that is dependent on it

### DIFF
--- a/stm32-modules/heater-shaker/include/simulator/cli_parser.hpp
+++ b/stm32-modules/heater-shaker/include/simulator/cli_parser.hpp
@@ -1,0 +1,8 @@
+#include <string>
+
+
+
+namespace cli_parser {
+    enum SimType {STDIN, SOCKET};
+    SimType get_sim_type(int, char**);
+}
--- a/stm32-modules/heater-shaker/include/simulator/cli_parser.hpp
+++ b/stm32-modules/heater-shaker/include/simulator/cli_parser.hpp
@@ -1,8 +1,8 @@
-#include <string>
 #include <memory>
+#include <string>
+
 #include "simulator/sim_driver.hpp"
 
-
 namespace cli_parser {
-    sim_driver::SimDriver* get_sim_driver(int, char**);
+sim_driver::SimDriver* get_sim_driver(int, char**);
 }
--- a/stm32-modules/heater-shaker/include/simulator/cli_parser.hpp
+++ b/stm32-modules/heater-shaker/include/simulator/cli_parser.hpp
@@ -1,8 +1,8 @@
 #include <string>
-
+#include <memory>
+#include "simulator/sim_driver.hpp"
 
 
 namespace cli_parser {
-    enum SimType {STDIN, SOCKET};
-    SimType get_sim_type(int, char**);
+    sim_driver::SimDriver* get_sim_driver(int, char**);
 }
--- a/stm32-modules/heater-shaker/include/simulator/cli_parser.hpp
+++ b/stm32-modules/heater-shaker/include/simulator/cli_parser.hpp
@@ -4,5 +4,5 @@
 #include "simulator/sim_driver.hpp"
 
 namespace cli_parser {
-std::unique_ptr<sim_driver::SimDriver> get_sim_driver(int, char**);
+std::shared_ptr<sim_driver::SimDriver> get_sim_driver(int, char**);
 }
--- a/stm32-modules/heater-shaker/include/simulator/cli_parser.hpp
+++ b/stm32-modules/heater-shaker/include/simulator/cli_parser.hpp
@@ -4,5 +4,5 @@
 #include "simulator/sim_driver.hpp"
 
 namespace cli_parser {
-sim_driver::SimDriver* get_sim_driver(int, char**);
+std::unique_ptr<sim_driver::SimDriver> get_sim_driver(int, char**);
 }
--- a/stm32-modules/heater-shaker/include/simulator/comm_thread.hpp
+++ b/stm32-modules/heater-shaker/include/simulator/comm_thread.hpp
@@ -10,7 +10,8 @@
 namespace comm_thread {
 using SimCommTask = host_comms_task::HostCommsTask<SimulatorMessageQueue>;
 struct TaskControlBlock;
-auto build(std::shared_ptr<sim_driver::SimDriver>&&) -> tasks::Task<std::unique_ptr<std::jthread>, SimCommTask>;
+auto build(std::shared_ptr<sim_driver::SimDriver>&&)
+    -> tasks::Task<std::unique_ptr<std::jthread>, SimCommTask>;
 void handle_input(std::shared_ptr<sim_driver::SimDriver>&& driver,
                   tasks::Tasks<SimulatorMessageQueue>& tasks);
 };  // namespace comm_thread
--- a/stm32-modules/heater-shaker/include/simulator/comm_thread.hpp
+++ b/stm32-modules/heater-shaker/include/simulator/comm_thread.hpp
@@ -10,7 +10,7 @@
 namespace comm_thread {
 using SimCommTask = host_comms_task::HostCommsTask<SimulatorMessageQueue>;
 struct TaskControlBlock;
-auto build() -> tasks::Task<std::unique_ptr<std::jthread>, SimCommTask>;
-void handle_input(std::unique_ptr<sim_driver::SimDriver>&& driver,
+auto build(std::shared_ptr<sim_driver::SimDriver>&&) -> tasks::Task<std::unique_ptr<std::jthread>, SimCommTask>;
+void handle_input(std::shared_ptr<sim_driver::SimDriver>&& driver,
                   tasks::Tasks<SimulatorMessageQueue>& tasks);
 };  // namespace comm_thread
--- a/stm32-modules/heater-shaker/include/simulator/comm_thread.hpp
+++ b/stm32-modules/heater-shaker/include/simulator/comm_thread.hpp
@@ -4,13 +4,13 @@
 
 #include "heater-shaker/host_comms_task.hpp"
 #include "heater-shaker/tasks.hpp"
-#include "simulator/simulator_queue.hpp"
 #include "simulator/sim_driver.hpp"
-
+#include "simulator/simulator_queue.hpp"
 
 namespace comm_thread {
 using SimCommTask = host_comms_task::HostCommsTask<SimulatorMessageQueue>;
 struct TaskControlBlock;
 auto build() -> tasks::Task<std::unique_ptr<std::jthread>, SimCommTask>;
-void handle_input(sim_driver::SimDriver* driver, tasks::Tasks<SimulatorMessageQueue>& tasks);
+void handle_input(sim_driver::SimDriver* driver,
+                  tasks::Tasks<SimulatorMessageQueue>& tasks);
 };  // namespace comm_thread
--- a/stm32-modules/heater-shaker/include/simulator/comm_thread.hpp
+++ b/stm32-modules/heater-shaker/include/simulator/comm_thread.hpp
@@ -5,10 +5,12 @@
 #include "heater-shaker/host_comms_task.hpp"
 #include "heater-shaker/tasks.hpp"
 #include "simulator/simulator_queue.hpp"
+#include "simulator/sim_driver.hpp"
+
 
 namespace comm_thread {
 using SimCommTask = host_comms_task::HostCommsTask<SimulatorMessageQueue>;
 struct TaskControlBlock;
 auto build() -> tasks::Task<std::unique_ptr<std::jthread>, SimCommTask>;
-auto handle_input(tasks::Tasks<SimulatorMessageQueue>& tasks) -> void;
+void handle_input(sim_driver::SimDriver* driver, tasks::Tasks<SimulatorMessageQueue>& tasks);
 };  // namespace comm_thread
--- a/stm32-modules/heater-shaker/include/simulator/comm_thread.hpp
+++ b/stm32-modules/heater-shaker/include/simulator/comm_thread.hpp
@@ -11,6 +11,6 @@ namespace comm_thread {
 using SimCommTask = host_comms_task::HostCommsTask<SimulatorMessageQueue>;
 struct TaskControlBlock;
 auto build() -> tasks::Task<std::unique_ptr<std::jthread>, SimCommTask>;
-void handle_input(sim_driver::SimDriver* driver,
+void handle_input(std::unique_ptr<sim_driver::SimDriver>&& driver,
                   tasks::Tasks<SimulatorMessageQueue>& tasks);
 };  // namespace comm_thread
--- a/stm32-modules/heater-shaker/include/simulator/sim_driver.hpp
+++ b/stm32-modules/heater-shaker/include/simulator/sim_driver.hpp
@@ -1,0 +1,41 @@
+#pragma once
+#include "heater-shaker/host_comms_task.hpp"
+#include "heater-shaker/messages.hpp"
+#include "heater-shaker/tasks.hpp"
+
+
+namespace sim_driver {
+    const std::string SOCKET_DRIVER_NAME = "Socket";
+    const std::string STDIN_DRIVER_NAME = "Stdin";
+    class SimDriver {
+        std::string name;
+
+        public:
+            virtual std::string get_name() = 0;
+            virtual void write() = 0;
+            virtual std::string read() = 0;
+    };
+
+    class SocketSimDriver: public SimDriver {
+        std::string name = SOCKET_DRIVER_NAME;
+        std::string host;
+        int port;
+
+        public:
+            SocketSimDriver(std::string);
+            std::string get_host();
+            int get_port();
+            std::string get_name();
+            void write();
+            std::string read();
+    };
+
+    class StdinSimDriver: public SimDriver {
+        std::string name = STDIN_DRIVER_NAME;
+        public:
+            StdinSimDriver();
+            std::string get_name();
+            void write();
+            std::string read();
+    };
+}
--- a/stm32-modules/heater-shaker/include/simulator/sim_driver.hpp
+++ b/stm32-modules/heater-shaker/include/simulator/sim_driver.hpp
@@ -12,4 +12,4 @@ class SimDriver {
     virtual void write(std::string message) = 0;
     virtual void read(tasks::Tasks<SimulatorMessageQueue>& tasks) = 0;
 };
-}
+}  // namespace sim_driver
--- a/stm32-modules/heater-shaker/include/simulator/sim_driver.hpp
+++ b/stm32-modules/heater-shaker/include/simulator/sim_driver.hpp
@@ -12,28 +12,4 @@ class SimDriver {
     virtual void write() = 0;
     virtual void read(tasks::Tasks<SimulatorMessageQueue>& tasks) = 0;
 };
-
-class SocketSimDriver : public SimDriver {
-    static const std::string name;
-    std::string host;
-    int port;
-
-  public:
-    SocketSimDriver(std::string);
-    std::string get_host();
-    int get_port();
-    const std::string& get_name() const;
-    void write();
-    void read(tasks::Tasks<SimulatorMessageQueue>& tasks);
-};
-
-class StdinSimDriver : public SimDriver {
-    static const std::string name;
-
-  public:
-    StdinSimDriver();
-    const std::string& get_name() const;
-    void write();
-    void read(tasks::Tasks<SimulatorMessageQueue>& tasks);
-};
-}  // namespace sim_driver
+}
--- a/stm32-modules/heater-shaker/include/simulator/sim_driver.hpp
+++ b/stm32-modules/heater-shaker/include/simulator/sim_driver.hpp
@@ -7,7 +7,6 @@
 namespace sim_driver {
 
 class SimDriver {
-
   public:
     virtual const std::string& get_name() const = 0;
     virtual void write() = 0;
--- a/stm32-modules/heater-shaker/include/simulator/sim_driver.hpp
+++ b/stm32-modules/heater-shaker/include/simulator/sim_driver.hpp
@@ -4,39 +4,39 @@
 #include "heater-shaker/tasks.hpp"
 #include "simulator/simulator_queue.hpp"
 
-
 namespace sim_driver {
-    const std::string SOCKET_DRIVER_NAME = "Socket";
-    const std::string STDIN_DRIVER_NAME = "Stdin";
-    class SimDriver {
-        std::string name;
+const std::string SOCKET_DRIVER_NAME = "Socket";
+const std::string STDIN_DRIVER_NAME = "Stdin";
+class SimDriver {
+    std::string name;
 
-        public:
-            virtual std::string get_name() = 0;
-            virtual void write() = 0;
-            virtual void read(tasks::Tasks<SimulatorMessageQueue>& tasks) = 0;
-    };
+  public:
+    virtual std::string get_name() = 0;
+    virtual void write() = 0;
+    virtual void read(tasks::Tasks<SimulatorMessageQueue>& tasks) = 0;
+};
 
-    class SocketSimDriver: public SimDriver {
-        std::string name = SOCKET_DRIVER_NAME;
-        std::string host;
-        int port;
+class SocketSimDriver : public SimDriver {
+    std::string name = SOCKET_DRIVER_NAME;
+    std::string host;
+    int port;
 
-        public:
-            SocketSimDriver(std::string);
-            std::string get_host();
-            int get_port();
-            std::string get_name();
-            void write();
-            void read(tasks::Tasks<SimulatorMessageQueue>& tasks);
-    };
+  public:
+    SocketSimDriver(std::string);
+    std::string get_host();
+    int get_port();
+    std::string get_name();
+    void write();
+    void read(tasks::Tasks<SimulatorMessageQueue>& tasks);
+};
 
-    class StdinSimDriver: public SimDriver {
-        std::string name = STDIN_DRIVER_NAME;
-        public:
-            StdinSimDriver();
-            std::string get_name();
-            void write();
-            void read(tasks::Tasks<SimulatorMessageQueue>& tasks);
-    };
-}
+class StdinSimDriver : public SimDriver {
+    std::string name = STDIN_DRIVER_NAME;
+
+  public:
+    StdinSimDriver();
+    std::string get_name();
+    void write();
+    void read(tasks::Tasks<SimulatorMessageQueue>& tasks);
+};
+}  // namespace sim_driver
--- a/stm32-modules/heater-shaker/include/simulator/sim_driver.hpp
+++ b/stm32-modules/heater-shaker/include/simulator/sim_driver.hpp
@@ -5,18 +5,17 @@
 #include "simulator/simulator_queue.hpp"
 
 namespace sim_driver {
-const std::string SOCKET_DRIVER_NAME = "Socket";
-const std::string STDIN_DRIVER_NAME = "Stdin";
+
 class SimDriver {
 
   public:
-    virtual const std::string& get_name() const = 0
+    virtual const std::string& get_name() const = 0;
     virtual void write() = 0;
     virtual void read(tasks::Tasks<SimulatorMessageQueue>& tasks) = 0;
 };
 
 class SocketSimDriver : public SimDriver {
-    std::string name = SOCKET_DRIVER_NAME;
+    static const std::string name;
     std::string host;
     int port;
 
@@ -24,17 +23,17 @@ class SocketSimDriver : public SimDriver {
     SocketSimDriver(std::string);
     std::string get_host();
     int get_port();
-    std::string get_name();
+    const std::string& get_name() const;
     void write();
     void read(tasks::Tasks<SimulatorMessageQueue>& tasks);
 };
 
 class StdinSimDriver : public SimDriver {
-    std::string name = STDIN_DRIVER_NAME;
+    static const std::string name;
 
   public:
     StdinSimDriver();
-    std::string get_name();
+    const std::string& get_name() const;
     void write();
     void read(tasks::Tasks<SimulatorMessageQueue>& tasks);
 };
--- a/stm32-modules/heater-shaker/include/simulator/sim_driver.hpp
+++ b/stm32-modules/heater-shaker/include/simulator/sim_driver.hpp
@@ -8,10 +8,9 @@ namespace sim_driver {
 const std::string SOCKET_DRIVER_NAME = "Socket";
 const std::string STDIN_DRIVER_NAME = "Stdin";
 class SimDriver {
-    std::string name;
 
   public:
-    virtual std::string get_name() = 0;
+    virtual const std::string& get_name() const = 0
     virtual void write() = 0;
     virtual void read(tasks::Tasks<SimulatorMessageQueue>& tasks) = 0;
 };
--- a/stm32-modules/heater-shaker/include/simulator/sim_driver.hpp
+++ b/stm32-modules/heater-shaker/include/simulator/sim_driver.hpp
@@ -2,6 +2,7 @@
 #include "heater-shaker/host_comms_task.hpp"
 #include "heater-shaker/messages.hpp"
 #include "heater-shaker/tasks.hpp"
+#include "simulator/simulator_queue.hpp"
 
 
 namespace sim_driver {
@@ -13,7 +14,7 @@ namespace sim_driver {
         public:
             virtual std::string get_name() = 0;
             virtual void write() = 0;
-            virtual std::string read() = 0;
+            virtual void read(tasks::Tasks<SimulatorMessageQueue>& tasks) = 0;
     };
 
     class SocketSimDriver: public SimDriver {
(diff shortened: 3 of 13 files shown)


DerekMaggio:
This is **SUPER** work in progress
There are a couple things I need help with
1. I can receive socket messages. But when I am trying to pass them inside of `handle_input` I get no response. I have an inkling that it is something with the second argument [here](https://github.com/Opentrons/opentrons-modules/pull/203/commits/5cb9ad0c30380a537201b8dc6a1d77af64c49487#diff-170cba0464601e13288a39ae9871eb0887df83e71211194d443938fccfb31ea2R70)
2. I need some advice on how to pull out the socket stuff. Should it stay in the comms file or go in it's own? 
3. How do I go about getting switching between reading from stdin as opposed to a socket?

stm32-modules/heater-shaker/simulator/comm_thread.cpp
```diff
@@ -39,16 +40,37 @@ auto comm_thread::build()
...
+-> void {
+    boost::asio::io_service io_context;
+
+    boost::asio::ip::tcp::socket mysocket(io_context);
+    boost::asio::ip::tcp::endpoint endpoint(
+            boost::asio::ip::address::from_string("127.0.0.1"), 9999);
+    boost::system::error_code ec;
+    mysocket.connect(endpoint, ec);
+    if (ec) {
+
```
sfoster1: this can be 
```
std::cerr << "Failed to create socket: " << ec.category().name() << ": " << ec.value() << std::endl;
exit(ec.value()); // this should be from <cstdlib>
```

stm32-modules/heater-shaker/simulator/comm_thread.cpp
```diff
@@ -39,16 +40,37 @@ auto comm_thread::build()
...
+    }
+
+    char pBuff[30];
+    boost::asio::mutable_buffer buff(pBuff, sizeof(pBuff));
+    std::string tot;
+
+    std::size_t l = mysocket.read_some(buff);
```
sfoster1: this probably wants to be inside the loop. the way this is typically structured in c++ would be
```c++
std::size_t bytes_read = 0;
do {
   l = mysocket.read_some(buff);
   ...
} while (l > 0);
```
`do...while` is a loop structure that works the same as `while` but checks the condition at the end of the loop body rather than the beginning. It would require making sure the loop is resilient to a zero-byte read, however.

sfoster1 (changes requested):
This is looking pretty good! Some inline stuff about error code handling.

In terms of structure and supporting different input strategies,

Configuration:
The right way to do configuration is, as in programs in other languages, command line arguments. There are facilities for this [built into libgcc](https://www.gnu.org/software/libc/manual/html_node/Getopt.html); do not use them they are awful. A commonly used c++ one that is also awful but not as bad as getopt is [boost::program_options](https://www.boost.org/doc/libs/1_63_0/doc/html/program_options.html), which should work for us (it will need some cmake twiddling though). That link should have a tutorial section. The normal way of handling an input selection like this is to do one of the following:
- have mutually-exclusive multiple options (i.e. `--listen-on=(url)` for a socket or `--stdin` or something, and you can't say both)
- require explicit protocol selection where protocol here means "some text that ends with :// before the rest of the value starts", i.e. `tcp://127.0.0.1:9999` or `pipe://stdin` (well, more usually `file://-` because `-` is a usual shorthand for "listen to stdin")
- or just "figure it out", aka have a positional argument and uh see if it's exactly `-`, in which case stdin; see if `boost::asio::ip::address::from_string()` accepts it, in which case socket; otherwise assume it's a file
The third is kind of the most user friendly but also annoying; the middle one is an middle ground that everybody hates; I'd go with the first

Architecture
Pull this out of `comm_thread.cpp`. Make this decision in something called by `main` _before_ control is given to the simulator. Pass the comm thread something with a nice api for reading and writing, like a python `TextIO`. Boost might have one of these.

Hopefully this helps! Feel free to ask anything.

stm32-modules/heater-shaker/simulator/cli_parser.cpp
```diff
@@ -0,0 +1,61 @@
+#include "simulator/cli_parser.hpp"
+#include <boost/program_options.hpp>
+#include <iostream>
+
+using namespace cli_parser;
+
+void no_options_specified_error(boost::program_options::options_description desc) {
+    std::cout << std::endl << "ERROR: You must provide either the --stdin OR the --socket option." << std::endl << std::endl;
```
sfoster1: probably wants to be `cerr`

stm32-modules/heater-shaker/simulator/cli_parser.cpp
```diff
@@ -0,0 +1,61 @@
...
+    std::cout << desc << std::endl;
+    exit(1);
+}
+
+void both_drivers_specified_error(boost::program_options::options_description desc) {
+    std::cout << std::endl << "ERROR: You may only provide either the --stdin OR the --socket option, not both." << std::endl << std::endl;
```
sfoster1: `cerr`

stm32-modules/heater-shaker/simulator/cli_parser.cpp
```diff
@@ -0,0 +1,61 @@
...
+    std::cout << desc << std::endl;
+    exit(1);
+}
+
+void neither_driver_error(boost::program_options::options_description desc) {
+    std::cout << std::endl << "ERROR: Neither --socket or --stdin was specified";
```
sfoster1: `cerr`

sfoster1 (commented):
Looking good!

stm32-modules/heater-shaker/simulator/cli_parser.cpp
DerekMaggio: This puts it on stderr instead of stdout right?

stm32-modules/heater-shaker/simulator/comm_thread.cpp
DerekMaggio: I left this as a regular `while` loop because having it as a `do while` was causing it to only read a single G-Code and I couldn't figure out why

stm32-modules/heater-shaker/include/simulator/cli_parser.hpp
```diff
@@ -0,0 +1,8 @@
+#include <memory>
+#include <string>
+
+#include "simulator/sim_driver.hpp"
+
+namespace cli_parser {
+sim_driver::SimDriver* get_sim_driver(int, char**);
```
sfoster1: This should be either a [shared_ptr](https://en.cppreference.com/w/cpp/memory/shared_ptr) or [unique_ptr](https://en.cppreference.com/w/cpp/memory/unique_ptr) based on its desired usage pattern and the semantics of the object. Raw pointers should only be used for either
- specific things that generally center around the desire to represent a specific memory address (rather than an object that has previously been allocated on the heap)
- system level requirements out of your control (no dynamic memory management allowed, as in firmware; foreign-language bindings that don't mesh well with c++ smart pointers)

The general rule is that if there should only be one owner and accessor of a given object _at a time_ - changing who owns it is fine, but not not multiple accesses - it should be a unique pointer; if multiple things may share access, shared pointer.

Use of shared pointers, which depend on reference counting generated by the compiler, completely avoid some large classes of bugs, and should always be preferred.

stm32-modules/heater-shaker/include/simulator/comm_thread.hpp
```diff
@@ -4,11 +4,13 @@
...
 #include "heater-shaker/tasks.hpp"
+#include "simulator/sim_driver.hpp"
 #include "simulator/simulator_queue.hpp"
 
 namespace comm_thread {
 using SimCommTask = host_comms_task::HostCommsTask<SimulatorMessageQueue>;
 struct TaskControlBlock;
 auto build() -> tasks::Task<std::unique_ptr<std::jthread>, SimCommTask>;
-auto handle_input(tasks::Tasks<SimulatorMessageQueue>& tasks) -> void;
+void handle_input(sim_driver::SimDriver* driver,
```
sfoster1: Given the above about smart pointers, either
- If `handle_input` is expected to own the simdriver object, it should take a `std::unique_ptr<SimDriver>&&` that is moved-into (i.e., you call this like `handle_input(std::move(driver))`)
- if `handle_input` is expected to share ownership of the simdriver object, it should take a `std::shared_ptr<SimDriver>` (by value, to trigger the reference counting)

stm32-modules/heater-shaker/include/simulator/sim_driver.hpp
```diff
@@ -0,0 +1,42 @@
...
+#include "simulator/simulator_queue.hpp"
+
+namespace sim_driver {
+const std::string SOCKET_DRIVER_NAME = "Socket";
+const std::string STDIN_DRIVER_NAME = "Stdin";
+class SimDriver {
+    std::string name;
+
+  public:
+    virtual std::string get_name() = 0;
```
sfoster1: probably should return a ref to the name, and be a [const member function](https://en.cppreference.com/w/cpp/language/member_functions#const-_and_volatile-qualified_member_functions): `virtual const std::string& get_name() const = 0`

stm32-modules/heater-shaker/include/simulator/sim_driver.hpp
```diff
@@ -0,0 +1,42 @@
...
+};
+
+class SocketSimDriver : public SimDriver {
+    std::string name = SOCKET_DRIVER_NAME;
+    std::string host;
+    int port;
+
+  public:
+    SocketSimDriver(std::string);
+    std::string get_host();
```
sfoster1: Consider, just as general good practice, bundling this into a 
```
struct AddrInfo {
   std::string host;
   uint32_t port;
};

const AddrInfo& get_port() const;
```

or use a struct that carries this information that's already defined by boost or stl if it exists.

stm32-modules/heater-shaker/include/simulator/sim_driver.hpp
```diff
@@ -0,0 +1,42 @@
...
+#include "heater-shaker/host_comms_task.hpp"
+#include "heater-shaker/messages.hpp"
+#include "heater-shaker/tasks.hpp"
+#include "simulator/simulator_queue.hpp"
+
+namespace sim_driver {
+const std::string SOCKET_DRIVER_NAME = "Socket";
+const std::string STDIN_DRIVER_NAME = "Stdin";
+class SimDriver {
+    std::string name;
```
sfoster1: Let's keep this a pure virtual base class (i.e. only pure virtual functions and nothing else) and let the storage of the `name` member be somethign decided by the children

stm32-modules/heater-shaker/simulator/cli_parser.cpp
```diff
@@ -0,0 +1,92 @@
+#include "simulator/cli_parser.hpp"
+
+#include <boost/program_options.hpp>
+#include <iostream>
+
+#include "simulator/sim_driver.hpp"
+
+using namespace cli_parser;
+
+void no_options_specified_error(
```
sfoster1: These probably should be specified as [noreturn](https://en.cppreference.com/w/cpp/language/attributes/noreturn): `[[noreturn]] void no_options_specified_error(...)`

stm32-modules/heater-shaker/simulator/cli_parser.cpp
```diff
@@ -0,0 +1,92 @@
...
+    boost::program_options::options_description desc) {
+    std::cerr
+        << std::endl
+        << "ERROR: You must provide either the --stdin OR the --socket option."
+        << std::endl
+        << std::endl;
+    std::cout << desc << std::endl;
```
sfoster1: should probably be cerr as well

stm32-modules/heater-shaker/simulator/cli_parser.cpp
```diff
@@ -0,0 +1,92 @@
...
+        use_socket = true;
+    }
+    if (num_args <= 1) {
+        no_options_specified_error(desc);
+    }
+    if (use_stdin && use_socket) {
+        both_drivers_specified_error(desc);
+    }
+
+    sim_driver::SimDriver* sim_driver;
```
sfoster1: here is where we'd create the shared or unique ptr. If it's unique, this function will return `std::unique_ptr<SimDriver>&&` and you'll return as `return std::move(sim_driver);`. If it's shared_ptr, you can return it by vaue,.

stm32-modules/heater-shaker/simulator/cli_parser.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+    if (use_stdin) {
+        sim_driver = new sim_driver::StdinSimDriver();
+    } else if (use_socket) {
+        sim_driver =
+            new sim_driver::SocketSimDriver(vm["socket"].as<std::string>());
```
sfoster1: This is the reason we need to use shared/unique pointers. Here, you create a sim driver with `new` and then return it. But where is it `delete`d? It doesn't appear to be in this code. That's a memory leak, and while it's probably fine in this instance because this essentially has the same lifetime as the program - precisely one is created when the program starts, and it's (implicitly) valid until the program ends, and we never create another - it's still good practice to clean it up.

stm32-modules/heater-shaker/simulator/comm_thread.cpp
```diff
@@ -38,17 +41,7 @@ auto comm_thread::build()
...
-        if (!std::cin.getline(linebuf->data(), linebuf->size() - 1, '\n')) {
-            return;
-        }
-        auto wrote_to = std::cin.gcount();
-        linebuf->at(wrote_to - 1) = '\n';
-        auto message = messages::IncomingMessageFromHost(
-            linebuf->data(), linebuf->data() + wrote_to);
-        static_cast<void>(tasks.comms->get_message_queue().try_send(message));
-    }
+void comm_thread::handle_input(sim_driver::SimDriver* driver,
```
sfoster1: This would want to be a smart pointer

stm32-modules/heater-shaker/simulator/sim_driver.cpp
```diff
@@ -0,0 +1,101 @@
...
+sim_driver::SocketSimDriver::SocketSimDriver(std::string url) {
+    std::regex url_regex(":\\/\\/([a-zA-Z0-9.]*):(\\d*)$");
+    std::smatch url_match_result;
+
+    if (std::regex_search(url, url_match_result, url_regex)) {
+        host = url_match_result[1];
+        port = std::stoi(url_match_result[2]);
+
+    } else {
+        std::cerr << "Malformed url." << std::endl;
```
sfoster1: It's good practice to only call `exit` from as close to the top level as you can, or at least one designated place. Here, consider throwing an exception (probably a custom one) or doing this parse at a higher level, where you parse the arguments, and making sure that we only get here if it's valid.

stm32-modules/heater-shaker/simulator/sim_driver.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+std::string sim_driver::SocketSimDriver::get_host() { return this->host; }
+
+int sim_driver::SocketSimDriver::get_port() { return this->port; }
+
+std::string sim_driver::SocketSimDriver::get_name() { return this->name; }
+
+void sim_driver::SocketSimDriver::write() {}
+
+auto get_socket(std::string host, int port) {
```
sfoster1: Consider the relative lifetimes of the socket and the `SocketSimDriver` object. It's reasonable to tie the acquisition of the socket for `bind()` to the lifetime of the `SocketSimDriver`; this is a c++ pattern known as [RAII](https://en.cppreference.com/w/cpp/language/raii), "resource acquisition is initialization". 

It becomes a little more complex for receive-side sockets (more on that later) because a receive-side BSD socket has _two_ lifetimes associated with it:
- the `bind()` lifetime, which lasts while the program can accept incoming connections (or, rather, is ready to handle incoming connection _requests_ - it's allowed to reject them while maintaining the `bind()`)
- the `listen()` lifetime, which is the same as the lifetime of an individual connection.

The typical pattern is to have a socket manager own the bind, and create and spin off new objects for each connection.

Here, we'll only have one connection - we're emulating a single physical wire, after all - so I don't think we need two objects, but the different lifetimes still have different semantics, especially since there could be a connection that starts, and ends, and then restarts, etc. 

A good pattern might be to essentially replace `get_socket` with a class that RAII's a bind-socket, so you have like
```
class Socket {
   explicit Socket(AddrInfo addr) {
      sock = bind(addr.host, addr.port);
      if(!sock) { // or whatever error signalling mechanism
        throw std::runtime_error("Could not acquire socket ");
      }
   }
   ~Socket() {
        sock.destroy(); // or whatever teardown
    }
   Socket sock;
};
```

At this point, you can replace a call to `get_socket` with just creating a `Socket`, and you can trust that whenever the function ends, whether through return or exception, this object will get destroyed (because it's created by-value in the function) and the socket will get cleaned up.

You can then also have methods to listen or whatever.
and replace calls to `get_socket` with just building s.a clas

stm32-modules/heater-shaker/simulator/sim_driver.cpp
```diff
@@ -0,0 +1,101 @@
...
+    boost::asio::io_service io_context;
```
sfoster1: If you made the `Socket` class discussed above actually own the _boost asio socket_, you could wrap all this up with similar initialization/destruction guarantees.

stm32-modules/heater-shaker/simulator/sim_driver.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+    boost::asio::ip::tcp::socket socket(io_context);
+    boost::asio::ip::tcp::endpoint endpoint(
+        boost::asio::ip::address::from_string(host), port);
+    boost::system::error_code ec;
+    socket.connect(
+        endpoint,
+        ec);  // TODO: Change this to socket.bind - Derek Maggio 9/30/21
```
sfoster1: Definitely needs to switch to a listen-side socket before we merge it.

stm32-modules/heater-shaker/simulator/sim_driver.cpp
```diff
@@ -0,0 +1,101 @@
...
+    auto linebuf = std::string(1024, 'c');
+    boost::asio::mutable_buffer buff(pBuff, sizeof(pBuff));
+    std::string read_data;
+
+    /*
+     * TODO: This algorithm starts dropping G-Codes if they are sent faster than
+     * once every 0.005 seconds. Need to handle that Derek Maggio 9/30/21
+     */
+    std::size_t l = socket.read_some(buff);
+    while (l > 0) {
```
sfoster1: one thing you could do is make this `while (size_t l = socket.read_some(buff), l!=0)` which should work the same but let you get rid of line 80

stm32-modules/heater-shaker/simulator/sim_driver.cpp
```diff
@@ -0,0 +1,101 @@
...
+        read_data.append(static_cast<const char*>(buff.data()), l);
+        std::size_t pos = read_data.find("\n");
+
+        while (pos != std::string::npos) {
```
sfoster1: same deal here - this could be `while (std::size_t pos = read_data.find("\n"); pos != std::string::npos)`

stm32-modules/heater-shaker/simulator/sim_driver.cpp
```diff
@@ -0,0 +1,101 @@
...
+        std::cerr << "Failed to create socket: " << ec.category().name() << ": "
+                  << ec.value() << std::endl;
+        exit(ec.value());
+    }
+    return socket;
+}
+
+void sim_driver::SocketSimDriver::read(
+    tasks::Tasks<SimulatorMessageQueue>& tasks) {
+    char pBuff[30];
```
sfoster1: why 30?

stm32-modules/heater-shaker/simulator/sim_driver.cpp
```diff
@@ -0,0 +1,101 @@
...
+    auto socket = get_socket(this->host, this->port);
+    auto linebuf = std::string(1024, 'c');
```
sfoster1: Data this big shouldn't be stored on the stack (allocated inline). This should at some point be created with a `new` (i.e., actually a `shared_ptr`).

stm32-modules/heater-shaker/simulator/sim_driver.cpp
```diff
@@ -0,0 +1,101 @@
...
+     * once every 0.005 seconds. Need to handle that Derek Maggio 9/30/21
+     */
+    std::size_t l = socket.read_some(buff);
+    while (l > 0) {
+        read_data.append(static_cast<const char*>(buff.data()), l);
+        std::size_t pos = read_data.find("\n");
+
+        while (pos != std::string::npos) {
+            std::string msg = read_data.substr(0, pos + 1);
+            linebuf.replace(0, msg.length(), msg);
```
sfoster1: This uses, I believe, 4 different string-storage buffers:
- a `boost::asio::mutable_buffer buff` backed by `char[30] cbuff` written to directly by the socket
- a `std::string read_data` that has all the data from the last socket read copied into it from `buff` (after which `buff` is erased), using `append` with a cstring, dynamic size
- a `std::string msg` which is a `substr` from `read_data` from its beginning to its first newline, dynamic size
- a `std::string linebuf(1024...)`, fixed size, which has the data from `msg` copied to its beginning and is a storage location for the message that goes into the simulator

This is a lot of buffers, we should cut down how many there are. We should have a bigger buffer given to the socket, just in case; and then we should have a double buffer (there's an implementation in `include/common` that is given to the message. We need to use a double buffer here because the `IncomingMessageFromHost` needs pointers into a linear buffer rather than a circular buffer, and we need it to live long enough for the message to handle. Look at how `freertos_host_comms_task` does it (definitely not perfect, but works ok).

We should also see what happens when we send it a bunch more data in a chunk. Doesn't have to handle it nicely but shouldn't crash.

sfoster1 (changes requested):
Definitely a solid structure. I think the things we should look at fall into a couple categories:
- Use smart pointers
- Use RAII for the sockets
- Fix up some logic and storage
- Use a listen-side socket

And then finally, one more-structural thing: if I'm going to send something data via a socket, I expect it to send me responses via that socket. The SimDriver should handle output as well as input, and the SocketSimDriver should send responses via the socket it received its inputs into.

Overall, this is a _lot_ for a first PR, and it's definitely better than my first PR! Nice work!

amitlissack:
> * Use a listen-side socket

Curious. Why a listen side socket? I feel I made a mistake making the python module and smoothie emulators listeners. I believe the robot server should listen.

stm32-modules/heater-shaker/simulator/sim_driver.cpp
DerekMaggio: Idk. Because that was the value that was in the snippet @amitlissack gave me.

stm32-modules/heater-shaker/include/simulator/sim_driver.hpp
DerekMaggio: https://github.com/Opentrons/opentrons-modules/pull/203/commits/9f206046b43e601875be34d9f867b674d0bf0fd6

stm32-modules/heater-shaker/include/simulator/sim_driver.hpp
DerekMaggio: https://github.com/Opentrons/opentrons-modules/pull/203/commits/9f206046b43e601875be34d9f867b674d0bf0fd6

stm32-modules/heater-shaker/simulator/cli_parser.cpp
DerekMaggio: https://github.com/Opentrons/opentrons-modules/pull/203/commits/b833906af8518f2a39ebae76c6b6e9494015b1b7

stm32-modules/heater-shaker/simulator/cli_parser.cpp
DerekMaggio: https://github.com/Opentrons/opentrons-modules/pull/203/commits/b833906af8518f2a39ebae76c6b6e9494015b1b7

stm32-modules/heater-shaker/simulator/sim_driver.cpp
DerekMaggio: https://github.com/Opentrons/opentrons-modules/pull/203/commits/72871e8e2a60c7dd7c8de0685d31952cfd852613

stm32-modules/heater-shaker/simulator/sim_driver.cpp
DerekMaggio: https://github.com/Opentrons/opentrons-modules/pull/203/commits/72871e8e2a60c7dd7c8de0685d31952cfd852613

sfoster1:
> > * Use a listen-side socket
> 
> Curious. Why a listen side socket? I feel I made a mistake making the python module and smoothie emulators listeners. I believe the robot server should listen.

Yeah that's a point I hadn't thought of. I guess the tradeoff would be that keeping it as a connection-initiator makes it a little more annoying to use standalone, but it does seem like it _should_ be a robot-is-central arrangement.

stm32-modules/heater-shaker/include/simulator/cli_parser.hpp
DerekMaggio: c2796ad1477d7dfc075be13eb98e7e3a2efef070

stm32-modules/heater-shaker/simulator/sim_driver.cpp
DerekMaggio: 0029e0e22b3c22ccf10005c73fd2d8f57e0f5a9a

stm32-modules/heater-shaker/include/simulator/comm_thread.hpp
DerekMaggio: c2796ad1477d7dfc075be13eb98e7e3a2efef070

stm32-modules/heater-shaker/include/simulator/sim_driver.hpp
DerekMaggio: 573d9e03165564c8053a11a52dace480986f5d47

stm32-modules/heater-shaker/simulator/socket_sim_driver.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+const std::string& socket_sim_driver::SocketSimDriver::get_name() const {
+    return this->name;
+}
+
+void socket_sim_driver::SocketSimDriver::write(std::string message) {
+    this->s->write_some(boost::asio::buffer(message));
+}
+
+int has_char(char* char_array, char value_to_find) {
```
sotaoverride, line 66: we could also use std::string::find?

stm32-modules/heater-shaker/include/simulator/socket_sim_driver.hpp
```diff
@@ -0,0 +1,32 @@
...
+    static const std::string name;
+    AddressInfo address_info;
+    std::unique_ptr<boost::asio::ip::tcp::socket> s;
+
+  public:
+    SocketSimDriver(std::string);
+    std::string get_host();
+    int get_port();
+    const std::string& get_name() const;
+    std::unique_ptr<boost::asio::ip::tcp::socket> get_socket();
```
sfoster1: this probably doesn't want to be `public` if it returns a unique pointer by-value. That means that calling it will implicitly strip ownership of the pointer from `socket_sim_driver`. I think this is used internally in the constructor, right? let's either (a) make it private or (b) make it just a free function defined in the `socket_sim_driver.cpp` file and not exposed outside of the class at all.

stm32-modules/heater-shaker/include/simulator/socket_sim_driver.hpp
```diff
@@ -0,0 +1,32 @@
...
+};
+
+class SocketSimDriver : public sim_driver::SimDriver {
+    static const std::string name;
+    AddressInfo address_info;
+    std::unique_ptr<boost::asio::ip::tcp::socket> s;
+
+  public:
+    SocketSimDriver(std::string);
+    std::string get_host();
```
sfoster1: probably could be `const std::string& get_host() const;`

stm32-modules/heater-shaker/include/simulator/socket_sim_driver.hpp
```diff
@@ -0,0 +1,32 @@
...
+    int get_port();
+    const std::string& get_name() const;
+    std::unique_ptr<boost::asio::ip::tcp::socket> get_socket();
+
+    void write(std::string message);
```
sfoster1: take a `const std::string&`?

stm32-modules/heater-shaker/simulator/socket_sim_driver.cpp
```diff
@@ -0,0 +1,99 @@
...
+    }
+}
+
+const std::string socket_sim_driver::SocketSimDriver::name = SOCKET_DRIVER_NAME;
+
+std::string socket_sim_driver::SocketSimDriver::get_host() {
+    return this->address_info.host;
+}
+
+int socket_sim_driver::SocketSimDriver::get_port() {
```
sotaoverride: this can also be a const function?

stm32-modules/heater-shaker/include/simulator/socket_sim_driver.hpp
```diff
@@ -0,0 +1,32 @@
...
+
+class SocketSimDriver : public sim_driver::SimDriver {
+    static const std::string name;
+    AddressInfo address_info;
+    std::unique_ptr<boost::asio::ip::tcp::socket> s;
+
+  public:
+    SocketSimDriver(std::string);
+    std::string get_host();
+    int get_port();
```
sfoster1: probably could be `int get_port() const;`

stm32-modules/heater-shaker/include/simulator/socket_sim_driver.hpp
sfoster1: If it _is_ meant to be an accessor for the socket, then honestly it probably shoudln't be a `unique_ptr`. `unique_ptr`s should typically only be accessed by the thing that owns them (where "owning" is a semantic concept inside the program, not "literally the object which the pointer is a member of"

sfoster1 (approved):
Awesome that it works and I'm good with the TODOs, some slight cleanup stuff around const-ness but overall great!

sotaoverride (approved):
v.cool!!

stm32-modules/heater-shaker/include/simulator/socket_sim_driver.hpp
DerekMaggio: You know what's funny, I defined this function in the header and never provided an implementation. I am just loading the value directly. So I think I can just take this line out

DerekMaggio: 0a57994a32fda6c1be9343eca04130a6ac7781bb

stm32-modules/heater-shaker/include/simulator/socket_sim_driver.hpp
DerekMaggio: 0a57994a32fda6c1be9343eca04130a6ac7781bb

stm32-modules/heater-shaker/include/simulator/socket_sim_driver.hpp
DerekMaggio: 0a57994a32fda6c1be9343eca04130a6ac7781bb

stm32-modules/heater-shaker/include/simulator/socket_sim_driver.hpp
DerekMaggio: 0a57994a32fda6c1be9343eca04130a6ac7781bb

stm32-modules/heater-shaker/simulator/socket_sim_driver.cpp
DerekMaggio: 0a57994a32fda6c1be9343eca04130a6ac7781bb

stm32-modules/heater-shaker/include/simulator/socket_sim_driver.hpp
DerekMaggio: 0a57994a32fda6c1be9343eca04130a6ac7781bb

PR #204, opened 2021-09-28
chore(heater-shaker): Add : separating key and val

sfoster1:
The OT-2 software is structured to look for response formats like
key:value rather than keyvalue like we have here (and it's nicer this
way anyway). Add a : character in between the key and value; for
instance, a temperature message is now M123 C:10.00 T:20.01 rather than
M123 C10.00 T20.01.


pmoegenburg (approved):
lgtm

PR #206, opened 2021-09-30
fix(heater_shaker): serial numbers up to 23 characters

pmoegenburg:
first two addresses of serial number flash were getting erased in previous code

### DIFF
--- a/stm32-modules/heater-shaker/firmware/system/system_serial_number.h
+++ b/stm32-modules/heater-shaker/firmware/system/system_serial_number.h
@@ -8,7 +8,7 @@ extern "C" {
 #include <stdbool.h>
 #include <stddef.h>
 
-bool system_set_serial_number(uint64_t to_write, uint8_t address);
+bool system_set_serial_number(uint64_t to_write1, uint64_t to_write2, uint64_t to_write3);
 
 uint64_t system_get_serial_number(uint8_t address);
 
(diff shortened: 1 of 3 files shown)


stm32-modules/heater-shaker/firmware/system/system_serial_number.h
```diff
@@ -8,7 +8,7 @@ extern "C" {
 #include <stdbool.h>
 #include <stddef.h>
 
-bool system_set_serial_number(uint64_t to_write, uint8_t address);
+bool system_set_serial_number(uint64_t *to_write);
```
sfoster1: since we're not really constrained on code speed or size, can we make this more like

```c
struct writable_serial {
   uint64_t contents[3];
};

bool system_set_serial_number(struct writable_serial to_write);
// or
bool system_set_serial_number(struct writable_serial* to_write);
``` 

that plays a bit better with static analysis since it's now a bounded array; it's in a struct so it doesn't [decay](https://stackoverflow.com/questions/1461432/what-is-array-to-pointer-decay)

sfoster1 (changes requested):
Looks good functionality wise (nice work writing it so that this change is easy!) but can we make the size a little more explicit on the c interface side (see comment)

sfoster1 (approved):
lgtm!

PR #207, opened 2021-10-11
fix(heater-shaker/simulator): no hardcoded error

sfoster1:
The motor subsystem used to use error detection as a mechanism to know
when it homed. That's not true any more, but we didn't change the simulator.


DerekMaggio (approved):
Tested with the Heater Shaker Emulator. 
Built and ran emulator.
Ran `G28`, got back `G28 OK`.
Ran `M242` to open plate lock, this was previously not working, got back `M242 OK`.
Ran `M243` to close plate lock, this was previously not working, got back `M243 OK`.

PR #210, opened 2021-10-21
feat(heater-shaker): added plate lock motor timeout

pmoegenburg:


stm32-modules/heater-shaker/firmware/motor_task/freertos_motor_task.cpp
```diff
@@ -79,8 +79,17 @@ void run(void *param) {
     memset(&_local_task.handles, 0, sizeof(_local_task.handles));
     _local_task.handles.plate_lock_complete = handle_plate_lock;
     motor_hardware_setup(&_local_task.handles);
-
     auto policy = MotorPolicy(&_local_task.handles);
+    auto &queue = _task.get_message_queue();
+    // ensure plate lock closed via message at startup (needed for homing)
+    auto message1 = messages::ClosePlateLockMessage{.from_startup = true};
+    static_cast<void>(
+        queue.try_send(message1, 1500));  // adjust ticks to ensure fully closed
```
sfoster1: this time is the timeout before an error, it's not a delay before sending

sfoster1: and since the task's not actually running at this time, no matter what number you put here the task is going to have these two messages immediately next to each other in the queue. so it will see closeplatelock, then home, on subsequent calls to `try_once`. this is fine if the `ClosePlateLock` message execution blocks until it's done, but otherwise it'll start homing before the close finishes.

sfoster1 (changes requested):
We definitely need tests for the state management here. We should test that the right messages are sent when it gets an open/close, including in the case where it's already opened or closed. We should test the timeout as well.

Also, I don't think the at-boot behavior quite works right. Because the task isn't being called when the two messages are sent, when the `run_once` loop happens you'll have the `ClosePlateLock` and `Home` messages in the queue right after each other, and pull them on subsequent calls to `run_once`. That means that they'll both start executing.

We should try and test that logic - either duplicate it in a test and check the behavior, or have a free function in the motor task header that the test and the firmware can both call - and make sure the state machines move the way we want.

stm32-modules/heater-shaker/firmware/motor_task/freertos_motor_task.cpp
pmoegenburg: updated startup routine. I think we're mechanically unable to home when plate lock is open, but mechanically able to close plate lock (huge tolerance on optical_switch:flag interface) when not homed. I commented out an error I had the idea of creating if plate lock is open when trying to home. I've heard from hardware guys that it's best to home before moving plate lock, but if we can't close plate lock before homing (when needed), we're in a catch-22, so we close plate lock if needed before homing at startup.

pmoegenburg: This new startup routine tested positively on physical module

sfoster1: Did you try it with the plate lock driven open and held by inertia when the system booted?

pmoegenburg: yup, plate lock first closes, then main motor homes

sfoster1 (approved):
All right! nice work.

PR #212, opened 2021-10-26
feat(heater-shaker): adding LED basic functionality

pmoegenburg:


stm32-modules/heater-shaker/firmware/system/system_hardware.c
```diff
@@ -60,3 +87,209 @@ asm volatile (
...
+  HAL_GPIO_Init(I2Cx_SCL_GPIO_PORT, &GPIO_InitStruct);
+    
+  /* I2C RX GPIO pin configuration  */
+  GPIO_InitStruct.Pin       = I2Cx_SDA_PIN;
+  GPIO_InitStruct.Alternate = I2Cx_SCL_SDA_AF;
+  HAL_GPIO_Init(I2Cx_SDA_GPIO_PORT, &GPIO_InitStruct);
+    
+  /*##-4- Configure the NVIC for I2C ########################################*/   
+  /* NVIC for I2Cx */
+  HAL_NVIC_SetPriority(I2Cx_ER_IRQn, 0, 1);
```
sfoster1: These are really high priority and probably won't play well with calling freertos functions from within this ISR. we should try setting them to around I think 10 or so? The MAX_LIBRARY_PRIORITY value from freertoscofnig.

stm32-modules/heater-shaker/firmware/system/system_hardware.c
```diff
@@ -60,3 +87,209 @@ asm volatile (
...
+  HAL_GPIO_DeInit(I2Cx_SDA_GPIO_PORT, I2Cx_SDA_PIN);
+  
+  /*##-3- Disable the NVIC for I2C ##########################################*/
+  HAL_NVIC_DisableIRQ(I2Cx_ER_IRQn);
+  HAL_NVIC_DisableIRQ(I2Cx_EV_IRQn);
+}
+
+//bool system_hardware_set_led(uint8_t aTxBuffer[SYSTEM_WIDE_TXBUFFERSIZE], I2C_Operations operation) {
+bool system_hardware_set_led(uint8_t* aTxBuffer, I2C_Operations operation) {
+  //loop thru bits
```
sfoster1: these probably shouldn't be checked in I think - maybe store them as notes elsewhere?

stm32-modules/heater-shaker/firmware/system/system_hardware.c
```diff
@@ -60,3 +87,209 @@ asm volatile (
...
+
+  //check all files in stm32tools template project
+
+  //do transmit like main.c lines 132-154? How does that syntax work?
+  //Should error check. Use HAL_I2C_ErrorCallback()?!
+  //won't callback signal end of transmission?
+  //returns bool for transmit_start_success
+
+  uint16_t base_register = 0;
+  uint8_t UpdateBuffer[1] = {};
```
sfoster1: [nit] let's be consistent with variable names - we're a mix of `camelCase`, `PascalCase`, and `snake_case` right now. I tend to like `snake_case` for variable names but really the big thing is consistency, whatever it is.

stm32-modules/heater-shaker/firmware/system/system_hardware.c
```diff
@@ -60,3 +87,209 @@ asm volatile (
...
+
+  HAL_StatusTypeDef status = HAL_I2C_Mem_Write_IT(&I2cHandle, (uint16_t)I2C_ADDRESS, base_register,
+    (uint16_t)REGISTER_SIZE, (uint8_t*)aTxBuffer, (uint16_t)SYSTEM_WIDE_TXBUFFERSIZE);
+  if ((status == HAL_OK) && (operation == LED_Control)) { //only update after LED control transmit at initialization
+    status = HAL_I2C_Mem_Write_IT(&I2cHandle, (uint16_t)I2C_ADDRESS, (uint16_t)UPDATE_REGISTER,
+      (uint16_t)REGISTER_SIZE, (uint8_t*)UpdateBuffer, (uint16_t)(sizeof(UpdateBuffer)));
+  }
+  return (status == HAL_OK);
+
+/*HAL_StatusTypeDef HAL_I2C_Mem_Write_IT(I2C_HandleTypeDef *hi2c, uint16_t DevAddress, uint16_t MemAddress,
```
sfoster1: should probably remove

stm32-modules/heater-shaker/firmware/system/system_hardware.c
```diff
@@ -60,3 +87,209 @@ asm volatile (
...
+  } else {
+    result.success = false;
+  }
+  SYSTEM_HW_HANDLE->led_transmit_complete(&result);
+}
+
+/**
+  * @brief  I2C error callbacks.
+  * @param  I2cHandle: I2C handle
+  * @note   This example shows a simple way to report transfer error, and you can
```
sfoster1: update or remove

stm32-modules/heater-shaker/firmware/system/system_hardware.c
```diff
@@ -60,3 +87,209 @@ asm volatile (
...
+  *         add your own implementation.
+  * @retval None
+  */
+void HAL_I2C_ErrorCallback(I2C_HandleTypeDef *I2cHandle)
+{
+  /** 1- When Slave don't acknowledge it's address, Master restarts communication.
```
sfoster1: should probaly update or remove

stm32-modules/heater-shaker/firmware/system/system_hardware.c
```diff
@@ -60,3 +87,209 @@ asm volatile (
...
+    */
+  led_transmit_result result;
+  result.success = false; //need?
+  if (HAL_I2C_GetError(I2cHandle) != HAL_I2C_ERROR_AF) { //does this signal error?!
+    result.error = true; //can we get specific error?
+  }
+  SYSTEM_HW_HANDLE->led_transmit_error_complete(&result);
+}
+
+//*********** was in example project, but think including is wrong and cause of build errors *************
```
sfoster1: yes, this stuff shoudln't be here, just the I2C stuff. let's remove it.

stm32-modules/heater-shaker/firmware/system/system_stm32f3xx.c
```diff
@@ -177,6 +177,31 @@ void SystemInit(void)
   SCB->CPACR |= ((3UL << 10*2)|(3UL << 11*2));  /* set CP10 and CP11 Full Access */
 #endif
 
+/* Reset the RCC clock configuration to the default reset state ------------*/
+/* Set HSION bit */
+RCC->CR |= 0x00000001U;
```
sfoster1: Where'd this stuff come from? Do we really want it? Is it to support soft reset better? If so, we should note what it's needed for.

stm32-modules/heater-shaker/include/heater-shaker/gcodes.hpp
```diff
@@ -676,6 +675,68 @@ struct SetSerialNumber {
     }
 };
 
+struct SetLED { //make sure M998 isn't used. Just use this to turn them on, cycle power to reset
```
sfoster1: bit of a nit, but could this be `M998.D`? I think this is more of a debug gcode because we probably won't let the higher levels set things this directly - it'll be built into internal state changes, or more generic commands with semantics like "identify".

stm32-modules/heater-shaker/include/heater-shaker/gcodes.hpp
```diff
@@ -676,6 +675,68 @@ struct SetSerialNumber {
...
+    requires std::forward_iterator<InputIt> &&
+        std::sized_sentinel_for<InputLimit, InputIt>
+    static auto write_response_into(InputIt buf, InputLimit limit) -> InputIt {
+        return write_string_to_iterpair(buf, limit, response);
+    }
+
+    template <typename InputIt, typename InputLimit>
+    requires std::forward_iterator<InputIt> &&
+        std::sized_sentinel_for<InputLimit, InputIt>
+    static auto write_response_into(InputIt buf, InputLimit limit, bool success) -> InputIt {
```
sfoster1: do we need a special handler here? if we reuse the acknowledgement with the error code, we can either say just `M998.D OK\n` if it succeeded or return the error code if it fails.

stm32-modules/heater-shaker/include/heater-shaker/gcodes.hpp
```diff
@@ -676,6 +675,68 @@ struct SetSerialNumber {
...
+            return std::make_pair(ParseResult(), input);
+        }
+
+        auto value_res = parse_value<uint8_t>(working, limit);
+
+        if (!value_res.first.has_value()) {
+            return std::make_pair(ParseResult(), input);
+        }
+        std::array<uint8_t, SYSTEM_WIDE_TXBUFFERSIZE> UpdateBuffer = {};
+        uint8_t index = static_cast<uint8_t>(value_res.first.value());
```
sfoster1: definitely need to bounds check this index and return a failure if you specify e.g. `255`.

stm32-modules/heater-shaker/include/heater-shaker/system_task.hpp
```diff
@@ -173,6 +174,37 @@ class SystemTask {
...
+            messages::LEDTransmitCompleteResponse{.responding_to_id = cached_led_id, 
+                                                        .success = msg.transmitted};
+        if (msg.error) {
+            response.error = errors::ErrorCode::SYSTEM_LED_TRANSMIT_COMPLETION_ERROR;
+        }
+        static_cast<void>(task_registry->comms->get_message_queue().try_send(
+            messages::HostCommsMessage(response)));
+
+        //create a new LEDTransmitCompleteResponse Host_Comms_Task message w .success and .error vars
+        //what kind of gcode output does Ack message w/o .id create? Need to print success and error vars
```
sfoster1: We should probably defer sending an ok back until the transmit is successful, i.e. make the logic  in `StartSetLEDMessage`'s visitor look like this, where we send an ack-with-error on failure and nothing on success, and then have this function be the one that sends the success message.

sfoster1: out of date comment?

sfoster1 (changes requested):
The functionality looks pretty good but there's some stuff I think we should change:
- Let's use the already existing ack-with-error for responses rather than having the response formatter take a bool
- Let's defer success responses to completion of the transmit rather than queuing the transmit
- I think the interrupt priorities are too high
- Need to clean up a lot of out-of-date comments that are notes to self or from the examples
- Tests!

stm32-modules/heater-shaker/firmware/system/stm32f3xx_hal_msp.c
```diff
@@ -0,0 +1,97 @@
+/**
+  ******************************************************************************
+  * @file    I2C/I2C_TwoBoards_ComIT/Src/stm32f3xx_hal_msp.c
```
sfoster1: We should probably get some of this boilerplate out of here

stm32-modules/heater-shaker/firmware/system/system_hardware.c
```diff
@@ -60,3 +90,149 @@ asm volatile (
   : "r" (*sysmem_boot_loc)
   : "memory"  );
 }
+
+bool system_hardware_setup_led(void) {
+  uint32_t ulNotificationValue;
+  const TickType_t xMaxBlockTime = pdMS_TO_TICKS( 100 );
+  HAL_StatusTypeDef status;
+
+  configASSERT( xTaskToNotify == NULL );
```
sfoster1: are `configAsserts` the right way to handle this? should we instead be returning early and sending an error?

stm32-modules/heater-shaker/firmware/system/system_hardware.c
```diff
@@ -60,3 +90,149 @@ asm volatile (
...
+  xTaskToNotify = xTaskGetCurrentTaskHandle();
+  status = HAL_I2C_Mem_Write_IT(&I2cHandle, ((uint16_t)I2C_ADDRESS)<<1, (uint16_t)SHUTDOWN_REGISTER,
+    (uint16_t)REGISTER_SIZE, (uint8_t*)ShutdownBuffer, (uint16_t)(sizeof(ShutdownBuffer)));
+  ulNotificationValue = ulTaskNotifyTake( pdTRUE, xMaxBlockTime );
+
+  if ((status == HAL_OK) && (ulNotificationValue == 1) && (I2cHandle.State == HAL_I2C_STATE_READY)) {
```
sfoster1: [nit] if we make this return early on error, there will be fewer nested `if`s and it'll be a bit easier to read

stm32-modules/heater-shaker/firmware/system/system_hardware.c
```diff
@@ -60,3 +90,149 @@ asm volatile (
...
   : "memory"  );
 }
+
+bool system_hardware_setup_led(void) {
+  uint32_t ulNotificationValue;
+  const TickType_t xMaxBlockTime = pdMS_TO_TICKS( 100 );
+  HAL_StatusTypeDef status;
+
+  configASSERT( xTaskToNotify == NULL );
+  xTaskToNotify = xTaskGetCurrentTaskHandle();
```
sfoster1: can we wrap blocks like this in a function?

stm32-modules/heater-shaker/firmware/system/system_hardware.h
```diff
@@ -1,16 +1,65 @@
...
 #include "stm32f3xx_hal_gpio.h"
+#include "stm32f3xx_hal.h"
 #ifdef __cplusplus
 extern "C" {
 #endif  // __cplusplus
 
+#include <stdbool.h>
+#include "systemwide.h"
+
+#define I2C_ADDRESS        0x6C
```
sfoster1: these values are only used in `system_hardware.c`, can we have them at the beginning of that file? then we can keep this file for just things that need to be shared between `system_hardware.c` and `freertos_system_task.cpp`

stm32-modules/heater-shaker/firmware/system/system_hardware.h
```diff
@@ -1,16 +1,65 @@
...
+/* I2C TIMING Register is defined when I2C clock source is SYSCLK (true) */
+/* I2C TIMING is calculated using SYSCLK = 72 MHz, I2C Speed Frequency = 100 KHz, Rise Time = 100ns, and Fall Time = 100ns */
+#define I2C_TIMING      0x00201D2B
+
+/* Definition for I2Cx clock resources */
+#define I2Cx                            I2C1
+#define RCC_PERIPHCLK_I2Cx              RCC_PERIPHCLK_I2C1
+#define RCC_I2CxCLKSOURCE_SYSCLK        RCC_I2C1CLKSOURCE_SYSCLK
+#define I2Cx_CLK_ENABLE()               __HAL_RCC_I2C1_CLK_ENABLE()
```
sfoster1: having these in this header, for instance, will probably cause compiler errors when it's included from c++

stm32-modules/heater-shaker/include/heater-shaker/gcodes.hpp
```diff
@@ -676,6 +675,116 @@ struct SetSerialNumber {
...
+        if (working != limit && !std::isspace(*working)) {
+            return std::make_pair(ParseResult(), input);
+        }
+        return std::make_pair(ParseResult(IdentifyModuleStartLED()), working);
+    }
+};
+
+struct IdentifyModuleStopLED {
+    /*
+    ** IdentifyModuleStopLED uses a random gcode, M999, adjacent to the identify
```
sfoster1: [nit] `M999` is an emergency stop handling gcode in a lot of other implementations. can we pick something else?

stm32-modules/heater-shaker/tests/test_host_comms_task.cpp
```diff
@@ -475,6 +475,88 @@ SCENARIO("message passing for ack-only gcodes from usb input") {
...
+                    tasks->get_system_queue().backing_deque.front();
+                auto set_led_message =
+                    std::get<messages::StartSetLEDMessage>(system_message);
+                tasks->get_system_queue().backing_deque.pop_front();
+                std::array<uint8_t, SYSTEM_WIDE_TXBUFFERSIZE> aTxBufferTest = {0};
+                aTxBufferTest[0] = 0x30;
+                REQUIRE(set_led_message.aTxBuffer == aTxBufferTest);
+                REQUIRE(written_firstpass == tx_buf.begin());
+                REQUIRE(tasks->get_host_comms_queue().backing_deque.empty());
+                /*AND_WHEN("sending a good response back to the comms task") {
```
sfoster1: can we either uncomment this or remove it

sfoster1 (changes requested):
The functionality looks really good! I think the big things we _have_ to fix before merging are some commented out code; left over boilerplate that can be removed; and some defines in `system_hardware.h` that could be in `system_hardware.c` and would cause compiler warnings. There's also some nitpicks around structuring some of the code.

However, the structure as a whole looks great, and I really like the use of task notifications!

sfoster1 (approved):
Awesome stuff!

PR #213, opened 2021-11-04
refactor(heater-shaker): proper plate lock status checking and added delay before homing at startup to ensure motor setup complete

pmoegenburg:


stm32-modules/heater-shaker/include/heater-shaker/motor_task.hpp
```diff
@@ -369,64 +372,83 @@ class MotorTask {
...
+        }
+        static_cast<void>(get_message_queue().try_send(check_state_message));
+    }
+
+    template <typename Policy>
+    auto visit_message(const messages::CheckPlateLockStatusMessage& msg,
+                       Policy& policy) -> void {
+        if (msg.with_error != errors::ErrorCode::NO_ERROR) {
+            static_cast<void>(
+                task_registry->comms->get_message_queue().try_send(
```
sfoster1, line 421: this probably doesn't want to happen if `msg.from_startup == true` right?

stm32-modules/heater-shaker/include/heater-shaker/motor_task.hpp
```diff
@@ -369,64 +372,83 @@ class MotorTask {
...
+                    messages::AcknowledgePrevious{
+                        .responding_to_id = msg.responding_to_id,
+                        .with_error = msg.with_error}));
+        } else if ((plate_lock_state.status == PlateLockState::IDLE_CLOSED) ||
```
sfoster1, line 425: What happens if
- Plate lock status is currently `IDLE_OPEN`
- we get a `ClosePlateLockMessage`
- this fires? Won't this trigger incorrectly because the plate lock status is still IDLE_OPEN? or is there something that's guaranteed to change the state to OPENING in between? If it the interrupt firing when the switch closes, is there a guarantee that happens before the check message is recieved? Maybe we should set the state to OPENING/CLOSING in the `OpenPlateLock` and `ClosePlateLock` message handlers

sfoster1 (changes requested):
Couple more little things but looking good

stm32-modules/heater-shaker/include/heater-shaker/motor_task.hpp
pmoegenburg: Motor shouldn't be running if from startup so we should never hit this condition. I think this is a good protection to have, but maybe don't create an error message? But want to give the user some feedback...

sfoster1: I think it's a good error message to have for the normal case. Your point is good, let's just keep it.

stm32-modules/heater-shaker/include/heater-shaker/motor_task.hpp
pmoegenburg: `                policy.plate_lock_set_power(ClosePower);
                plate_lock_state.status = PlateLockState::CLOSING;
                polling_time = 0;
`
`ClosePlateLockMessage` will enter this condition, changing state to `CLOSING`

sfoster1: Oh! I completely missed that, my bad.

pmoegenburg: First code segment is lines 408-410. Does that answer your question?

sfoster1 (approved):
LGTM! The problems I had were just things I misread. Nice work!

PR #214, opened 2021-11-08
feat(thermocycler-refresh): Initial thermocycler-refresh bringup

fsinapi:
Initial build system & code setup for the thermocycler-refresh board. Compiles a binary that turns on the debug LED on the board and then enters an infinite while loop. Host side compilation "passes" with some dummy executables, but cross compilation cmake toolchain appears to be all set.

We probably need to do some refactoring involving common code - right now everything is siloed into the two stm32 module subfolders and there's no overlap, so some code that _should_ be shared is instead duplicated. I didn't want to get too crazy restructuring the heater/shaker right now so I tried to let that folder be, except for a couple cmake changes required to get both projects to build.


stm32-modules/thermocycler-refresh/firmware/CMakeLists.txt
```diff
@@ -0,0 +1,176 @@
...
+  $<$<COMPILE_LANGUAGE:CXX>:-Wextra-semi>
+  $<$<COMPILE_LANGUAGE:CXX>:-Wctor-dtor-privacy>
+  $<$<COMPILE_LANGUAGE:CXX>:-fno-rtti>)
+
+
+target_link_libraries(
+  ${TARGET_MODULE_NAME}
+  STM32G4xx_Drivers_${TARGET_MODULE_NAME} 
+  STM32G4xx_FreeRTOS_${TARGET_MODULE_NAME} 
+  #STM32G4xx_USB_${TARGET_MODULE_NAME}
```
sfoster1, line 75: this is ok even though the include directories are added right

stm32-modules/thermocycler-refresh/firmware/system/system_hardware.c
```diff
@@ -0,0 +1,68 @@
...
+      .Pin = GPIO_PIN_6,
+      .Mode = GPIO_MODE_OUTPUT_PP,
+      .Pull = GPIO_NOPULL,
+      .Speed = GPIO_SPEED_FREQ_LOW
+    };
+    __HAL_RCC_GPIOE_CLK_ENABLE();
+    HAL_GPIO_Init(GPIOE, &gpio_init);
+}
+
+// This is the start of the sys memory region from the datasheet. It should be the
```
sfoster1: are these values right? should make sure and change the comment pr update and remove

sfoster1 (approved):
awesome work! only thing is maybe we should remove some of the suff thats super heatershaker specific but still good!

stm32-modules/thermocycler-refresh/firmware/CMakeLists.txt
fsinapi: As far as I can tell...  it's compiling/linking correctly. This'll be uncommented, I just wanted to take out the USB dependency until I've got the supporting code.

stm32-modules/thermocycler-refresh/firmware/system/system_hardware.c
fsinapi: Good catch! Updating with values for the G491

sfoster1 (approved):
Nice!

PR #215, opened 2021-11-10
feat(thermocycler-refresh): usb bringup

fsinapi:
### Overview
Enables basic USB communication on the thermocycler-refresh board. Board enumerates over USB and accepts a couple basic messages, all of which have been tested through the command line + the simulator:
- **M115** get serial
- **M996** set serial
- **dfu** enter bootloader

### Changes
- Added FreeRTOS tasks for host communications and a main system task
- Fixed a couple issues with system initialization, namely boosting the APB clocks to their max level and changing the system timebase from SysTick to TIM7
- Added basic simulator code to compile a working simulator instead of just a dummy program
- Added basic tests for the Comms Task and the System Task


stm32-modules/thermocycler-refresh/firmware/host_comms_task/usbd_desc.c
```diff
@@ -0,0 +1,277 @@
...
+
+/* Includes ------------------------------------------------------------------*/
+#include "usbd_core.h"
+#include "usbd_desc.h"
+#include "usbd_conf.h"
+
+/* Private typedef -----------------------------------------------------------*/
+/* Private define ------------------------------------------------------------*/
+#define USBD_VID                      0x0483
+#define USBD_PID                      0x4853
```
sfoster1: what's this pid from? might want to make it closer to the other thermocycler pid: https://github.com/Opentrons/opentrons-modules/blob/88e533eb52889ded1cc08a3ef66b8e55d0a6b808/arduino-modules/arduino-board-files/opentrons-samd-1.1.0/boards.txt#L27-L30 

Mostly need to make sure it's not the same as any of the other modules

stm32-modules/thermocycler-refresh/firmware/system/STM32G491VETx_FLASH.ld
```diff
@@ -147,7 +162,25 @@ SECTIONS
     _edata = .;        /* define a global symbol at data end */
   } >RAM AT> FLASH
 
+  _siccmram = LOADADDR(.ccmram);
```
sfoster1: are you planning on using this? i don't know if the startup asm here has support for that note below

sfoster1 (approved):
Couple little nitpicks but this looks really solid!

stm32-modules/thermocycler-refresh/firmware/system/STM32G491VETx_FLASH.ld
fsinapi: Nope... taking out this section, easy enough to add it back + update the assembly in if we want to use it later.

PR #216, opened 2021-11-11
refactor(heater-shaker, thermocycler-refresh): Consolidate common code

fsinapi:
### Overview
Pretty big restructure of the repository to consolidate the common code between modules and more closely match the structure of the OT3 repository.

### Changes
- Added a folder `stm32-modules/common/` that holds all of the code common to multiple modules. This is its own cmake module that exports libraries for use in the other modules. In order to provide cross-MCU support, the libraries imported by cross-compile builds take the form of `common-STM32F303` or `common-STM32F491`, with the option to add some more options if we use new MCU's in the future.
- Moved all headers to a common `stm32-modules/include/` directory to match the OT3 repository. Note that the cmake toolchain _only_ includes `include/common` and the module-specific `include` subdirectory for each module in order to provide safety against cross-contamination between module targets.
- Moved tests for common code to the common folder
- Added CI rules for the common code, and made the CI triggers for specific modules be dependent on the common directory

### Risk assessment
@pmoegenburg be aware that this will probably be annoying for you to pull into any active branches you have. Sorry about that.


stm32-modules/heater-shaker/firmware/CMakeLists.txt
```diff
@@ -73,30 +74,22 @@ add_executable(heater-shaker
...
              CXX_STANDARD_REQUIRED TRUE
              C_STANDARD 11
              C_STANDARD_REQUIRED TRUE)
 
-  target_include_directories(heater-shaker
-    PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/../include
-    PRIVATE ${MOTOR_DIR})
+target_include_directories(heater-shaker
+  PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/../../include/heater-shaker
+  PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/../../include/common
```
sfoster1: This should be included automatically as long as it's set as a public include dir for the common static lib

sfoster1 (approved):
This looks good overall; there's two things I think could be improved but I might be wrong and they're really nice-to-haves
1. I think with slightly more massaging of the way the common libraries work, by setting `include/common` to `PUBLIC` include directories, then adding them with `target_link_libraries` to the end executable will automatically include that directory. But if it works without it it still works and we can get to it later.
2. I'm a little unsure about whether the simulator cli parser and friends belongs in common or if it's really specific to heater-shaker... but again, if it works now it works now, and we'll figure out whether it really belongs as we implement more simulator stuff for thermocycler and we can move it at that time.

fsinapi:
@sfoster1 
> 1. I think with slightly more massaging of the way the common libraries work, by setting `include/common` to `PUBLIC` include directories, then adding them with `target_link_libraries` to the end executable will automatically include that directory. But if it works without it it still works and we can get to it later.

Yep, got rid of the extraneous `include` lines and everything still compiles.

> 2. I'm a little unsure about whether the simulator cli parser and friends belongs in common or if it's really specific to heater-shaker... but again, if it works now it works now, and we'll figure out whether it really belongs as we implement more simulator stuff for thermocycler and we can move it at that time.

You're correct, I misled myself by only checking the cpp files... the headers are pretty specific to each target so I moved everything back.

sfoster1 (approved):
Nice, let's merge it!

PR #217, opened 2021-11-15
feat(thermocycler-refresh): Create thermal tasks

fsinapi:
Adds two tasks to the TCR project. One task is for the thermal plate (aka the peltier subsystem), and the other for the lid heater (the heat pad). These are separated into two distinct tasks even though they'll share a bit of the same hardware control code because they are logically separate control loops.

Code is added for basic testing policy support for the new tasks, but no tests are added because there isn't anything (usefully) testable - will include test code with the thermistor code.

Also includes some cmake edits
- In #216 I messed up the variables pointing towards gdbinit
- Consolidated the _many_ calls to `find_package(clang)`

### DIFF
--- a/stm32-modules/CMakeLists.txt
+++ b/stm32-modules/CMakeLists.txt
@@ -3,9 +3,12 @@
 if (${CMAKE_CROSSCOMPILING})
     find_package(CrossGCC)
     find_package(OpenOCD)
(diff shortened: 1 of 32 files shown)


sfoster1 (approved):
Nice, good setup!

PR #218, opened 2021-11-16
feat(thermocycler-refresh): Add KS103J2G thermistor conversions

fsinapi:
### Summary
Added code to support the KS103J2G thermistor used on the thermocycler-refresh. Includes a CSV pulled from the vendor that contains nominal thermistor data.

### Review requests
Right now, _both_ thermistor tables are being included in projects (at least based on my examining through `objdump`), which is a waste of flash memory. Not sure if there's a good way to selectively include the tables without moving this code out of `common/`- if you have any good ideas for this then I'll happily add them into this PR


sfoster1 (changes requested):
Code looks good but yeah I'd be worried about having both tables in flash. Honestly the best thing to do might be to take the table generation itself out of top-level `common`. The thermistor stuff is already template-parametrized to accept which thermistor lookup table you want; leave it to the project to generate the table and instantiate it. We can keep the code shared by putting it in some other shared directory with cmake to generate the table and then have the project explicitly add the file rather than having it be part of the static library.

sfoster1 (approved):
Nice work!

PR #220, opened 2021-11-19
feat(thermocycler-refresh): set up thermistor readings

fsinapi:
Adds support to read thermistors periodically for the Thermal Plate and Lid Heater tasks. Includes:

- Added `thermal_hardware` which handles all HAL integration. Provides thread-safe I2C support using FreeRTOS mutex access and FreeRTOS notifications to signal the end of I2C transmissions.
- Added `ADS1115` class to handle higher level control of the ADC chips on the board. Provides thread-safe access to each ADC chip on the board through mutexes, and uses notifications to signal the end of a conversion (the IC sends a GPIO pulse at the end of a conversion).
- Added dedicated FreeRTOS threads that read the thermistors for each of the thermal tasks.
- Added Gcodes to request temperature-debug information for either the lid or the plate.
- Fixed an issue with the generation of the thermistor table - it was in Ohms instead of Kilohms

Test cases check that the temperature data is properly handled by the task and errors are propagated to host_comms as expected.

Was able to test by connecting to the thermistors on an actual thermocycler. Everything reads around 23.9 celsius in the office. Doing the conversion from reported ADC counts by hand, it looks correct.


stm32-modules/thermocycler-refresh/firmware/thermal/ads1115.cpp
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    return ReadVal(_last_result);
+}
+
+auto ADC::initialized() -> bool {
+    return (_adc_hardware[_id]._initialization_done == true);
+}
+
+auto ADC::get_lock() -> bool {
```
sfoster1, line 113: can we use [this](https://github.com/Opentrons/ot3-firmware/blob/main/include/common/core/freertos_synchronization.hpp) from ot3 firmware, probably just by copying for now

sfoster1 (approved):
Looks great! Only change I think we should make is to use the freertos c++ locking adaptors from ot3 firmware (linked in original comment)

stm32-modules/thermocycler-refresh/firmware/thermal/ads1115.cpp
fsinapi: Oh nice, didn't know this existed!

PR #221, opened 2021-11-20
fix(heater-shaker): motor unable to move error creation

pmoegenburg:


pmoegenburg:
mc_tasks buffers commands and is a tick behind motor_task commands, so this seems like a reasonable route. Also, mc_tasks doesn't currently check if the torque ramp process is making progress (to throw an error), and even if it did, it'd still be a tick behind and not throw an error response directly back up the function calls to the M3 gcode input

pmoegenburg:
Still need to update/fix tests

sfoster1 (changes requested):
This seems like it will work but the motor controller library in theory should be detecting this and sending us an error, I'd love to know why it's not.

The errors from the motor control library should be sent [here](https://github.com/Opentrons/opentrons-modules/blob/d8ec2f378bb62a3f802e305703269d8f971d0ec6/stm32-modules/heater-shaker/firmware/motor_task/freertos_motor_task.cpp#L93-L104) and come from the c side [here](https://github.com/Opentrons/opentrons-modules/blob/d8ec2f378bb62a3f802e305703269d8f971d0ec6/stm32-modules/heater-shaker/firmware/motor_task/mc_tasks.c#L601).

CMakePresets.json
```diff
@@ -42,7 +42,7 @@
         "MODULE_TYPE": "stm32",
         "CMAKE_FIND_APPBUNDLE": "NEVER",
         "CMAKE_TOOLCHAIN_FILE": "${sourceDir}/cmake/STM32CortexM4GCCCrossToolchain.cmake",
-        "CMAKE_BUILD_CONFIG": "MinSizeRel"
+        "CMAKE_BUILD_CONFIG": "RelWithDebInfo"
```
sfoster1: Is this ok? I know we previously had this as MinSizeRel (which requires `-Os`) so the program actually fit on the part.

stm32-modules/heater-shaker/firmware/motor_task/motor_hardware.c
```diff
@@ -690,8 +698,8 @@ void HAL_TIM_Base_MspInit(TIM_HandleTypeDef* htim_base)
     PA11     ------> TIM1_BKIN2
     */
     GPIO_InitStruct.Pin = M1_OCP_Pin;
-    GPIO_InitStruct.Mode = GPIO_MODE_AF_PP;
-    GPIO_InitStruct.Pull = GPIO_PULLUP;
+    GPIO_InitStruct.Mode = GPIO_MODE_AF_PP; //OD output from driver
+    GPIO_InitStruct.Pull = GPIO_PULLUP; //need? Have external. Break input may never trigger, but motor would stop in fault state. How big are driver internal resistors?
```
sfoster1: should get rid of comments with question marks in them or rephrase (and break lines)

sfoster1 (dismissed):
Probably need to change the default build mode back (you can ooverride it by doing `-DCMAKE_BUILD_CONFIG=RelWithDebInfo` when you cmake configure).

I'm also a bit confused about whether this is setting up a DAC reference and onboard comparator for current control, relying on the driver's fault output, doing a software-level RPM comparison, or some combination of the above to detect the errors. Maybe a comment block describing the strategy?

stm32-modules/include/heater-shaker/heater-shaker/motor_task.hpp
```diff
@@ -167,18 +169,37 @@ class MotorTask {
...
             state.status = State::RUNNING;
-            auto response = messages::AcknowledgePrevious{
-                .responding_to_id = msg.id, .with_error = error};
-            if (msg.from_system) {
-                static_cast<void>(
-                    task_registry->system->get_message_queue().try_send(
-                        messages::SystemMessage(response)));
+            policy.delay_ticks(MOTOR_START_WAIT_TICKS);
+            static_cast<void>(
+                get_message_queue().try_send(messages::CheckMotorStartMessage{
```
fsinapi: I'm not clear on why this is an extra message. Since the handling for this message includes the necessary delay, couldn't the error handling be done in the same function?

fsinapi (commented):
Looks good to me, one minor comment. I'll let @sfoster1 make the approval, I don't have as much insight on the development of this error checking.

sfoster1 (approved):
Looks good to me, really like the tests! We'll need to make sure that the higher level integrations are ok with a second of delay in a successful set rpm, but we don't really have a choice about it. Awesome!

PR #223, opened 2021-11-22
chore(thermocycler-refresh, heater-shaker): update CMakePresets to version 2

fsinapi:
Updated CMakePresets.json to use the cmake 3.20 preset file version, and added build presets on top of the configure presets. This also allows nice integration with vscode (although the debug targets don't really work from vscode's terminal...)

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -117,9 +117,20 @@
         },
         {
             "name": "format-firmware",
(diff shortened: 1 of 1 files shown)


sfoster1 (approved):
cool. you can also set `jobs` in those build presets fyi

PR #224, opened 2021-11-22
feat: Add host name resolving to heater-shaker and thermocycler

DerekMaggio:
Add ability to use a hostname when using `--socket` cli option

`socket://myhostname:8000` and `socket://172.0.21.2:8000` will now both be able to be parsed

### DIFF
--- a/stm32-modules/heater-shaker/simulator/socket_sim_driver.cpp
+++ b/stm32-modules/heater-shaker/simulator/socket_sim_driver.cpp
@@ -17,9 +17,16 @@ std::unique_ptr<boost::asio::ip::tcp::socket> connect_to_socket(
     std::string host, int port) {
     boost::asio::io_service io_context;
     boost::asio::ip::tcp::resolver resolver(io_context);
-    auto endpoints = resolver.resolve(host, std::to_string(port));
-    std::string parsed_host =
-        endpoints.begin()->endpoint().address().to_string();
+    std::string parsed_host;
+    try {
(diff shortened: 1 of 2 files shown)


stm32-modules/heater-shaker/simulator/socket_sim_driver.cpp
```diff
@@ -16,10 +16,14 @@ const std::string SOCKET_DRIVER_NAME = "Socket";
 std::unique_ptr<boost::asio::ip::tcp::socket> connect_to_socket(
     std::string host, int port) {
     boost::asio::io_service io_context;
+    boost::asio::ip::tcp::resolver resolver(io_context);
+    auto endpoints = resolver.resolve(host, std::to_string(port));
+    std::string parsed_host =
```
amitlissack: I think you should check `endpoints` and print a nice error if it's empty.

amitlissack (changes requested):

stm32-modules/heater-shaker/simulator/socket_sim_driver.cpp
DerekMaggio: Is https://github.com/Opentrons/opentrons-modules/pull/224/commits/903c0d426b9bb5f0168f5c1dd335fd0fbbeeabf5 better

fsinapi (approved):
Nice!

amitlissack (approved):

PR #225, opened 2021-11-23
chore(thermocycler-refresh): Remove HAL from all C++ code

fsinapi:
All my homies hate `#pragma GCC diagnostic ignored "-Wvolatile"`

Mostly small changes, with the exception of a new `usb_hardware.c/h` file that acts as a glue layer over the STM32 USB CDC libraries. Tested on my board that USB still works fine.


sfoster1 (approved):
nice, love to see it

PR #226, opened 2021-11-30
feat(thermocycler-refresh): Add peltier driver code

fsinapi:
### Summary
This pull request enables basic control of the peltiers on the system, as well as fixing some issues with mapping to the thermistors on the system.

- A new gcode (M104.D) is added for debug control over the peltiers. The command provides independent control of each peltier (L/R/C/A for left/right/center/all) in heating or cooling (H/C) directions and a variable pulse width (0.0 to 1.0).
- Timer 1 channels 1-3 are configured for PWM output to control the peltiers. The timer is set to operate at 25kHz and counts from 0 to 100. _A higher count can be configured but either the frequency will have to change a bit or the count won't be a power of ten_
- The peltiers are driven with a _single_ PWM channel and then a simple GPIO output. For cooling, the GPIO is low and the PWM is calculated normally. To heat the peltier, the GPIO is set high and then the PWM polarity is inverted.
- Unit tests exercise parsing of the new gcode and behavior of the Thermal Plate Task when sent SetPeltierDebug messages.
- Added some python scripts that were used in debugging

Code has been tested with the original thermocycler system. M104.D commands result in increasing/decreasing temperature on different thermistor regions as expected.


stm32-modules/thermocycler-refresh/firmware/thermal/thermal_peltier_hardware.c
```diff
@@ -0,0 +1,324 @@
...
+}
+
+bool thermal_peltier_get_enable(void) {
+    return _peltiers.enabled;
+}
+
+void thermal_peltier_set_power(const PeltierID id, 
+                               double power,
+                               const PeltierDirection direction) {
+    if(!(id < PELTIER_NUMBER)) {
```
sfoster1, line 291: should probably return some sort of error here, even if it's just a bool that we can put in to a general "something broke in the peltiers" message

sfoster1 (changes requested):
Looks good to me but let's hit that last one place where we should have an error

stm32-modules/thermocycler-refresh/firmware/thermal/thermal_peltier_hardware.c
fsinapi: Added a bool return here + an error code to send over host comms 👍

sfoster1 (approved):
Nice!

PR #227, opened 2021-12-01
feat(thermocycler-refresh): Add fan driver

fsinapi:
Adds PWM driver support for the system fans. They are incredibly loud.

- M106 sets the power of the fans as a percentage from 0 to 1.0
- Fan PWM uses the same timing configuration as the peltiers (25kHz, counts from 0 to 99)
- Unit tests included for gcode parsing and message handling in the Thermal Plate task

Tested on a unit by hooking up the fans and setting PWM through the gcode.


sfoster1 (approved):
lgtm!

PR #228, opened 2021-12-01
feat(thermocycler-refresh): add heat pad driver

fsinapi:
Adds a driver for the thermal heating pad in the lid. 

- Controls heater with a PWM line from TIM15 and an enable pin. 
- Includes a debug gcode for setting the PWM of the heater directly (M140.D)

Tested by sending M140.D with a few different powers (0.2, 0.5, 0.75) and watching the temperature increase at different rates. Also tested that sending power of 0.0 makes the temperature stop going up.


sfoster1 (approved):
Noice

PR #229, opened 2021-12-07
feat(thermocycler-refresh): heat pad closed loop control

fsinapi:
### Summary
Adds a few different gcodes:

- GetPlateTemp and GetLidTemp for non-debugging temperature information (average value of the plate, includes target temp if defined)
- SetLidTemperature to set the target temperature of the lid
- DeactivateLidHeating to turn off the lid heater
- SetPIDConstants to reconfigure the PID constants for the heater

Also adds support for the closed-loop heater control. Tested controlling the heater up to 90C, it moves up nice and steady and then settles. The PID constants are just pulled from the heater-shaker so we will still have to tune the loop, but that can wait for a full hardware system.


sfoster1 (approved):
Looks good!

PR #230, opened 2021-12-09
feat(thermocycler-refresh): Add Peltier closed-loop control

fsinapi:
### Summary
Adds closed-loop peltier control. PID constants are just default from the heater shaker (definitely not correct for this system BUT tuning will happen soon). Closed loop fan control is excluded from this PR.

- Adds two gcodes, one for setting the plate target temp and one for disabling the plate.
- Adds closed loop control for the peltiers. Each peltier gets independent control based on the average temp of its pair of thermistors. Error checking is TODO
- Adds response handling in the Plate task for reconfiguring PID constants. Includes a set of constants for the fan that isn't used yet
- Added a python script that does some cycling between 50C and 15C. Screenshot attached just for fun.
- Unit tests for gcode handling and some behavior in the heater/plate tasks

image

### DIFF
--- a/stm32-modules/include/thermocycler-refresh/thermocycler-refresh/gcodes.hpp
+++ b/stm32-modules/include/thermocycler-refresh/thermocycler-refresh/gcodes.hpp
@@ -679,6 +679,96 @@ struct DeactivateLidHeating {
     }
 };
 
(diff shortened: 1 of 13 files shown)


sfoster1 (approved):
noice

PR #231, opened 2021-12-14
feat(thermocycler-refresh): add automatic fan control

fsinapi:
### Overview
Adds automatic fan control, per the control regimes outlined in the TC-R page on Confluence. Tested on a unit by ramping between a few different temperatures with the `test_plate_pid.py` script and changing the setpoints. 

### Changelog
- Moved logic for setting Peltier & Fan outputs to a PlateControl module to reduce the logic encapsulated in the Thermal Plate task
- Added a gcode M107 that resets the fan to automatic control if it has been set to manual mode
- Changed structuring of public structures in `thermal_general` such that the peltiers/fans maintain references to thermistor handles instead of just storing the current temperature
- Disabled asynchronous thermistor error messages, which were the only messages that would be sent asynchronously on the system... functionality is still there but it's commented out with a `#define` 

### Review Requests
- Right now, the structs in `thermal_general` are only used as basic shells for holding information, while the actual control of everything takes place in the thermal plate task _or_ the PlateControl class. Is that a bit of an antipattern? Should the actual setting of outputs move to PlateControl instead of being in the thermal plate task?

### DIFF
--- a/stm32-modules/include/thermocycler-refresh/systemwide.h
+++ b/stm32-modules/include/thermocycler-refresh/systemwide.h
@@ -7,9 +7,6 @@
 /* size of array for setting serial number */
 #define SYSTEM_WIDE_SERIAL_NUMBER_LENGTH 24
 
-/* Uncomment this to include aynchronous error messages */
-//#define SYSTEM_ALLOW_ASYNC_ERRORS
-
 typedef enum PeltierID {
     PELTIER_RIGHT,
     PELTIER_CENTER,
--- a/stm32-modules/include/thermocycler-refresh/systemwide.h
+++ b/stm32-modules/include/thermocycler-refresh/systemwide.h
@@ -7,6 +7,9 @@
 /* size of array for setting serial number */
 #define SYSTEM_WIDE_SERIAL_NUMBER_LENGTH 24
 
+/* Uncomment this to include aynchronous error messages */
+//#define SYSTEM_ALLOW_ASYNC_ERRORS
+
 typedef enum PeltierID {
     PELTIER_RIGHT,
     PELTIER_CENTER,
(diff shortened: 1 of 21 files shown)


stm32-modules/include/thermocycler-refresh/thermocycler-refresh/thermal_plate_task.hpp
```diff
@@ -598,6 +621,47 @@ class ThermalPlateTask {
...
+     * and new temperatures have been stored in the thermistor handles.
+     * @param[in] policy The thermal plate policy
+     * @return True if outputs are updated fine, false if any error occurs
+     */
+    template <ThermalPlateExecutionPolicy Policy>
+    auto update_control(Policy& policy) -> bool {
+        policy.set_enabled(true);
+        auto values = _plate_control.update_control();
+        auto ret = values.has_value();
+        if (ret) {
```
sfoster1, line 636: I'm worried about this pattern - if we have some error, don't we want to turn everything off rather than not update the others? will the others get stuck on?

stm32-modules/include/thermocycler-refresh/systemwide.h
```diff
@@ -7,6 +7,9 @@
 /* size of array for setting serial number */
 #define SYSTEM_WIDE_SERIAL_NUMBER_LENGTH 24
 
+/* Uncomment this to include aynchronous error messages */
+//#define SYSTEM_ALLOW_ASYNC_ERRORS
```
sfoster1: This might be better off as a build system define done by cmake rather than a change in source

sfoster1 (approved):
Looks good to me, though maybe take a look at the error handling in the main peltier control function

stm32-modules/include/thermocycler-refresh/thermocycler-refresh/thermal_plate_task.hpp
fsinapi: Good point, this doesn't make sense. Added error handling to disable the peltiers if something goes wrong.

PR #232, opened 2021-12-20
feat(thermocycler-refresh): Add TMC2130 SPI interface

fsinapi:
### Summary

Adds in a class to control the TMC2130 motor IC. Primary interface to the class is through bit-packed structures. Right now this _just_ adds the SPI register control, there isn't any support for moving the motors.

- Added the bit_utils header from OT3 as a nice utility
- Tested on a board by butchering the `freertos_system_task` and pinging registers. Confirmed that it can set the gconfig register and readback from the gstat register as expected.
- Unit tests cover setting/getting all of the registers that we are expecting to use
- Should be fine to merge without the other motor setup code, really shouldn't conflict with any of that work. 

### Review Requests

Sanity check that the system for serializing/deserializing the bitpacked structures won't have any issues... the data seemed fine in a debugger but it's on the riskier side with the pointer reinterpret casts and such

### DIFF
--- a/stm32-modules/include/thermocycler-refresh/thermocycler-refresh/tmc2130.hpp
+++ b/stm32-modules/include/thermocycler-refresh/thermocycler-refresh/tmc2130.hpp
@@ -42,9 +42,17 @@ class TMC2130 {
         return false;
     }
     /**
-     * @brief Get the current GCONF register status
+     * @brief Get the current GCONF register status. This register can
+     * be read, so this function gets it from the actual device.
      */
-    auto get_gconf() -> GConfig { return _registers.gconfig; }
+    template <TMC2130Policy Policy>
+    auto get_gconf(Policy& policy) -> GConfig { 
+        auto ret = read_register<GConfig>(policy);
+        if(ret.has_value()) {
+            _registers.gconfig = ret.value();
+        }
+        return _registers.gconfig;
+    }
 
     /**
      * @brief Update IHOLDIRUN register
--- a/stm32-modules/include/thermocycler-refresh/thermocycler-refresh/tmc2130.hpp
+++ b/stm32-modules/include/thermocycler-refresh/thermocycler-refresh/tmc2130.hpp
@@ -4,370 +4,201 @@
  */
 #pragma once
 
-#include <bitset>
 #include <concepts>
 #include <cstdint>
 #include <optional>
 
 #include "systemwide.h"
+#include "thermocycler-refresh/tmc2130_registers.hpp"
 
 namespace tmc2130 {
 
-// Register mapping
-
-enum class Registers {
-    GCONF = 0x00,
-    GSTAT = 0x01,
-    IOIN = 0x04,
-    IHOLD_IRUN = 0x10,
-    TPOWERDOWN = 0x11,
-    TSTEP = 0x12,
-    TPWMTHRS = 0x13,
-    TCOOLTHRS = 0x14,
-    THIGH = 0x15,
-    XDIRECT = 0x2D,
-    VDCMIN = 0x33,
-    CHOPCONF = 0x6C,
-    COOLCONF = 0x6D,
-    DCCTRL = 0x6E,
-    DRVSTATUS = 0x6F,
-    PWMCONF = 0x70,
-    ENCM_CTRL = 0x72,
-    MSLUT_0 = 0x60,
-    MSLUT_1 = 0x61,
-    MSLUT_2 = 0x62,
-    MSLUT_3 = 0x63,
-    MSLUT_4 = 0x64,
-    MSLUT_5 = 0x65,
-    MSLUT_6 = 0x66,
-    MSLUT_7 = 0x67,
-    MSLUTSEL = 0x68,
-    MSLUTSTART = 0x69,
-    MSCNT = 0x6A,
-    MSCURACT = 0x6B,
-    PWM_SCALE = 0x71,
-    LOST_STEPS = 0x73,
-};
-
-struct GConfig {
-    static constexpr Registers address = Registers::GCONF;
-    static constexpr bool readable = true;
-    static constexpr bool writable = true;
-    static constexpr uint64_t bitlen = 17;
-
-    uint8_t i_scale_analog : 1 = 0;
-    uint8_t internal_rsense : 1 = 0;
-    uint8_t en_pwm_mode : 1 = 0;
-    uint8_t enc_commutation : 1 = 0;  // MUST be 0
-    uint8_t shaft : 1 = 0;
-    uint8_t diag0_error : 1 = 0;
-    uint8_t diag0_otpw : 1 = 0;
-    uint8_t diag0_stall : 1 = 0;
-    uint8_t diag1_stall : 1 = 0;
-    uint8_t diag1_index : 1 = 0;
-    uint8_t diag1_onstate : 1 = 0;
-    uint8_t diag1_steps_skipped : 1 = 0;
-    uint8_t diag0_int_pushpull : 1 = 0;
-    uint8_t diag1_pushpull : 1 = 0;
-    uint8_t small_hysteresis : 1 = 0;
-    uint8_t stop_enable : 1 = 0;
-    uint8_t direct_mode : 1 = 0;
-    uint8_t test_mode : 1 = 0;  // MUST be 0
-} __attribute__((packed));
-
-struct GStatus {
-    static constexpr Registers address = Registers::GSTAT;
-    static constexpr bool readable = true;
-    static constexpr bool writable = false;
-    static constexpr uint64_t bitlen = 3;
-
-    uint8_t reset : 1 = 0;
-    uint8_t driver_error : 1 = 0;
-    uint8_t undervoltage_error : 1 = 0;
-} __attribute__((packed));
-
-struct CurrentControl {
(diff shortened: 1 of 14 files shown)


stm32-modules/include/thermocycler-refresh/thermocycler-refresh/tmc2130.hpp
```diff
@@ -0,0 +1,316 @@
...
+    /**
+     * @brief Update IHOLDIRUN register
+     * @param reg New configuration register to set
+     * @param policy Instance of abstraction policy to use
+     * @return True if new register was set succesfully, false otherwise
+     */
+    template <TMC2130Policy Policy>
+    auto set_current_control(CurrentControl reg, Policy& policy) -> bool {
+        // Assert that bits that MUST be 0 are actually 0
+        if ((reg.bit_padding_1 != 0) || (reg.bit_padding_2 != 0)) {
```
sfoster1: Seems like it should be ok to just set these to 0 here

stm32-modules/include/thermocycler-refresh/thermocycler-refresh/tmc2130.hpp
```diff
@@ -0,0 +1,316 @@
...
+    /**
+     * @brief Update GCONF register
+     * @param reg New configuration register to set
+     * @param policy Instance of abstraction policy to use
+     * @return True if new register was set succesfully, false otherwise
+     */
+    template <TMC2130Policy Policy>
+    auto set_gconf(GConfig reg, Policy& policy) -> bool {
+        // Assert that bits that MUST be 0 are actually 0
+        if ((reg.enc_commutation != 0) || (reg.test_mode != 0)) {
```
sfoster1: just set to 0?

stm32-modules/include/thermocycler-refresh/thermocycler-refresh/tmc2130.hpp
```diff
@@ -0,0 +1,316 @@
...
+        if (set_register(policy, reg)) {
+            _registers.ihold_irun = reg;
+            return true;
+        }
+        return false;
+    }
+    /**
+     * @brief Get the current IHOLDIRUN register status
+     */
+    [[nodiscard]] auto get_current_control() const -> CurrentControl {
```
sfoster1: It's a little weird that the getters are inconsistent between retrieving values from the remote and returning the cached version. Especially since we have this whole other class for register map, IMO everything that's a getter in this class should be incurring a bus read and getting the remote status - even if it should be the same since the last write - and if you want cached values, you can check the map object.

stm32-modules/include/thermocycler-refresh/thermocycler-refresh/tmc2130_registers.hpp
```diff
@@ -0,0 +1,380 @@
...
+concept TMC2130Register = requires(Reg& r, uint64_t value) {
+    // Struct has a valid register address
+    std::same_as<decltype(Reg::address), Registers&>;
+    // Struct has a bool saying whether it can be read
+    std::same_as<decltype(Reg::readable), bool>;
+    // Struct has a bool saying whether it can be written
+    std::same_as<decltype(Reg::writable), bool>;
+    // Struct has an integer with the total number of bits in a register.
+    // This is used to mask the 64-bit value before writing to the IC.
+    std::integral<decltype(Reg::bitlen)>;
```
sfoster1: given what `bitlen` is used for, does it make more sense for it to be `value_mask` or something and either written as a mask directly or `consteval`ed from a bitshift?

stm32-modules/include/thermocycler-refresh/thermocycler-refresh/tmc2130.hpp
```diff
@@ -0,0 +1,316 @@
...
+     * @tparam Policy Abstraction class for actual writing
+     * @param[in] policy Instance of the abstraction policy to use
+     * @param[in] reg The register to write
+     * @return True if the register could be written, false otherwise.
+     * Attempts to write to an unwirteable register will throw a static
+     * assertion.
+     */
+    template <TMC2130Register Reg, TMC2130Policy Policy>
+    auto set_register(Policy& policy, Reg reg) -> bool {
+        static_assert(Reg::writable);
```
sfoster1: this could also be 
```
template <TMC2130Register Reg, TMC2130Policy Policy>
requires (const Reg& reg) { {reg::writable }; }
auto set_register(...
```
i believe. Doing it this way might give better error messages, and would be consistent with our use of requirements elsewhere.

I think you could also use a named-requirement or concept for this, possibly by making a `WritableTMC2130Register` and `ReadableTMC2130Register` that looks like the requirement above `&& TMC2130Register`

stm32-modules/include/thermocycler-refresh/thermocycler-refresh/tmc2130_registers.hpp
```diff
@@ -0,0 +1,380 @@
...
+#pragma once
+
+#include <concepts>
+#include <cstdint>
+
+namespace tmc2130 {
+
+// Register mapping
+
+enum class Registers {
```
sfoster1:
```suggestion
enum class Registers: uint8_t {
```

stm32-modules/thermocycler-refresh/firmware/motor_task/motor_spi_hardware.c
```diff
@@ -0,0 +1,194 @@
...
+    if(!_spi.initialized) { return false; }
+    _spi.enabled = enable;
+    HAL_GPIO_WritePin(MOTOR_SPI_ENABLE_PORT, MOTOR_SPI_ENABLE_PIN, 
+        (enable) ? GPIO_PIN_RESET : GPIO_PIN_SET);
+    return true;
+}
+
+void SPI2_IRQHandler(void)
+{
+  /* USER CODE BEGIN SPI2_IRQn 0 */
```
sfoster1: let's get rid of these

stm32-modules/thermocycler-refresh/firmware/system/stm32g4xx_hal_msp.c
```diff
@@ -255,6 +255,92 @@ void HAL_TIM_Base_MspDeInit(TIM_HandleTypeDef* htim_base)
...
+    */
+    GPIO_InitStruct.Pin = GPIO_PIN_13|GPIO_PIN_14|GPIO_PIN_15;
+    GPIO_InitStruct.Mode = GPIO_MODE_AF_PP;
+    GPIO_InitStruct.Pull = GPIO_NOPULL;
+    GPIO_InitStruct.Speed = GPIO_SPEED_FREQ_LOW;
+    GPIO_InitStruct.Alternate = GPIO_AF5_SPI2;
+    HAL_GPIO_Init(GPIOB, &GPIO_InitStruct);
+
+    // Uncomment to intialize PD15 as a hardware NSS signal
+    //GPIO_InitStruct.Pin = GPIO_PIN_15;
```
sfoster1: let's get rid of the commented out code

sfoster1 (commented):
Keeping this as a comment because I don't want to block anything rest of week, but I think there's some small stuff we should change.

For the reinterprets, the only thing I'd be worried about there is byte order, and bitutils should help.

stm32-modules/include/thermocycler-refresh/thermocycler-refresh/tmc2130.hpp
fsinapi: Issue is that a lot of these registers are write-only. GCONF is RW and CHOPCONF is RW (which I just realized), but everything else in the config map is Write-only

sfoster1: then IMO we shouldn't have reads for those - just the register map read

stm32-modules/include/thermocycler-refresh/thermocycler-refresh/tmc2130.hpp
fsinapi: Hmm, tried this and it doesn't work as expected - you can read from a nonreadable register and write to a nonwritable register. I think this just checks that the variable exists - looking through the `concepts` library I'm not sure what the right syntax would be to check that the value of the variable is `true`.

sfoster1: There's a way to do this I think maybe but the better way is something like this, from the "type requirements" section of https://en.cppreference.com/w/cpp/language/constraints (no section anchors, bad cppreference). These can also be `using`-declarations. You can also do this with straight up old style SFINAE using `enable_if` but that's getting to be a little much I think

```c++
template <typename _RegType>
concept WritableRegister = requires {
    typename _RegType::writable; 
}

template <typename _RegType>
concept ReadableRegister = requires {
    typename _RegType::readable; 
}

struct SomeRWRegister {
   static constexpr bool writable = true;
   static constexpr bool readable = true;
};

struct SomeRORegister {
   static constexpr bool readable = true;
   // do not define writable
};

struct SomeWORegister {
  static constexpr bool writable = true;
  // do not define readable
};
```

fsinapi: Ahh makes sense, I wasn't getting rid of the `false` flags

sfoster1 (approved):
This is great and I'm totally gonna steal this for ot3

PR #233, opened 2021-12-21
feat(thermocycler-refresh): Add board revision checking

fsinapi:
### Summary

Adds functionality to read the revision number of the PCB. This will allow firmware to account for any changes in pin allocation/configuration for future revisions.

- Added a `hal_util` header that has (for now) a structure type to better encapsulate GPIO definitions for the STM32 HAL. This can be expanded in the future.
- Added an interface class that allows either getting the cached board rev number or reading it new. The `read` function is mostly for testing, on the firmware it's much more useful to just use the static cached version.

Tested on board, it correctly reads board version 1. Can't test with a rev 2 until those PCBs come in.


sfoster1 (approved):
Nice

PR #234, opened 2022-01-07
feat(thermocycler-refresh): add LED support

fsinapi:
### Summary
Adds support for the LED strip used as the UI on the thermocycler. Currently just turns all of the LEDs to white on startup.

Uses a DMA stream connected to TIM17's capture compare event to send all of the data.

Tested on a board by changing a couple of pixel values manually and verifying that the LEDs light up as expected.


sfoster1 (approved):
LGTM if it works on the hardware

PR #236, opened 2022-01-14
feat(thermocycler-refresh): LED actions support

fsinapi:
### Summary

Adds in UI action support to the system task.

- The periodic updating of the LED's is driven by a FreeRTOS periodic timer, lifted and somewhat modified from OT3
- XT1511 driver is modified to support the half-baud protocol options; this uses 400kHz instead of 800kHz and has slightly different control values. The code is switched to use this protocol because the peripheral clocks are at 170MHz and the math to get a timer going at 800kHz simply doesn't work out - the LEDs work fine most of the time, but repeated writes show some small glitches. Fortunately, 400kHz works fine without having to redo the clock muxing.
- Added a color mapping driver to standardize color values (lifted from TC 1.0)

Tested on hardware by changing System Task initialization code to each of the different actions + confirming they show the correct actions. The pulsing/blinking/wiping behavior is written to emulate the original thermocycler as close as possible.

Does not have automatic UI state setting. This will be added in the future to be controlled by the Thermal Plate Task.

### DIFF
--- a/stm32-modules/thermocycler-refresh/src/colors.cpp
+++ b/stm32-modules/thermocycler-refresh/src/colors.cpp
@@ -1,0 +1,47 @@
+/**
+ * @file colors.cpp
+ * @brief Implements fetching color structures from codes
+ */
+
+#include "thermocycler-refresh/colors.hpp"
+
+using namespace colors;
+
+static constexpr uint8_t FULL = 0xFF;
+static constexpr uint8_t HIGH = 0xEE;
+static constexpr uint8_t MED = 0x50;
+
+// Public function implementation
+auto colors::get_color(Colors color, double brightness) -> xt1511::XT1511 {
+    if (brightness > 1.0F) {
+        brightness = 1.0F;
+    }
+    xt1511::XT1511 led{};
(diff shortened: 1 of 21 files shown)


stm32-modules/thermocycler-refresh/src/colors.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include "thermocycler-refresh/colors.hpp"
+
+using namespace colors;
+
+static constexpr uint8_t FULL = 0xFF;
+static constexpr uint8_t HIGH = 0xEE;
+static constexpr uint8_t MED = 0x50;
+
+// Public function implementation
+auto colors::get_color(Colors color, double brightness) -> xt1511::XT1511 {
```
sfoster1, line 23: I think we're going to end up having a fixed color palette that gets fine tuned in implementation to look correct; I think we should get ahead of the game a bit here by just specifying values for the four color channels. We should also establish and document an explicit color depth we support.

sfoster1 (approved):
Looks good to me although I'd anticipate a lot of finetuning with colors.

PR #238, opened 2022-01-21
feat(thermocycler-refresh): add motor task with solenoid and lid hinge stepper support

fsinapi:
### Summary

Adds in a `motor_task` that is responsible for anything related to the lid open/close subsystem. Right now this supports activating/deactivating the solenoid for the lid, as well as moving the lid hinge stepper motor with open-loop relative movements.

- Debug gcodes for moving the stepper motor and for actuating the solenoid are included
- The hinge stepper is controlled with TIM2 set to Output Compare mode with toggling output. Might revisit this when integrating control for the seal stepper motor.

Tested on a V1 thermocycler using the gcodes through included python utility commands.


stm32-modules/include/thermocycler-refresh/firmware/motor_hardware.h
```diff
@@ -0,0 +1,35 @@
...
+
+#include <stdbool.h>
+#include <stdint.h>
+
+// ----------------------------------------------------------------------------
+// Type definitions
+
+// This structure is used to define callbacks out of motor interrupts
+typedef struct {
+    void (*lid_stepper_complete)();
```
sfoster1: [nit] due to an extreme case of Hating Function Pointer Syntax, I hate function pointer syntax. Can we typedef this?

stm32-modules/include/thermocycler-refresh/thermocycler-refresh/motor_task.hpp
```diff
@@ -0,0 +1,178 @@
...
+        }
+    }
+
+    template <typename Policy>
+    auto visit_message(const messages::LidStepperComplete& msg, Policy& policy)
+        -> void {
+        static_cast<void>(msg);
+        if (lid_stepper_state.status == LidStepperState::MOVING) {
+            lid_stepper_state.status = LidStepperState::IDLE;
+            policy.lid_stepper_set_dac(0);
```
sfoster1: do we want to bind this in here? you're the expert but i'm a bit surprised we want to turn off the motor when done moving, as opposed to maintaining some amount of holding current. I guess in the fully-open or fully-closed case this is right, but maybe it's better if that's a separate message if we end up needing to either change that or string moves together.

stm32-modules/thermocycler-refresh/firmware/motor_task/motor_hardware.c
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    HAL_TIM_OC_Start_IT(&_motor_hardware.lid_timer, LID_STEPPER_STEP_Channel);
+}
+
+void motor_hardware_lid_stepper_stop() {
+    HAL_TIM_OC_Stop_IT(&_motor_hardware.lid_timer, LID_STEPPER_STEP_Channel);
+}
+
+void HAL_TIM_OC_DelayElapsedCallback(TIM_HandleTypeDef *htim)
```
sfoster1, line 110: you know, if you're feeling your oats you could do a timer cascade here. Have a second counter whose input trigger is TIM2 channel reload, with a compare value of the number of steps, and have _that_ timer's interrupt value stop the first. Or maybe you can make that part automatic too.

sfoster1 (approved):
Looks good, though maybe we should change the turn-motor-off-when-done logic and there's a couple other inline comments - but this would be fine to merge.

stm32-modules/include/thermocycler-refresh/thermocycler-refresh/motor_task.hpp
fsinapi: This function (and possibly the whole handling methodology) is going to have to change to have some state machines for opening/closing the lid, this is really just here as a stopgap since we can only move the lid for debug purposes. I _will_ add a constant for the hold current though.

PR #239, opened 2022-01-24
fix(heater-shaker): fix NTC disconnect error handling and recovery

pmoegenburg:
An NTC disconnect does not max out the ADC, so a pad reading error does not occur, which is what is used to attempt to reset the latch when the NTC is reconnected. An NTC disconnect does create a temp reading that pretty quickly stabilizes around -20C. This fix uses this condition, along with HEAT_POWER_GOOD being low, to create the pad reading error. I will update the failing tests if you guys approve this method of fixing.


sfoster1 (changes requested):
Damn. Looks like it will work, agree with the general approach. Two key things though:

- Need tests
- IMO this wants to be a threshold that is in ADC values, not in temperature, since this isn't a valid temperature. But this approach does also work.

stm32-modules/include/common/core/thermistor_conversion.hpp
```diff
@@ -45,6 +45,7 @@ struct Conversion {
     /** First is After, second is Before (after - 1) */
     using TableEntryPair = std::pair<TableEntry, TableEntry>;
     using TableResult = std::variant<TableEntryPair, TableError>;
+    static constexpr double HEATER_PAD_DISCONNECT_SAFETY_LIMIT_ADC = 3642.3; //0C equivalent
```
fsinapi: This should be a configurable parameter of the thermistor conversion class instead of being hard coded. The same driver is used on the thermocycler with a different ADC depth and a different resistor ladder.

fsinapi (changes requested):
Overall approach makes sense. Need to be careful about changes to common drivers, and there should be some tests added for both the Heater Task as well as any additions to the `thermistor conversion` driver.

fsinapi (approved):
Looks good to me!

sfoster1 (approved):
Nice!

PR #241, opened 2022-01-29
fix(heater-shaker): overtemp latch handling

pmoegenburg:
Currently, the hardware and software overtemp thresholds are very close to each other and one may fire before the other. This fix designates an overtemp from either source to the same error. After an overtemp error, if the temp is below the software threshold and above the hardware threshold when a latch reset is attempted, the HEAT_POWER_GOOD pin will not immediately be set, leading to some troublesome errors and incorrect state settings. This fix only resets the latch after an overtemp after the temp has dropped 5*C below the software threshold, ensuring HEAT_POWER_GOOD will be set after the latch reset. The hardware threshold is being increased to 105*C (with the software threshold remaining at 100*C) to further fix this issue.


stm32-modules/include/heater-shaker/heater-shaker/heater_task.hpp
```diff
@@ -93,6 +93,8 @@ class HeaterTask {
     static constexpr uint8_t ADC_BIT_DEPTH = 12;
     static constexpr uint16_t HEATER_PAD_NTC_DISCONNECT_THRESHOLD_ADC =
         3642;  // 0C equivalent
+    static constexpr uint8_t HEATER_PAD_HARDWARE_OVERTEMP_OFFSET_C = 1;
```
fsinapi: Since `overtemp_limit_c` is a double, these should probably also be doubles.

stm32-modules/include/heater-shaker/heater-shaker/heater_task.hpp
```diff
@@ -451,6 +453,14 @@ class HeaterTask {
     auto visit_conversion(double value, TemperatureSensor& sensor) -> void {
         if (value > sensor.overtemp_limit_c) {
             sensor.error = sensor.overtemp_error;
+        } else if ((value > (sensor.overtemp_limit_c -
```
fsinapi: A comment in this function documenting the error logic would be helpful.

fsinapi (approved):
Looks good, just a couple of small nits! I do think that we should add a test case for the error recovery, but the code itself looks good.

stm32-modules/include/heater-shaker/heater-shaker/heater_task.hpp
sfoster1: agree, and it might even be a good idea to factor the conditionals into their own `static inline` functions. That way, you can call the functions in the if statement, but inside the functions you're not constraints to having a single expression, so you could do like

```c++
static inline bool is_hardware_overtemp(double value, TemperatureSensor& sensor, uint8_t bitmap) {
    if (value > (sensor.overtemp_limit_c - HEATER_PAD_HARDWARE_OVERTEMP_OFFSET_C) ) {
       if ((bitmap & State::POWER_GOOD_ERROR) != 0) {
          return true;
       } 
    }
    return false;
}
```

sfoster1: i guess it depends whether you find that more readable or not; I thought it was pretty strictly an improvement until I typed it out and now I think it's kind of up to preference.

sfoster1: I guess if you did the function it would be more reasonable to combine both the new else ifs here if you wanted.

sfoster1 (approved):
Agree on the test case. I also think there could be a readability win from making the new if cases their own function but it's kind of up to you.

PR #243, opened 2022-01-31
feat(heater-shaker, thermocycler-refresh): Get simulator serial number from environment variable

fsinapi:
The `system_task` for each module now checks on startup for an environment variable `SERIAL_NUMBER` and sets the serial number for the system appropriately if it is defined. Tested on my local terminal by using gcode `M115`

### DIFF
--- a/stm32-modules/heater-shaker/simulator/system_thread.cpp
+++ b/stm32-modules/heater-shaker/simulator/system_thread.cpp
@@ -11,6 +11,7 @@
 
 #include "heater-shaker/errors.hpp"
 #include "heater-shaker/tasks.hpp"
+#include "simulator/simulator_utils.hpp"
 #include "systemwide.h"
 
 using namespace system_thread;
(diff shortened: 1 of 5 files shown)


stm32-modules/heater-shaker/simulator/system_thread.cpp
```diff
@@ -66,10 +67,31 @@ struct system_thread::TaskControlBlock {
     SimSystemTask task;
 };
 
+template <size_t N>
```
amitlissack: I know it's a tiny function, but can't it be in `include/common/simulation`  rather than duplicated?

amitlissack (approved):

DerekMaggio (approved):
From what I understand about C++, it looks right. 
The name `SERIAL_NUMBER` is correct

DerekMaggio:
Tested this against my emulation system and the SERIAL_NUMBER env var is working as intended. 
Go @fsinapi :partying_face:

stm32-modules/heater-shaker/simulator/system_thread.cpp
fsinapi: Tbh I was avoiding setting up a `common-simulator` target... will set that up now 

EDIT nevermind, don't even need a library target to do that...

stm32-modules/include/common/simulator/simulator_utils.hpp
```diff
@@ -0,0 +1,40 @@
...
+        return RT();
+    }
+
+    auto *env_p = std::getenv(var_name);
+    if (!env_p || (strlen(env_p) == 0)) {
+        return RT();
+    }
+
+    std::array<char, N> ret;
+    memcpy(ret.data(), env_p, std::min(strlen(env_p), N));
```
sfoster1: https://en.cppreference.com/w/c/string/byte/strncpy ?
```suggestion
    strncpy(ret.data(), env_p, N));
```

sfoster1: not a big change but good habit to get in

sfoster1 (approved):
looks good but maybe strncpy

PR #244, opened 2022-02-01
feat(thermocycler-refresh): add seal motor driver

fsinapi:
### Summary

This PR adds two main features:

- A `MovementProfile` class that provides an interface for generating stepper motor acceleration profiles. Supports movements of either a fixed distance or an unknown distance (e.g. movements to be ended by stalls, stop switches, etc)
- Adds firmware to set up and control the Seal Stepper Motor on the TCR. Utilizes `TIM6` as a simple timer running at 1MHz and invokes a callback to the motor task with every interrupt. Uses `MovementProfile` to decide when to step the pulse pin on the TMC2130.

Also adds in a default configuration of the TMC2130 that gets loaded the first time `motor_task.run_once()` is invoked.

Tested on a bare board with a seal stepper attached by using the debug gcode M241.D, and the stepper moves as expected. Also tested changing the microstepping scaling on the TMC2130 configuration to make sure the speed increases as expected.


sfoster1 (approved):
Looks extremely solid!

PR #246, opened 2022-02-04
feat(thermocycler-refresh): Seal stepper stall detection

fsinapi:
### Summary

Adds GPIO interrupts for the two Diagnostic outputs from the TMC2130, one for errors and one for stalls. Now if an error or stall occurs during a seal motor movement a message containing the reason for is sent to the `motor_task` and the movement stops early.

* Added the Drive Status register to the TMC2130 registers driver. Read-only register that gives runtime status information.
* Added a gcode to get the Drive Status register info from the TMC2130. Requested by Jim and also useful for debugging.
* Added some code that clears out the StallGuard information on the TMC2130 before starting a movement. This isn't really documented by the datasheet, but in testing it resets the stall flag and makes stall checks reliable.
* Had to drop the motor interrupt to 500kHz instead of 1MHz. Messages weren't getting handled at all during motor movements. The nominal velocity we're using is ~50kHz stepping so we're still far above the Nyquist frequency.


sfoster1 (approved):
Looks good to me. Sounds like the frequency change won't be an issue.

PR #247, opened 2022-02-07
feat(thermocycler-refresh): Seal motor configuration command

fsinapi:
Adds a debug gcode (M243.D) intended for internal development testing with the seal motor driver subsystem

* M243.D allows configuration of velocity, acceleration, stallguard threshold, stallguard minimum velocity, and hold & drive current
* Tested by running commands for each config option through python utilities. As a bonus, was able to confirm that overheat detection works (if you crank the drive current up to max you get a fault during movement).

### DIFF
--- a/stm32-modules/include/thermocycler-refresh/thermocycler-refresh/messages.hpp
+++ b/stm32-modules/include/thermocycler-refresh/thermocycler-refresh/messages.hpp
@@ -34,7 +34,7 @@ concept Response = requires(ResponseType rt) {
 
 /*
 ** Message structs initiate actions. These may be changes in physical state, or
-** a request to send back some data. Each carries an ID, which should be copied4
+** a request to send back some data. Each carries an ID, which should be copied
 *V into
 ** the response.
 */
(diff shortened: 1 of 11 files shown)


stm32-modules/include/thermocycler-refresh/thermocycler-refresh/messages.hpp
```diff
@@ -33,8 +34,8 @@ concept Response = requires(ResponseType rt) {
 
 /*
 ** Message structs initiate actions. These may be changes in physical state, or
-** a request to send back some data. Each carries an ID, which should be copied
-*into
+** a request to send back some data. Each carries an ID, which should be copied4
```
sfoster1:
```suggestion
** a request to send back some data. Each carries an ID, which should be copied
```

stm32-modules/include/thermocycler-refresh/thermocycler-refresh/messages.hpp
```diff
@@ -33,8 +34,8 @@ concept Response = requires(ResponseType rt) {
...
+*V into
```
sfoster1:
```suggestion
*into
```

sfoster1 (approved):
Awesome. Just got a couple vim things in a comment looks like

PR #248, opened 2022-02-08
feat(thermocycler): Lid limit switches

fsinapi:
### Summary

Adds code to read the open/close limit switches, and polls them during lid movements to stop the motor when they are triggered.

The switches are _not_ set as interrupts because one of them conflicts with one of the ADC Ready interrupts on line 9. The ADC read timing is more important to the system as a whole, so the limit switches get read on a polling basis.

* Adds M119 GetLidStatus, which mirrors the version from the original Thermocycler but adds in a return variable including the status of the seal stepper.
* Tested by sending move commands to open/close the lid and making sure that 1) triggering the correct switch stops movement 2) the irrelevant switch is ignored completely.

Might end up having to add overdrive handling for close commands to make sure the lid is fully closed, that is TBD.


sfoster1 (approved):
That is very sad about the interrupts. Looks good!

PR #249, opened 2022-02-09
docs(modules): update readme for cmake

fsinapi:
Added a note in the readme that configuring the cmake presets will require a C and C++ compiler after noticing that `cmake --preset arduino` fails if there's no g++ on a Linux machine.


sfoster1 (approved):
👍

PR #250, opened 2022-02-09
feat(thermocycler-refresh): thermal debug command

fsinapi:
### Summary
Adds a command M103.D to get the current power output level for the peltiers, heater, and fans. This has been requested for initial thermal profiling data gathering.

- Also includes a much larger stack for the host comms task. There were crashes happening and the debugger showed the stack pointer way too high for what was allocated. I want to try to set up stack height monitoring when I've got a chance, for now just bumping up the stack size as a mitigation...
- New command depends on multiple tasks (plate and lid tasks), and thus uses the Ack Cache a little differently than other commands - it caches the original gcode while querying the first task, and then caches the response of the first task while querying the second task.
- Tested by setting some peltier/fan/heater levels on a PCBand reading them back with the command.

### DIFF
--- a/stm32-modules/thermocycler-refresh/firmware/host_comms_task/freertos_comms_task.cpp
+++ b/stm32-modules/thermocycler-refresh/firmware/host_comms_task/freertos_comms_task.cpp
@@ -50,7 +50,7 @@ static auto _top_task = host_comms_task::HostCommsTask(_comms_queue);
 // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
 static auto _tasks = std::make_pair(&_top_task, &_local_task);
 
-static constexpr uint32_t stack_size = 500;
+static constexpr uint32_t stack_size = 2000;
 // Stack as a std::array because why not
 // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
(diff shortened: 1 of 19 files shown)


stm32-modules/thermocycler-refresh/firmware/host_comms_task/freertos_comms_task.cpp
```diff
@@ -50,7 +50,7 @@ static auto _top_task = host_comms_task::HostCommsTask(_comms_queue);
 // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
 static auto _tasks = std::make_pair(&_top_task, &_local_task);
 
-static constexpr uint32_t stack_size = 500;
+static constexpr uint32_t stack_size = 2000;
```
sfoster1: make this a power of two you monster

sfoster1 (approved):
Interesting stuff with the two level cache, looks well tested.

stm32-modules/thermocycler-refresh/firmware/host_comms_task/freertos_comms_task.cpp
fsinapi: 🙃

PR #251, opened 2022-02-11
feat(thermocycler-refresh): Lid stepper overdrive movements

fsinapi:
Adds support for "overdrive" movements, during which the stop flags are ignored. This will mostly be used for the automated lid open/close state machines to move to precise angles based on limit switch positions, which will _not_ be at the literal end of travel.

The overdrive option can be set by setting a flag on the Lid Stepper debug movement command. Tested by sending the command to drive into the limit switches, works as intended on hardware.


sfoster1 (approved):
LGTM

PR #252, opened 2022-02-11
fix(thermocycler-refresh): DFU bootloader fix

fsinapi:
### Summary 

For some reason, after #225 was merged the DFU bootloader stopped working on the Thermocycler-Refresh MCU, and this wasn't noticed because the bootloader wasn't being used during development. The issue doesn't seem to be directly related to that PR, but instead relates to the NVIC not being fully cleared when the system goes into reset. This PR manually clears out the relevant registers and, in testing on both Rev 1 and Rev 2 boards, lets the bootloader successfully enumerate over USB as a DFU device.


sfoster1 (approved):
Yuck

PR #253, opened 2022-02-11
chore(thermocycler-refresh, heater-shaker): add preset to build all simulators

fsinapi:
Adds a cmake preset to build all of the simulator targets for the modules repo.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -144,6 +144,28 @@
             ]
         },
         {
+            "name": "simulators",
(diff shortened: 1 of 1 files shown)


DerekMaggio (approved):
LGTM!

I still need to run the build host command first right?

fsinapi:
@DerekMaggio  Yep, you should need two commands:
- `cmake --preset stm32-host`
- `cmake --build --preset simulators`

PR #255, opened 2022-02-14
feat(thermocycler-refresh): Add front button firmware setup

fsinapi:
### Summary

Adds in the firmware-specific code for the front button hardware.
- Sets up pin C11 as an interrupt for reading the button presses. This debounces the input and invokes an optional callback function if it has been over 200ms since the last press.
- Sets up pin D10 as an output to drive the LED on the button
- Does _not_ include higher level code to open/close the lid based on current state.

Tested by setting the C11 callback to toggle the LED if it would be invoking a callback, works as intended. Sometimes releasing the button triggers an interrupt but future code will be checking the state of the button _after_ the debounce period to decide if it was pressed. We don't have use for the full functionality yet, but having the LED working is nice.


stm32-modules/thermocycler-refresh/firmware/system/system_hardware.c
```diff
@@ -86,3 +125,23 @@ void system_hardware_enter_bootloader(void) {
...
+bool system_front_button_pressed(void) {
+    // Active low button, passively pulled high
+    return HAL_GPIO_ReadPin(FRONT_BUTTON_IN_PORT, FRONT_BUTTON_IN_PIN) 
+                == GPIO_PIN_RESET;
+}
+
+void system_front_button_callback(void) {
+    if(__HAL_GPIO_EXTI_GET_IT(FRONT_BUTTON_IN_PIN) != 0x00u) {
+        __HAL_GPIO_EXTI_CLEAR_IT(FRONT_BUTTON_IN_PIN);
+        uint32_t new_tick = HAL_GetTick();
```
sfoster1, line 155: we should probably think about overflow here - it's an appliance that might be left on all the time.

sfoster1 (changes requested):
Let's do overflow handling here. also a bit of a nit, but maybe set up an LED flasher as the system callback for demo purposes?

sfoster1 (approved):
Looks good to me!

PR #256, opened 2022-02-16
feat(thermocycler-refresh): initial thermal testing scripts and tuning

fsinapi:
### Summary

- Added `test_utils.py` functions for higher level lid open/close functionality for testing purposes.
- Added some scripting utilities for getting unit step response data (mostly hard coded), as well as for running an individual thermal cycle to 95 -> 5 -> 20 with hold times (again hard coded). 
- Adds some initial PID values for the peltiers and the lid heater. Tuning data for peltiers is [here](https://pidtuner.com/#/twHlGsVCsl), lid is [here](https://pidtuner.com/#/Urix6CbTm1). Values are not the default ones, I made the system a little more underdamped because the overshoot by default is more than reasonable.

The PID values are kind of rough for the plate, in particular the cooling is very strong when you get up around 95C so the damping is very asymmetrical.


sfoster1 (approved):
nice first pass!

PR #257, opened 2022-02-18
feat(thermocycler-refresh): automated LED strip behavior

fsinapi:
### Summary 

Sets up LED control matching the original thermocycler, where the state of the LED strip is based on the current Plate Control status:
- Errors cause the Orange Light Of Death (OLOD), and if the error condition goes away the light will cease
- When idle, the LED's are solid white
- Hot targets (>23C) are red, cold targets are blue
- The LED's pulse while ramping and are solid at target

Tested on a thermocycler by setting temperature targets and/or unplugging the thermistors, all of the states work as expected. There is some infrastructure for errors to come from the Lid and Motor tasks but right now that isn't utilized.


sfoster1 (approved):
awesome, love that LED work paying off

PR #258, opened 2022-02-23
fix(thermocycler-refresh, heater-shaker): remove iostream dependency

fsinapi:
Taking out this `#include` decreased the flash usage from ~50% to ~25%... and took out one of the `malloc` calls on system initialization.


sfoster1 (approved):
Guh can't believe I left this in there! Sorry!

DerekMaggio:
> Guh can't believe I left this in there! Sorry!

Oh how the mighty have fallen :P

PR #259, opened 2022-02-24
feat(thermocycler-refresh): add EEPROM driver

fsinapi:
### Summary

Adds a driver for the AT24C02C EEPROM used on the Thermocycler Refresh board. 
- EEPROM driver reads & writes in terms of 8-byte pages, with support for writing arbitrary data types to each page.
- Tested the firmware portions by temporarily modifying the GetPlateTemp response to read a page in the EEPROM as a `double`, report that number back in the response, and then write the number it read + 1 back to the EEPROM. The number incremented as expected.

### DIFF
--- a/stm32-modules/include/common/core/at24c0xc.hpp
+++ b/stm32-modules/include/common/core/at24c0xc.hpp
@@ -59,7 +59,7 @@ class AT24C0xC {
     // Either a 1024 or 2048 bit device
     static_assert((PAGES == 16) || (PAGES == 32),
                   "EEPROM size must be 1024 or 2048 bits");
-        
+
     // Address is lower 7 bits
     static_assert(ADDRESS < MAX_ADDR, "Address must be a 7-bit value");
 
@@ -91,8 +91,8 @@ class AT24C0xC {
         // Actual address is based on the byte.
         BufferT buffer;
         buffer.at(0) = page * PAGE_LENGTH;
-        auto *itr = bit_utils::int_to_bytes(value_int,
-                                           buffer.begin() + 1, buffer.end());
+        auto *itr = bit_utils::int_to_bytes(value_int, buffer.begin() + 1,
+                                            buffer.end());
         if (itr != buffer.end()) {
             // Error converting data
             return false;
@@ -116,7 +116,8 @@ class AT24C0xC {
      */
     template <typename T, AT24C0xC_Policy Policy>
     requires std::is_trivially_copyable_v<T>
-    [[nodiscard]] auto read_value(uint8_t page, Policy &policy) -> std::optional<T> {
+    [[nodiscard]] auto read_value(uint8_t page, Policy &policy)
+        -> std::optional<T> {
         using RT = std::optional<T>;
         using BufferT = std::array<uint8_t, PAGE_LENGTH>;
         // Check memory bounds
--- a/stm32-modules/include/common/core/at24c0xc.hpp
+++ b/stm32-modules/include/common/core/at24c0xc.hpp
@@ -3,10 +3,10 @@
  * @brief Implements a generic driver for the AT24C0xC EEPROM IC's.
  */
 
-#pragma once 
+#pragma once
 
 #include <array>
-#include <cstring> // For memcpy
+#include <cstring>  // For memcpy
 
 #include "core/bit_utils.hpp"
 
@@ -16,9 +16,9 @@ namespace at24c0xc {
 static constexpr const size_t PAGE_LENGTH = 8;
 
 template <typename Policy>
-concept AT24C0xC_Policy = requires(Policy& policy, uint8_t addr, 
-        std::array<uint8_t, PAGE_LENGTH + 1> send,
-        std::array<uint8_t, PAGE_LENGTH> receive) {
+concept AT24C0xC_Policy = requires(Policy &policy, uint8_t addr,
+                                   std::array<uint8_t, PAGE_LENGTH + 1> send,
+                                   std::array<uint8_t, PAGE_LENGTH> receive) {
     // Function to write a page (8 bytes)
     { policy.i2c_write(addr, send) } -> std::same_as<bool>;
     // Function to write a singlye byte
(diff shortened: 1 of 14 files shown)


stm32-modules/include/common/core/at24c0xc.hpp
```diff
@@ -0,0 +1,156 @@
...
+namespace at24c0xc {
+
+static constexpr const size_t PAGE_LENGTH = 8;
+
+template <typename Policy>
+concept AT24C0xC_Policy = requires(Policy &policy, uint8_t addr,
+                                   std::array<uint8_t, PAGE_LENGTH + 1> send,
+                                   std::array<uint8_t, PAGE_LENGTH> receive) {
+    // Function to write a page (8 bytes)
+    { policy.i2c_write(addr, send) } -> std::same_as<bool>;
```
sfoster1: I never really like passing arrays around like this because it makes it harder and wordier to do things like have this write into/out of a larger buffer at an offset. Maybe this could be an iterator pair or a [span](https://en.cppreference.com/w/cpp/container/span)? cpp20 even has [as_writable_bytes](https://en.cppreference.com/w/cpp/container/span/as_bytes) for integration with the c side

stm32-modules/include/common/core/at24c0xc.hpp
```diff
@@ -0,0 +1,156 @@
...
+    // Function to write a singlye byte
```
sfoster1:
```suggestion
    // Function to write a single byte
```

stm32-modules/include/common/core/at24c0xc.hpp
sfoster1: That said, the implementation of this only ever being called with a stack array created right before the call makes me worry a little less

stm32-modules/include/common/core/at24c0xc.hpp
```diff
@@ -0,0 +1,156 @@
...
+        if (page > PAGES) {
+            return false;
+        }
+        // Because T must be trivially copyable, this is not a dangerous copy
+        uint64_t value_int = 0;
+        std::memcpy(&value_int, &value, sizeof(T));
+        // Actual address is based on the byte.
+        BufferT buffer;
+        buffer.at(0) = page * PAGE_LENGTH;
+        auto *itr = bit_utils::int_to_bytes(value_int, buffer.begin() + 1,
```
sfoster1: do we get anything out of this? this isn't an on the wire transaction to some other system, it's just us writing to and reading from this eeprom. feels like we might as well just `memcpy(buffer.data() + 1, &value, sizeof(value))`

stm32-modules/include/common/core/at24c0xc.hpp
```diff
@@ -0,0 +1,156 @@
...
+        static_assert(sizeof(T) <= PAGE_LENGTH,
+                      "Type T must be 8 bytes max to serialize");
+        using BufferT = std::array<uint8_t, PAGE_LENGTH + 1>;
+        // Check memory bounds
+        if (page > PAGES) {
+            return false;
+        }
+        // Because T must be trivially copyable, this is not a dangerous copy
+        uint64_t value_int = 0;
+        std::memcpy(&value_int, &value, sizeof(T));
```
sfoster1:
```suggestion
        std::memcpy(&value_int, &value, sizeof(value));
```
kind of a nitpick since this is templated but it's a reflex

stm32-modules/include/common/core/at24c0xc.hpp
```diff
@@ -0,0 +1,156 @@
...
+        // Must write the address before reading everything else
+        if (!policy.i2c_write(_address, page * PAGE_LENGTH)) {
+            return std::nullopt;
+        }
+
+        BufferT buffer = {0};
+        if (!policy.i2c_read(_address, buffer)) {
+            return std::nullopt;
+        }
+        uint64_t value_int = 0;
```
sfoster1: the dual comment to `write` here - what do we get out of `bytes_to_int` followed by `memcpy`? why not just `memcpy`?

stm32-modules/include/common/core/at24c0xc.hpp
```diff
@@ -0,0 +1,156 @@
...
+        const auto *itr = bit_utils::bytes_to_int(buffer, value_int);
+        if (itr != buffer.end()) {
+            return std::nullopt;
+        }
+        T value;
+        memcpy(&value, &value_int, sizeof(T));
```
sfoster1:
```suggestion
        memcpy(&value, &value_int, sizeof(value));
```
reflex

sfoster1 (approved):
Looks good generally but I had some nitpicks in the eeprom policy spec and portable object

stm32-modules/include/common/core/at24c0xc.hpp
fsinapi: Yeah in hindsight I'm not sure why I added the intermediate conversion. Makes sense to just use `memcpy`

stm32-modules/include/common/core/at24c0xc.hpp
fsinapi: same as above

stm32-modules/include/common/core/at24c0xc.hpp
fsinapi: Iterator pair sounds like a solid improvement

stm32-modules/include/common/core/at24c0xc.hpp
```diff
@@ -0,0 +1,146 @@
...
+
+#include "core/bit_utils.hpp"
+
+// Namespace for AT24C01C / AT24C02C driver
+namespace at24c0xc {
+
+static constexpr const size_t PAGE_LENGTH = 8;
+
+template <typename Policy>
+concept AT24C0xC_Policy = requires(Policy &policy, uint8_t addr,
```
sfoster1, line 19: looks good; I tend to personally prefer iterator-sentinel pairs rather than iterator-size, and I'd love to find a way to make it not require specifically array iterator,s but this definitely is good

sfoster1 (approved):
Nice, look good to me!

stm32-modules/include/common/core/at24c0xc.hpp
fsinapi: I was going to initially go with that but the issue is passing the buffer to the HAL I2C driver, which takes a pointer + length. With the iterator-sentinel pair the policy would end up just having to call `std::distance` on the two iterators to get the length.

I've got some i2c refactor stuff ticketed in Jira, will make sure to take another pass at this with that work

PR #260, opened 2022-03-02
feat(thermocycler-refresh): Plate to Thermistor Offset Configuration

fsinapi:
### Summary

The temperature measured by the thermistors on the plate PCB is not the same as the temperature on the actual plate. Early testing shows that a simple linear regression based off of the current temperature of a thermistor can approximate (within spec) the temperature on the plate.

- Adds a dedicated `Eeprom` class for reading/writing offsets to the onboard EEPROM. Includes a flag on one of the pages to mark whether the data should be "good" or not, and in the future can mark whether a checksum is included
- Adds calculation of the expected temperature for the 6 plate thermistors.
- Adds a short FreeRTOS delay after setting the write protect pin on the EEPROM. The EEPROM needs time after each write and the datasheet suggest polling the bus, but it's a fixed time and a delay gets the job done. Earlier testing had all been with a single write at a time, so no failures popped up until now.

Tested on hardware that:
- Setting constants affects the reported temperature as expected
- Constants are persistent across power cycles

### DIFF
--- a/stm32-modules/include/thermocycler-refresh/thermocycler-refresh/eeprom.hpp
+++ b/stm32-modules/include/thermocycler-refresh/thermocycler-refresh/eeprom.hpp
@@ -1,0 +1,158 @@
+/**
+ * @file eeprom.hpp
+ * @brief Implements an EEPROm class that is specialized towards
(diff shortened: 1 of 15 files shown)


sfoster1 (approved):
looks good to me

PR #261, opened 2022-03-07
feat(thermocycler-refresh): Open and Close Lid Commands

fsinapi:
### Summary

Adds M126 and M127 from the original thermocycler to Open and Close the lid, respectively.  

- Reporting of the lid position (open/closed/unknown) is now dependent on the last command sent for the lid rather than just the switches. This is required because the resting positions for open & closed may not necessarily trigger the endstop switches.
- The state machines for each command are TBD to a certain extent, but they are tested and working correctly on the 1.4 units (most recent prototypes).
- Edited the Python utils to use the new commands instead of chaining together individual commands.

Tested on prototype 1.4A

### DIFF
--- a/stm32-modules/include/thermocycler-refresh/thermocycler-refresh/motor_task.hpp
+++ b/stm32-modules/include/thermocycler-refresh/thermocycler-refresh/motor_task.hpp
@@ -77,13 +77,53 @@ concept MotorExecutionPolicy = requires(Policy& p,
 }
 &&tmc2130::TMC2130Policy<Policy>;
 
-// Structure to encapsulate state of either of the stepper motors
-struct StepperState {
-    enum StepperTaskStatus { IDLE, MOVING };
+// Structure to encapsulate state of the lid stepper
+struct LidStepperState {
+    // Full open/close movements run until they hit an endstop switch, so the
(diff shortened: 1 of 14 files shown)


stm32-modules/include/thermocycler-refresh/thermocycler-refresh/motor_task.hpp
```diff
@@ -77,13 +77,56 @@ concept MotorExecutionPolicy = requires(Policy& p,
...
-// Structure to encapsulate state of either of the stepper motors
-struct StepperState {
-    enum StepperTaskStatus { IDLE, MOVING };
+// Structure to encapsulate state of the lid stepper
+struct LidStepperState {
+    // Full open/close movements run until they hit an endstop switch, so the
+    // distance is 120 degrees which is far wider than the actual travel angle.
+    constexpr static double FULL_OPEN_DEGREES =
+        motor_util::LidStepper::angle_to_microsteps(120);
+    // After opening to the open switch, the lid must close ~5º to
```
sfoster1: wait is it 5 or 17? maybe just take the explicit number out of the comment

sfoster1 (approved):
Looks good to me, comment/code disagreement aside.

This gcode cache stuff is getting pretty bad, huh.

PR #262, opened 2022-03-08
feat(thermocycler-refresh): front button controls the lid

fsinapi:
### Summary

This PR makes the front button control the lid. Pressing the button while the lid is fully open will close the lid, and pressing it at any other time will close the lid. If both of the endstop switches sense the lid, nothing happens because something is broken.

Tested on 1.4A.


sfoster1 (approved):
lgtm, didn't know about giving names to lambda captures like that, cool!

fsinapi:
> lgtm, didn't know about giving names to lambda captures like that, cool!

Haha that's basically copy+pasted from what the Linter suggests if you try to use `std::bind`

PR #263, opened 2022-03-10
feat(thermocycler-refresh): Hold time tracking and more detailed GetPlateTemp response

fsinapi:
### Summary

Adds hold time tracking for plate targets. The firmware tracks the amount of time that has passed once the target temperature is hit. This doesn't actually control anything in the firmware, but it's reported back to the host computer.

- Updated the python utilities to grab new information from the GetPlateTemp response
- Tested on 1.4A unit, and unit tests were updated to check for new functionality

### DIFF
--- a/stm32-modules/include/thermocycler-refresh/thermocycler-refresh/gcodes.hpp
+++ b/stm32-modules/include/thermocycler-refresh/thermocycler-refresh/gcodes.hpp
@@ -200,15 +200,26 @@ struct GetPlateTemp {
         std::sized_sentinel_for<InputIt, InLimit>
     static auto write_response_into(InputIt buf, InLimit limit,
                                     double current_temperature,
-                                    double setpoint_temperature) -> InputIt {
+                                    double setpoint_temperature = 0.0F,
+                                    double remaining_hold = 0.0F,
+                                    double total_hold = 0.0F,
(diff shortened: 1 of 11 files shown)


sfoster1 (changes requested):
If the update rate specification is stable this looks good to me but it's a little... gross. I guess the stability is required for the control anyway, but at this point it seems like `update_rate` is only used to drive the time-left estimation; maybe getting the current tick, or using a freertos time-elapsed method, would be better?

fsinapi:
@sfoster1 That change makes sense. The sample time is also used for ramp rate calculation and it's implied to match the PID controller's sample time. If we're getting the tick difference for each update then I think we should also update the PID controller to accept a variable sample time.

sfoster1:
> @sfoster1 That change makes sense. The sample time is also used for ramp rate calculation and it's implied to match the PID controller's sample time. If we're getting the tick difference for each update then I think we should also update the PID controller to accept a variable sample time.

We definitely should at some point, though I see your point that maybe that shouldn't be now.

Oh, and I see where this value is SSOT'd from the control period in the freertos task, cool. Just didn't know where it's coming from.

sfoster1 (approved):
lgtm!

PR #264, opened 2022-03-10
feat(heater-shaker): added thermal offsets

pmoegenburg:
tested successfully (values persisted through multiple power cycles), need to write tests

### DIFF
--- a/stm32-modules/heater-shaker/firmware/heater_task/heater_hardware.c
+++ b/stm32-modules/heater-shaker/firmware/heater_task/heater_hardware.c
@@ -1,4 +1,5 @@
 #include <stdlib.h>
+#include <stddef.h>
 
 #include "stm32f3xx_hal.h"
 #include "stm32f3xx_hal_adc.h"
@@ -59,6 +60,7 @@ heater_hardware *HEATER_HW_HANDLE = NULL;
 #define HEATER_PAD_ENABLE_TIM_CHANNEL TIM_CHANNEL_3
 #define HEATER_PAD_LL_SETCOMPARE LL_TIM_OC_SetCompareCH3
 
+static uint32_t OFFSETS_PAGE_ADDRESS = 0x0807F000; //second last page in flash memory. Last page reserved for serial number storage
 
 static void gpio_setup(void) {
     // NTC sense pis all routed to the ADC
@@ -282,6 +284,33 @@ void HAL_ADC_ConvCpltCallback(ADC_HandleTypeDef* hadc) {
             }
             break;}
     }
+}
+
+bool heater_hardware_set_offsets(struct writable_offsets* to_write) {
+    FLASH_EraseInitTypeDef pageToErase = {.TypeErase = FLASH_TYPEERASE_PAGES, .PageAddress = OFFSETS_PAGE_ADDRESS, .NbPages = 1};
+    uint32_t pageErrorPtr = 0; //pointer to variable  that contains the configuration information on faulty page in case of error
+    uint32_t ProgramAddress1 = OFFSETS_PAGE_ADDRESS + (offsetof(struct writable_offsets, const_b) * 8); //offsetof returns bytes
+    uint32_t ProgramAddress2 = OFFSETS_PAGE_ADDRESS + (offsetof(struct writable_offsets, const_c) * 8);
+    uint32_t ProgramAddress3 = OFFSETS_PAGE_ADDRESS + (offsetof(struct writable_offsets, const_flag) * 8);
+
+    HAL_StatusTypeDef status = HAL_FLASH_Unlock();
+    if (status == HAL_OK) {
+        status = HAL_FLASHEx_Erase(&pageToErase, &pageErrorPtr);
+        if (status == HAL_OK) {
+            status = HAL_FLASH_Program(FLASH_TYPEPROGRAM_DOUBLEWORD, ProgramAddress1, to_write->const_b);
+            status = HAL_FLASH_Program(FLASH_TYPEPROGRAM_DOUBLEWORD, ProgramAddress2, to_write->const_c);
+            status = HAL_FLASH_Program(FLASH_TYPEPROGRAM_DOUBLEWORD, ProgramAddress3, to_write->const_flag);
+            if (status == HAL_OK) {
+                status = HAL_FLASH_Lock();
+            }
+        }
+    }
+    return (status == HAL_OK);
+}
+
+uint64_t heater_hardware_get_offset(size_t addr_offset) {
+    uint32_t AddressToRead = OFFSETS_PAGE_ADDRESS + (addr_offset * 8); //addr_offset in bytes
+    return *(uint64_t*)AddressToRead;
 }
 
 static void init_error(void) {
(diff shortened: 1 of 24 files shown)


stm32-modules/include/heater-shaker/heater-shaker/gcodes.hpp
```diff
@@ -1010,4 +1010,123 @@ struct GetPlateLockStateDebug {
     }
 };
 
+/**
+ * Uses M116, as defined on Gen 1 thermocyclers.
```
fsinapi: Should probably update comment here

stm32-modules/include/heater-shaker/heater-shaker/gcodes.hpp
```diff
@@ -1010,4 +1010,123 @@ struct GetPlateLockStateDebug {
...
+            ret.const_c.defined = true;
+            ret.const_c.value = c.first.value();
+            working = c.second;
+        }
+        return std::make_pair(ParseResult(ret), working);
+    }
+};
+
+/**
+ * Uses M117, as defined on Gen 1 thermocyclers.
```
fsinapi: Same as previous comment

stm32-modules/heater-shaker/firmware/heater_task/heater_hardware.c
```diff
@@ -59,6 +60,7 @@ heater_hardware *HEATER_HW_HANDLE = NULL;
 #define HEATER_PAD_ENABLE_TIM_CHANNEL TIM_CHANNEL_3
 #define HEATER_PAD_LL_SETCOMPARE LL_TIM_OC_SetCompareCH3
 
+static uint32_t OFFSETS_PAGE_ADDRESS = 0x0807F000; //second last page in flash memory. Last page reserved for serial number storage
```
fsinapi: Does the linker overwrite this page? Might have to update the linker script.

sfoster1:
```suggestion
static const uint32_t OFFSETS_PAGE_ADDRESS = 0x0807F000; //second last page in flash memory. Last page reserved for serial number storage
```

fsinapi (commented):
Looks good, just a couple of comment holdovers and the linker might need updating. I'll review again once the tests are pushed 👍

stm32-modules/heater-shaker/firmware/heater_task/heater_hardware.c
```diff
@@ -284,6 +286,33 @@ void HAL_ADC_ConvCpltCallback(ADC_HandleTypeDef* hadc) {
     }
 }
 
+bool heater_hardware_set_offsets(struct writable_offsets* to_write) {
+    FLASH_EraseInitTypeDef pageToErase = {.TypeErase = FLASH_TYPEERASE_PAGES, .PageAddress = OFFSETS_PAGE_ADDRESS, .NbPages = 1};
+    uint32_t pageErrorPtr = 0; //pointer to variable  that contains the configuration information on faulty page in case of error
+    uint32_t ProgramAddress1 = OFFSETS_PAGE_ADDRESS + (offsetof(struct writable_offsets, const_b) * 8); //offsetof returns bytes
```
sfoster1: wait this doesn't quite click with me - there's no way program address is a bit address. Shouldn't this be a `/ 8`?

stm32-modules/heater-shaker/firmware/heater_task/heater_hardware.c
```diff
@@ -284,6 +286,33 @@ void HAL_ADC_ConvCpltCallback(ADC_HandleTypeDef* hadc) {
...
+                status = HAL_FLASH_Lock();
+            }
+        }
+    }
+    return (status == HAL_OK);
+}
+
+uint64_t heater_hardware_get_offset(size_t addr_offset) {
+    uint32_t AddressToRead = OFFSETS_PAGE_ADDRESS + (addr_offset * 8); //addr_offset in bytes
+    return *(uint64_t*)AddressToRead;
```
sfoster1, line 313:
```suggestion
    return (uint64_t)AddressToRead;
```

stm32-modules/include/heater-shaker/heater-shaker/flash.hpp
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * One of the FLASH addresses is reserved for a flag to indicate whether
+ * the values have been written. The \ref FLASHFlag enum captures the
+ * valid states of this page. The page indicates what error detection,
+ * if any, is included with the FLASH constant values.
+ *
+ */
+struct OffsetConstants {
+    // The value of the constants B and C
+    double b, c, flag;
```
sfoster1: this makes `flag` a `double` also, should it be a bool or something

sfoster1 (changes requested):
The rest of the code looks pretty could but I've gotten myself pretty confused about these flash addresses. Those are doubleword addresses, and offsetof gives byte offsets, so shouldn't 
- byte offset / 8 = doubleword offset
- doubleword offset * 8 = byte offset
?

stm32-modules/include/heater-shaker/heater-shaker/flash.hpp
pmoegenburg: I made it a double for consistency and because it will eventually be converted to uint64_t to be programmed into flash

stm32-modules/heater-shaker/firmware/heater_task/heater_hardware.c
pmoegenburg: AddressToRead is flash register memory address, so here we can convert it to a pointer to match/indicate that and then dereference it to access its data, as was done with serial number retrieval

stm32-modules/heater-shaker/firmware/heater_task/heater_hardware.c
pmoegenburg: ProgramAddressX is in bytes, so we can just use offsetof(). Got rid of the multiply/divide by 8

stm32-modules/heater-shaker/firmware/heater_task/heater_hardware.c
sfoster1: Oh yup I was misinterpreting that, sorry

stm32-modules/include/heater-shaker/heater-shaker/flash.hpp
sfoster1: Here in the cpp side, we should make the type be what it actually means - an enum or something. Writing it as a uint64_t is an implementation detail that should be hidden at this level.

sfoster1 (approved):
Nice, looks good to me!

stm32-modules/include/heater-shaker/heater-shaker/flash.hpp
```diff
@@ -92,7 +92,7 @@ class Flash {
     [[nodiscard]] auto initialized() const -> bool { return _initialized; }
 
     // Enumeration of the FLASH_CONST_FLAG values
-    enum class FLASHFlag : uint64_t { WRITTEN_NO_CHECKSUM = 1, INVALID = 0xFF };
+    enum class FLASHFlag : uint64_t { WRITTEN_NO_CHECKSUM = 1, INVALID = 0 };
```
fsinapi: If this is matching to the value read off of the flash, then you'll never see the value `INVALID` because the default erase value of the flash is all `1`'s

fsinapi (approved):
Looking good!

PR #266, opened 2022-03-16
feat(heater-shaker): added UART capability

pmoegenburg:
Initial code, testing/debugging in progress


stm32-modules/heater-shaker/firmware/host_comms_task/freertos_comms_task.cpp
```diff
@@ -27,10 +28,12 @@ struct CommsTaskFreeRTOS {
     USBD_CDC_ItfTypeDef cdc_class_fops;
     USBD_HandleTypeDef usb_handle;
     USBD_CDC_LineCodingTypeDef linecoding;
-
+    UART_HandleTypeDef uart_handle;
     double_buffer::DoubleBuffer<char, CDC_DATA_HS_MAX_PACKET_SIZE * 4> rx_buf;
     double_buffer::DoubleBuffer<char, CDC_DATA_HS_MAX_PACKET_SIZE * 4> tx_buf;
     char *committed_rx_buf_ptr;
+    double_buffer::DoubleBuffer<char, CDC_DATA_HS_MAX_PACKET_SIZE * 4> uart_rx_buf; //make same size as USB buffers?
```
fsinapi: Update comment

sfoster1: it doesn't really have to be, it could be smaller if you want. the USB double buff has to be that size because thats the maximum possible size of a USB packet. the uart needs less but also if we've got the memory available who cares

stm32-modules/heater-shaker/firmware/host_comms_task/uart_hardware.c
```diff
@@ -0,0 +1,108 @@
...
+  huart->Init.HwFlowCtl  = UART_HWCONTROL_NONE;
+  huart->Init.Mode       = UART_MODE_TX_RX;
+  huart->AdvancedInit.AdvFeatureInit = UART_ADVFEATURE_NO_INIT; 
+  if(HAL_UART_DeInit(huart) != HAL_OK)
+  {
+    Error_Handler();
+  }  
+}
+
+static void Error_Handler(void)
```
fsinapi: If we're using the HAL library's `Error_Handler`, this file probably isn't the right place to define it. This should either be in a more obvious file location, or you can switch to using the `configASSERT` function from FreeRTOS

stm32-modules/heater-shaker/firmware/host_comms_task/freertos_comms_task.cpp
```diff
@@ -27,10 +28,12 @@ struct CommsTaskFreeRTOS {
     USBD_CDC_ItfTypeDef cdc_class_fops;
     USBD_HandleTypeDef usb_handle;
     USBD_CDC_LineCodingTypeDef linecoding;
-
+    UART_HandleTypeDef uart_handle;
```
sfoster1, line 32: this struct is starting to be a lot; maybe could we make a `struct CommsTaskUSBData` and `struct CommsTaskUARTData` and then have `struct CommsTaskFreeRTOS { CommsTaskUSBData usb; CommsTaskUARTData uart; };`?

stm32-modules/heater-shaker/firmware/host_comms_task/freertos_comms_task.cpp
```diff
@@ -116,14 +123,24 @@ void run(void *param) {  // NOLINT(misc-unused-parameters)
...
+            _local_task.uart_rx_buf.committed()->size()));
     while (true) {
         char *tx_end =
             top_task->run_once(local_task->tx_buf.accessible()->begin(),
                                local_task->tx_buf.accessible()->end());
         if (!top_task->may_connect()) {
             USBD_Stop(&_local_task.usb_handle);
-        } else if (tx_end != local_task->tx_buf.accessible()->data()) {
+            UART_DeInit(&_local_task.uart_handle);
+        } else if ((tx_end != local_task->tx_buf.accessible()->data()) && (UartReady == SET)) {
```
sfoster1: this makes me a little nervous; the UART is entirely a secondary capacity, and this will block USB output on the uart being ready. We should either use a doublebuf on the tx side for the uart or go full DMA or something so we can remove the `(UartReady == SET)` condition.

stm32-modules/heater-shaker/firmware/host_comms_task/freertos_comms_task.cpp
```diff
@@ -96,6 +101,8 @@ extern __ALIGN_BEGIN uint8_t
     USBD_CDC_CfgFSDesc[USB_CDC_CONFIG_DESC_SIZ] __ALIGN_END;
 }
 
+__IO ITStatus UartReady = SET;
```
fsinapi: I think this should be `static`

stm32-modules/include/heater-shaker/heater-shaker/host_comms_task.hpp
```diff
@@ -138,6 +138,7 @@ class HostCommsTask {
     }
 
     [[nodiscard]] auto may_connect() const -> bool { return may_connect_latch; }
+    [[nodiscard]] auto uart_may_connect() const -> bool { return uart_may_connect_latch; }
```
fsinapi: Is there a scenario where `uart_may_connect_latch` won't be the same as `may_connect_latch`? Might be able to omit this.

stm32-modules/heater-shaker/firmware/host_comms_task/freertos_comms_task.cpp
```diff
@@ -308,3 +332,52 @@ static auto CDC_Receive(uint8_t *Buf, uint32_t *Len) -> int8_t {
...
+    } else {
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic)
+        _local_task.committed_rx_buf_ptr += (buffer_count - Buf);
+    }
+    
+    HAL_UART_Receive_IT(
+        &_local_task.uart_handle,
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
+        reinterpret_cast<uint8_t *>(
+            _local_task.uart_rx_buf.committed()->data()),
```
sfoster1: this has to be `_local_task.committed_rx_buf_ptr`. the logic in the block above does the following:
- See if there's a newline here indicating that the command is completed, or the buffer is out of space
- If so, send it to the task
- If not, advance the buffer input cursor past what was received so we append the next data to it

I think this isn't actually used in the USB side because USB is packetized enough that you really don't get incomplete commands often, but it's important here - we're gonna be getting a lot of interrupts that have like one character in the input buffer.

So here we want to use `_local_task.committed_rx_buf_ptr` for the pointer and probably add something to keep track of the remaining size.

stm32-modules/heater-shaker/firmware/host_comms_task/uart_hardware.c
sfoster1: We should have a globally-visible `Error_Handler` somewhere; let's not add another one here

sfoster1 (commented):
A couple things I think we'll need to fix (didn't approve/request changes because you mentioned you're not done with it)

fsinapi (commented):
Looking pretty good! Not sure if you've already tested it, but I would double check that comms work fine with any combination of USB or UART connected (just USB, just UART, both connected).

stm32-modules/heater-shaker/firmware/host_comms_task/freertos_comms_task.cpp
```diff
@@ -101,7 +101,7 @@ extern __ALIGN_BEGIN uint8_t
     USBD_CDC_CfgFSDesc[USB_CDC_CONFIG_DESC_SIZ] __ALIGN_END;
 }
 
-__IO ITStatus UartReady = SET;
+static bool UartReady = true;
```
fsinapi: Using `std::atomic_bool` here might be useful, I think there's a chance that caching issues could pop up since this is set in an interrupt

stm32-modules/heater-shaker/firmware/host_comms_task/freertos_comms_task.cpp
```diff
@@ -308,3 +333,57 @@ static auto CDC_Receive(uint8_t *Buf, uint32_t *Len) -> int8_t {
     USBD_CDC_ReceivePacket(&_local_task.usb_handle);
     return USBD_OK;
 }
+
+void HAL_UARTEx_RxEventCallback(UART_HandleTypeDef *UartHandle, uint16_t Size) {
+    using namespace host_comms_control_task;
+    uint8_t *Buf =
+        reinterpret_cast<uint8_t *>(_local_task.committed_uart_rx_buf_ptr);
+    ssize_t Len = (ssize_t)(Size);
+    ssize_t remaining_buffer_count = (ssize_t)(UART_BUFFER_MAX_SIZE)-Len;
```
fsinapi: If this represents the remaining contents of the buffer, shouldn't it be `_local_task.committed() + UART_BUFFER_MAX_SIZE - (_local_task.committed_uart_rx_buf_ptr + Size)`? As it is, it only provides the total size of the UART buffer less the number of bytes _in this Rx callback_

stm32-modules/heater-shaker/firmware/host_comms_task/uart_hardware.h
```diff
@@ -0,0 +1,35 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif  // __cplusplus
+
+void UART_Init(UART_HandleTypeDef *huart);
+void UART_DeInit(UART_HandleTypeDef *huart);
+void USART2_IRQHandler(void);
+
+#define UART_BUFFER_MAX_SIZE 64U
+#define UART_BUFFER_MIN_SIZE 16U
```
fsinapi: Is this the max number of bytes that one UART callback will contain, or does it come from something else? A comment about the reasoning for this number would be helpful.

stm32-modules/heater-shaker/firmware/host_comms_task/freertos_comms_task.cpp
```diff
@@ -132,6 +148,15 @@ void run(void *param) {  // NOLINT(misc-unused-parameters)
                     local_task->tx_buf.committed()->data()),
                 tx_end - local_task->tx_buf.committed()->data());
             USBD_CDC_TransmitPacket(&_local_task.usb_handle);
+            if (UartReady) {
+                UartReady = false;
```
fsinapi, line 153: More of a question: does sending data on the UART interfere with debugging operation at all? I.e. if you're debugging the Heater Shaker with an STLink instead of using it as a UART bridge, does the debugger work fine?

fsinapi (changes requested):
The only change I feel strongly about is the update to `remaining_buffer_count`, otherwise looks really good!

stm32-modules/heater-shaker/firmware/host_comms_task/uart_hardware.h
pmoegenburg: The max value is an arbitrary value that is significantly larger than any gcode input command. If the remaining buffer size is less than the min value, we may not have enough storage for a subsequent gcode input.

stm32-modules/heater-shaker/firmware/host_comms_task/freertos_comms_task.cpp
pmoegenburg: We can actually debug while sending and receiving data over UART simultaneously!

fsinapi (approved):
Nice!

PR #267, opened 2022-03-16
chore(thermocycler-refresh): better PID and hold timing

fsinapi:
### Summary

Previously, the code would set a fixed frequency for thermistor readings to be polled at and would update all of the control loops under the assumption that the time was always the same. This PR adds a timestamp to the internal thermistor messages to give the tasks an actual idea of how long the readings take.

Tested on a TCR board polling thermistors by checking in the debugger that the timestamps increase at the right amount for each reading.

### DIFF
--- a/stm32-modules/include/thermocycler-refresh/thermocycler-refresh/lid_heater_task.hpp
+++ b/stm32-modules/include/thermocycler-refresh/thermocycler-refresh/lid_heater_task.hpp
@@ -62,6 +62,7 @@ class LidHeaterTask {
   public:
     using Queue = QueueImpl<Message>;
     using Milliseconds = uint32_t;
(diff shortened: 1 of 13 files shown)


stm32-modules/include/thermocycler-refresh/thermocycler-refresh/lid_heater_task.hpp
```diff
@@ -61,6 +61,8 @@ requires MessageQueue<QueueImpl<Message>, Message>
 class LidHeaterTask {
   public:
     using Queue = QueueImpl<Message>;
+    using Milliseconds = uint32_t;
```
sfoster1: https://en.cppreference.com/w/cpp/chrono/duration ? would understand if you don't want to

sfoster1 (approved):
Nice, great call on how to do this

PR #268, opened 2022-03-17
feat(thermocycler-refresh): Report actual steps moved after seal stepper movements

fsinapi:
### Summary

Testing with the seal stepper needs the ability to see the actual number of steps that a movement ended up taking. This will be used to evaluate repeatability and accuracy of the stall system.

- Added the number of steps taken as a response to the M241.D command used for seal stepper debug movement
- Removed the error response if a stall occurs during a seal stepper movement - this is expected and whether it's an error or not can be inferred from the step count

Tested by sending M241.D for both full movements and stalled movements, confirmed that both positive and negative step counts are reported reliably.


sfoster1 (approved):
Nice.

PR #269, opened 2022-03-17
feat(heater-shaker): added thermal offset read precision

pmoegenburg:
Linear fit of temperature data requires constant value with three decimal places of precision. This allows the reading of those constants at that precision.


sfoster1 (approved):

fsinapi (approved):
Nice

PR #270, opened 2022-03-17
feat(thermocycler-refresh): Deactivate All command

fsinapi:
Adds command M18 to deactivate both the plate and the heater for backwards compatibility with the TC Gen 1. Similarly to Get Thermal Power, this gcode gets bounced to two separate tasks using the ack cache in a kind of kludgy way. 

Tested on my home setup.

Also bumped up the stack size of a few tasks, most importantly the Plate Thermistor reading task that was encountering stack overflow issues.

### DIFF
--- a/stm32-modules/include/thermocycler-refresh/thermocycler-refresh/gcodes.hpp
+++ b/stm32-modules/include/thermocycler-refresh/thermocycler-refresh/gcodes.hpp
@@ -1440,4 +1440,29 @@ struct CloseLid {
     }
 };
 
+struct DeactivateAll {
+    using ParseResult = std::optional<DeactivateAll>;
+    static constexpr auto prefix = std::array{'M', '1', '8'};
+    static constexpr const char* response = "M18 OK\n";
+
+    template <typename InputIt, typename Limit>
+    requires std::forward_iterator<InputIt> &&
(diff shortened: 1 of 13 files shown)


sfoster1 (approved):
Nice. Why sequence the internal messages and not just send them both at the same time?

fsinapi:
> Nice. Why sequence the internal messages and not just send them both at the same time?

Mostly to know when to send the response. In theory there's probably no harm in sending the response as soon as the message is received since the individual steps per task are 1) instantaneous and 2) shouldn't ever fail, but tbh I like that it would be very easy to add error handling into this setup if needed.

Sending them at the same time and maintaining a counter of responses could work too, but this method felt easier to fudge into the `AckCache` setup.

sfoster1:
> > Nice. Why sequence the internal messages and not just send them both at the same time?
> 
> Mostly to know when to send the response. In theory there's probably no harm in sending the response as soon as the message is received since the individual steps per task are 1) instantaneous and 2) shouldn't ever fail, but tbh I like that it would be very easy to add error handling into this setup if needed.
> 
> Sending them at the same time and maintaining a counter of responses could work too, but this method felt easier to fudge into the `AckCache` setup.

Makes sense!

PR #271, opened 2022-03-23
feat(thermocycler-refresh): add simulated temperature

fsinapi:
### Summary

Adds a new thread to the simulator target that acts as the time base for the simulator. Any data or behavior that depends on timing will be generated by this thread in order to keep the timebase consistent across the simulation.

This PR only adds periodic temperature data. There is a _really_ simplistic simulation of the temperature response from the peltiers and lid heater with two basic parts:
- Simple integrator with initial conditions at 23ºC and a gain constant derived from the PID tuning in #256 
- Small pull back to ambient temperature that scales with the difference from ambient temperature - i.e. `(ambient temp - current temp) * gain`. The gain constant is scaled to try to approximate the heat loss of the lid heater when it's turned off at 100ºC.

This also adds a command line option for the simulator, `--realtime`. If specified, the timing of everything will be based on the system clock to try to approximate the time it takes on a real thermocycler. If this flag is not specified, the periodic data will be generated much faster than real time.

The ticks for the motor drivers aren't simulated yet, so motor commands will still never return in the simulator.

### DIFF
--- a/stm32-modules/thermocycler-refresh/simulator/lid_heater_thread.cpp
+++ b/stm32-modules/thermocycler-refresh/simulator/lid_heater_thread.cpp
@@ -12,10 +12,16 @@ using namespace lid_heater_thread;
 class SimLidHeaterPolicy {
   private:
     double _power = 0.0F;
+    periodic_data_thread::PeriodicDataThread *_periodic_data;
 
   public:
+    SimLidHeaterPolicy(periodic_data_thread::PeriodicDataThread *periodic_data)
+        : _periodic_data(periodic_data) {}
+
     auto set_heater_power(double power) -> bool {
         _power = std::clamp(power, (double)0.0F, (double)1.0F);
+        _periodic_data->send_message(periodic_data_thread::PeriodicDataMessage(
+            periodic_data_thread::HeatPadPower{.power = _power}));
         return true;
     }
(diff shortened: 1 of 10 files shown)


stm32-modules/thermocycler-refresh/simulator/lid_heater_thread.cpp
```diff
@@ -12,10 +12,16 @@ using namespace lid_heater_thread;
 class SimLidHeaterPolicy {
   private:
     double _power = 0.0F;
+    periodic_data_thread::PeriodicDataThread *_periodic_data;
```
sfoster1: Is there a way we can get away from raw pointers here? It doesn't really give us that much since these objects have equivalent lifetime but it might save us some tears in the future if we start segfaulting while shutting down or something. I guess they might have to be weak_ptrs since I think there's circular refs here though.

sfoster1 (approved):
This looks awesome, extremely nice.

stm32-modules/thermocycler-refresh/simulator/lid_heater_thread.cpp
fsinapi: Yep I can shift things around to get a `shared_ptr` in here

PR #273, opened 2022-03-24
feat(thermocycler-refresh): simple ICT fixture commands

fsinapi:
Adds some commands intended for the ICT fixture for the TCR PCB.

- M242.D gets a new field, the standstill indicator bit from the TMC2130. This lets the ICT fixture verify that communication with the TMC is working.
- M900 GetBoardRevision checks the board revision. The board revision is also updated to be a 1-indexed enum so that the number actually matches what's printed on the board.
- M901 GetLidSwitches reads the raw value of the lid switches (open & closed). The original TC reports raw switch values in the GetLidStatus command, but the new TC cannot use the raw values to determine the lid position so an extra command is needed.
- M902 GetFrontButton reads the state of the front button.

Tested on a Rev2 board by sending all of the new commands through a terminal.

### DIFF
--- a/stm32-modules/common/STM32F303/STM32F303RETx_FLASH.ld
+++ b/stm32-modules/common/STM32F303/STM32F303RETx_FLASH.ld
@@ -42,7 +42,7 @@ _Min_Stack_Size = 0x400; /* required amount of stack */
 /* Specify the memory areas */
 MEMORY
 {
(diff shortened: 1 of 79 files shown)


sfoster1 (approved):
Looks good to me, though I wouldn't mind seeing the more debug-y commands like M901 and M902 get a `.D` suffix even if there's no non-debug command to disambiguate

PR #274, opened 2022-03-25
fix(heater-shaker): eliminate bad message acknowledgements

pmoegenburg:
Fixes for bad message acknowledgement error that's been popping up during testing. Error seems to be caused by hot-to-touch SetLEDMessage responding unnecessarily and without an originating id to the host comms task. The fix responds to host comms task with ErrorMessage only when necessary. Needs to be tested on latest built module.


sfoster1 (approved):
Looks good but oof we gotta fix this systematically at some point, adding an extra bool to all these messages separately is so annoying

PR #275, opened 2022-03-25
fix(thermocycler-refresh): use non-malloc snprintf implementation

fsinapi:
### Summary

The standard library implementation of `snprintf` being used was dynamically allocating memory when writing floating point numbers. This wasn't necessarily causing any issues, but general knowledge online seems to suggest that _ever_ calling `malloc` may cause errors when using FeeRTOS. In the spirit of mitigating that possibility, this PR allows us to keep using `snprintf` without worrying about memory allocation.

- Tested that the library is being used on Host by dropping the `PRINTF_MAX_FLOAT` constant to `1e1` and watching every test with a floating point print fail
- Tested that the library is being used on Cross by putting a breakpoint on malloc. We used to get a break when printing a float, now we don't.

There's still _something_ calling `malloc` on startup in `__libc_init_array`... but this gets rid of the only offender I could find during normal runtime.


sfoster1 (approved):
Awesome! I think you can find that last thing by running under a debugger and doing `break malloc` before starting the program

PR #276, opened 2022-03-28
feat(heater-shaker): added setpoint range bounds

pmoegenburg:
The current firmware does not have any range bounds on setpoint input from the SetTemperature gcode command. This places minimum and maximum limits of 0 degrees and 100 degrees Celsius, respectively, on the setpoint input.


sfoster1 (approved):
Looks good but let's get more of a pr comment

PR #280, opened 2022-04-06
feat(thermocycler-gen2): add startup application

fsinapi:
### Summary

Adds a startup application to the TC Gen2 that lives in the first sector of the MCU flash memory. On a hardware reset, the vector table will jump to this startup application's reset vector, which will either 1) jump to the application if it is valid or 2) jump to the built-in bootloader if there doesn't seem to be a valid application. Additionally, the startup application provides an exception handler that will catch any exceptions if the actual application is somehow faulty and can't load its own vector table, and this exception handler will also jump to the bootloader.

- Added module-startup target, which consists of a single assembly file
- Added a `thermocycler-gen2-image-hex` target that uses the same script as the OT3 repository to combine two hex files into a single, full image.
- Moved the thermocycler gen2 linker + vector table to be offset to the second page.

Tested on a Thermocycler board by:
- Erasing the MCU and then uploading _just_ the startup app, and confirming the MCU jumps to the bootloader
- Uploading the application firmware and confirming the startup app jumps to it (stepped through with STM32CubeProgrammer CPU debugger)
- Erasing the MCU and then uploading the combined image and confirming everything works correctly

Some more work will be needed to set this up for the Heater-Shaker, the bootloader address is slightly different so it will have to be defined on a per-target basis... but this works on the STM32G491 for now.


stm32-modules/common/STM32G491/STM32G491VETx_FLASH.ld
```diff
@@ -58,12 +58,18 @@ _estack = ORIGIN(RAM) + LENGTH(RAM);	/* end of "RAM" Ram type memory */
 _Min_Heap_Size = 0x800;	/* required amount of heap  */
 _Min_Stack_Size = 0x600;	/* required amount of stack */
 
+_flash_offset = 0x800; /* first page is occupied by bootloader shim */
+_flash_start = 0x8000000;
+_app_flash_start  = _flash_start + _flash_offset;
+_flash_size   = 508K; /* last page is occupied by device persistent memory.
```
sfoster1: i'm pretty sure you could do 
```
_flash_size = 512K;
_serial_size = 0x800;
_app_flash_size = _flash_size - _flash_offset - _serial_size;
```
i think it'll convert the units

stm32-modules/common/module-startup/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+    "LINKER:--gc-sections")
+
+find_program(CROSS_OBJCOPY "${CrossGCC_TRIPLE}-objcopy"
+    PATHS "${CrossGCC_BINDIR}"
+    NO_DEFAULT_PATH
+    REQUIRED)
+
+# Incurs at least a relink when you change the linker file (and a recompile of main
+# but hopefully that's quick)
+set_source_files_properties(./module-startup.s
```
sfoster1:
```suggestion
set_source_files_properties(./module-startup.S
```

stm32-modules/common/module-startup/STM32_MODULE_STARTUP.ld
```diff
@@ -0,0 +1,127 @@
...
+    . = ALIGN(4);
+  } >FLASH
+
+  /* The program code and other data goes into FLASH */
+  .text :
+  {
+    . = ALIGN(4);
+    *(.text)           /* .text sections (code) */
+    *(.text*)          /* .text* sections (code) */
+    *(.glue_7)         /* glue arm to thumb code */
```
sfoster1: it's more messing around than I think is necessarily needed, but I'm like 99% sure that we don't need anything beyond `*(.text)` and `*(.text*)`. we're not using arm/thumb interwork trampolines because those are a c thing; `eh_frame` is exceptions, c/c++; global offset table is c++; `init` and `fini` are c. `rodata` isn't really a thing here because we're just using immediates or values packed into the executable, not big object initializers.

again, this is kind of more messing around than perhaps needed since the code that goes in there won't be generated anyway, but still

stm32-modules/common/module-startup/module-startup.S
```diff
@@ -0,0 +1,186 @@
...
+    /* Jump to the user application */
+    ldr     r0, =_user_app_start
+    ldr     r1, [r0]
+    /* As an easy way to check for a missing program, check if the reset
+     * vector we're expecting points to 0xFFFFFFFx. If that's the case,
+     * there is definitely not a valid program to run. */
+    ldr     r2, =0xFFFFFFF0
+    and     r1, r1, r2
+    cmp     r1, r2 
+    bcs Post_Fault_Handler /* jump if upper 7 nibbles are all 0xF */
```
sfoster1, line 50: `cmp` does `r1-r2`. a valid program means `r1`<<`r2`, so the result will be very negative, with `N`, `V`, and... `C` (i think) set. Probably a good idea to verify with a debugger.

An invalid program probably means that `r1`>`r2` but not by much (e.g. this is erased flash, `0xFFFFFFFF` so the result of `CMP` is `0xF`) which leaves all flags clear. 

Basically, I've talked myself into being a bit confused about why this work; from my understanding of `cmp` as an alias of `sub` it kind of seems like this snippet does the opposite of what you want, where if the program is valid you'll always go to `Post_Fault_Handler`. Shouldn't you be doing `ble r0` here? Since you've tested it I assume I'm wrong, but maybe a more in-depth comment is in order.

stm32-modules/common/module-startup/module-startup.S
```diff
@@ -0,0 +1,186 @@
...
+    ldr     r0, =0xE000ED28 /* address of CFSR */
+    ldr     r1, [r0]
+    str     r1, [r0]        /* writing the reg value back clears 
+                             * any asserted bits */
+    /* Now clear HFSR register */
+    ldr     r0, =0xE000ED2C /* address of HFSR */
+    ldr     r1, [r0]
+    str     r1, [r0]        /* writing the reg value back clears 
+                             * any asserted bits */
+    /* Now ovewrite PC in the stack frame, stored at SP + 0x18 */
```
sfoster1: this probably wants a link out or section ref to the cortex docs on exception stacking

stm32-modules/common/module-startup/module-startup.S
```diff
@@ -0,0 +1,186 @@
...
+
+InitNvicClear:
+    /* Clear out the NVIC enable and pending bits */
+    ldr     r0, =0xE000E180UL  /* NVIC Enable base address */
+    ldr     r1, =0xE000E280UL  /* NVIC Pending base address */
+    ldr     r2, =0xFFFFFFFF    /* Constant */
+    ldr     r3, =0             /* Loop iterator */
+    b LoopNvicClear
+NvicClear:
+    str     r2, [r0,r3] /* Clear Enable at iterataor position */
```
sfoster1:
```suggestion
    str     r2, [r0,r3] /* Clear Enable at iterator position */
```

stm32-modules/common/module-startup/module-startup.S
```diff
@@ -0,0 +1,186 @@
...
+    str     r2, [r1,r3] /* Clear Pending at iterataor position */
```
sfoster1:
```suggestion
    str     r2, [r1,r3] /* Clear Pending at iterator position */
```

stm32-modules/common/module-startup/module-startup.S
```diff
@@ -0,0 +1,186 @@
...
+ *         This sends the processor to the built in bootloader.
+ *
+ * @param  None
+ * @retval : None
+*/
+    .section	.text.Post_Fault_Handler,"ax",%progbits
+    .thumb_func
+Post_Fault_Handler:
+    /* We make the assumption that the RCC is already not initialized.*/
+SystickDisable:
```
sfoster1, line 109: we probably want to have a critical section for basically this entire thing, `cpsid ai` here and `cpsie ai` right before the bootloader entry jump

sfoster1 (changes requested):
This is awesome. Some inline comments on some of the assembler stuff, but overall looks good.

Do we want to augment this eventually with fault handlers in the main programs dropping back here via system reset or something?

fsinapi:
> Do we want to augment this eventually with fault handlers in the main programs dropping back here via system reset or something?

@sfoster1 I had originally been thinking of having the fault handlers just use the `system_hardware_enter_bootloader` we've got, but I think that jumping back to here would work fine. The C code will probably have to deinit the RCC, since reimplementing `HAL_RCC_DeInit()` would be pretty rough.

sfoster1:
> > Do we want to augment this eventually with fault handlers in the main programs dropping back here via system reset or something?
> 
> @sfoster1 I had originally been thinking of having the fault handlers just use the `system_hardware_enter_bootloader` we've got, but I think that jumping back to here would work fine. The C code will probably have to deinit the RCC, since reimplementing `HAL_RCC_DeInit()` would be pretty rough.

Your way sounds a lot better honestly, it'll be super annoying to chain backwards

stm32-modules/common/module-startup/module-startup.S
fsinapi: Your logic is 100% right, but this works because the Carry Flag is inverted for `subs` operations. https://developer.arm.com/documentation/dui0801/a/CIADCDHH

sfoster1: Oh how I love digital logic

fsinapi:
> > > Do we want to augment this eventually with fault handlers in the main programs dropping back here via system reset or something?
> > 
> > 
> > @sfoster1 I had originally been thinking of having the fault handlers just use the `system_hardware_enter_bootloader` we've got, but I think that jumping back to here would work fine. The C code will probably have to deinit the RCC, since reimplementing `HAL_RCC_DeInit()` would be pretty rough.
> 
> Your way sounds a lot better honestly, it'll be super annoying to chain backwards

I thought about it some more and it's going to be annoying no matter what, because just like in `module-startup.S` we're going to have to mess with the exception frame to change the cached PC value to jump to our recovery function, otherwise the MCU is going to jump to the bootloader with a hard fault flag. Lots of fun :)

sfoster1 (approved):
Looks good to me, sorry about the bad clear interrupt suggestion

PR #281, opened 2022-04-07
feat(thermocycler-gen2): add socket builder for python utils

fsinapi:
Added this for the TC2 Simulator demo, but it's also quite useful for testing the simulator through python.


sfoster1 (approved):
nice

PR #284, opened 2022-04-08
fix(heater-shaker): eliminated unneeded plate latch state

pmoegenburg:
UNKNOWN was not being used. UNKNOWN_IDLE captures all possible states.


sfoster1 (approved):
nice

PR #285, opened 2022-04-11
fix(thermocycler-gen2): check for empty serial 

fsinapi:
### Summary

When reporting the serial in response to M115, the firmware used to happily send the contents of an empty flash page (`0xFF`) which is an invalid character. When trying to decode the serial number, the Python side would rightfully throw an error about invalid characters. This PR ensures that invalid ASCII characters don't end up making their way to the host.


sfoster1 (changes requested):
I don't think this is the right place to fix it, or at least not in this way, because it relies on a NUL, which is usually a bad sign.

I think there's three other reasonable approaches here which would be better:
1. Refactor this message to contain the size of the serial number, and send the size before the invalid byte
or
2. Don't do the check in the system task; rather, do it in the comms task since this is a comms problem
or
3. When we decode this value in python, the actual problem that throws the error is that we're decoding this UTF-8. That's a decision python makes, and a decision we don't want to deal with, since we don't want to know about UTF-8. Python should be able to safely handle arbitrary data and make decisions about validity on its own.

sfoster1 (approved):

PR #286, opened 2022-04-12
feat(heater-shaker): add deactivate heater gcode

pmoegenburg:
Currently, the hardware controller class `deactivate module` command issues a `set_temperature(0)` gcode, which sets setpoint to 0. This gcode will now be used, which disables heatpad power output. This tested successfully on my at-home frankenstein module. An associated hardware controller class PR is in the works.


sfoster1 (changes requested):
Let's add a test for this but otherwise looks good!

sfoster1 (approved):
Nice, looks good to me!

PR #287, opened 2022-04-15
fix(heater-shaker): update setpoint with deactivate_heater

pmoegenburg:
Need to set setpoint to 0 so hardware control class can properly utilize the setpoint to identify heater state.


sfoster1 (changes requested):
Tests please

sfoster1 (changes requested):
The functionality looks good, really love the use of optional, but we really need to add some tests for this.

sfoster1 (approved):
Nice!

PR #288, opened 2022-04-20
feat(heater-shaker): add safe startup app

fsinapi:
### Summary
* Added Startup App to the heater shaker, with new target `heater-shaker-image-hex`
* Startup app is made generic to run on either F303 or G491
* Heater shaker is compiled to start at Page 1 instead of Page 0
* Heater shaker sets VTable pointer to the correct location
* Full image hex files regenerate properly

### Review Requests

@pmoegenburg Can you test that this works correctly on the Heater Shaker board? Tests can be as follows:

- [x] Compile the target `heater-shaker-image-hex`
- [x] Erase the full flash memory on a Heater Shaker and then program it with the `STM32F303-startup.hex` file
- [x] Restart the board and confirm that you can connect to the USB DFU bootloader
- [x] Upload the file `heater-shaker.hex` through DFU, restart the board, and confirm the main firmware app starts as normal
- [x] Erase the board again with an STLink, upload the file `heater-shaker-image.hex`, restart the board, and confirm the main firmware app starts as normal


sfoster1 (approved):
Looks right to me.

pmoegenburg (approved):

PR #289, opened 2022-04-28
feat(thermocycler-gen2): automatic seal control

fsinapi:
### Summary

Adds the seal motor to the overall lid open/close state machine. The seal is retracted before all lid movements, and then is engaged at the end of every Lid Close action.

- The seal is extended by simply moving it a preset number of steps. A limit switch may take precedence in future firmware.
- The default settings configured to the TMC2130 result in the seal stalling out when retracting, hence the larger number of steps for retraction movements.
- If the seal has not been moved since system startup, it is _homed_ before the lid can be opened or closed. Homing is accomplished by extending the seal a small amount, and then retracting it to a stall. The small extension is to ensure the seal stall is detected immediately, since the TMC2130 only detects stalls above a certain velocity. 
- If the lid is closed on startup, the firmware assumes the seal is probably engaged and thus the seal skips the small extension. If the seal _isn't_ engaged then this can result in an annoying, but harmless, sound as the motor stalls out against the hardstop.

Seal motor drive values were tested with some cycle testing before hard coding into the firmware. Overall Motor Task changes were tested extensively on a prototype unit to check each starting condition of the state machine. Lid behaviors work correctly both through Gcode and by pressing the front switch.


sfoster1 (approved):
Awesome. Love the mermaid diagrams

PR #291, opened 2022-05-02
feat(thermocycler-gen2): split Plate Lift into its own command

fsinapi:
### Summary

Adds M128, a gcode dedicated to activating the plate lift mechanism on the Thermocycler Gen2. Also removes the plate lift portion from the lid open command, and switches the lid opening logic for the EVT placement of the Lid Open switch (it is now at 90º instead of the plate-lift position).

Tested on a Gen1 unit by manually blocking/unblocking a limit switch. The actual numeric distances will be tuned in a later PR once an EVT unit is available for testing.

### DIFF
--- a/stm32-modules/include/thermocycler-gen2/thermocycler-gen2/motor_task.hpp
+++ b/stm32-modules/include/thermocycler-gen2/thermocycler-gen2/motor_task.hpp
@@ -83,10 +83,10 @@ struct LidStepperState {
     // distance is 120 degrees which is far wider than the actual travel angle.
     constexpr static double FULL_OPEN_DEGREES =
         motor_util::LidStepper::angle_to_microsteps(120);
-    // After opening to the open switch, the lid must close a few degrees to
-    // be at the 90º position
-    constexpr static double OPEN_BACK_TO_90_DEGREES =
-        motor_util::LidStepper::angle_to_microsteps(-17);
+    // After opening to the open switch, the lid must open a few extra
+    // degrees to be fully seated in the open switch.
+    constexpr static double OPEN_OVERDRIVE_DEGREES =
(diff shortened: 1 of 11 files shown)


stm32-modules/include/thermocycler-gen2/thermocycler-gen2/motor_task.hpp
```diff
@@ -1009,13 +1098,13 @@ class MotorTask {
                 // Now that the lid is at the open position,
                 // the solenoid can be safely turned off
                 policy.lid_solenoid_disengage();
-                // Move to the Open Back To 90 step
+                // Overdrive into switch
                 policy.lid_stepper_start(
-                    LidStepperState::OPEN_BACK_TO_90_DEGREES, false);
+                    LidStepperState::OPEN_OVERDRIVE_DEGREES, true);
```
sfoster1, line 1103: this change is what removes the old open behavior of "always engage plate lift", right?

sfoster1 (approved):
Looks good I think - the only question is, does this PR also remove the old behavior of open that engages the plate lift every time?

stm32-modules/include/thermocycler-gen2/thermocycler-gen2/motor_task.hpp
fsinapi: Yep! Or sort of. On the older (1.4) prototypes, the limit switch is mounted far enough back that driving to it will put the lid at an angle much greater than 90º with the plate lift mechanism fully engaged. On the EVT units, the switch is at approximately 90º. So running the code before this PR on an EVT unit will result in the lid being open by like 70º and never activating the plate lift.

sfoster1 (approved):
Sounds good!

PR #292, opened 2022-05-05
feat(thermocycler-gen2): Seal motor limit switches

fsinapi:
### Summary

Adds support for limit switches for the seal motor. The two limit switches share a single falling edge interrupt due to PCB design. See `stm32-modules/thermocycler-gen2/firmware/motor_task/README.md` for more information on the scheme used.

- Added code to activate the correct input line as a falling interrupt and send a message to the motor task when the switch is triggered
- Modified the lid state machine to base movements off of the limit switches. When retracting or extending, the seal moves until one of the switches triggers, and then backs off by 0.5mm to leave the switch open while idle.
- Updated motor task tests to use a vector of conditions for each step in the lid state machine in order to simplify the tests

Tested that the seal switch interrupt works correctly and terminates seal movements, but without the correct switches I had to short the line to ground manually to test this. Was able to dry run the lid state machine stuff this way, but I am waiting on a unit to arrive in order to test with the actual mechanical assembly. A later PR will adjust the distances if necessary.

### DIFF
--- a/stm32-modules/thermocycler-gen2/firmware/motor_task/motor_hardware.c
+++ b/stm32-modules/thermocycler-gen2/firmware/motor_task/motor_hardware.c
@@ -45,6 +45,13 @@ extern "C" {
 /** Pin for the lid open optical switch.*/
 #define LID_OPEN_SWITCH_PIN (GPIO_PIN_7)
 
+/** Port for Seal Limit Switch input.*/
+#define SEAL_SWITCH_PORT (GPIOD)
+/** Pin for Seal Limit Switch input.*/
+#define SEAL_SWITCH_PIN  (GPIO_PIN_6)
+/** IRQ for the seal limit switch.*/
+#define SEAL_STEPPER_DIAG1_IRQ (EXTI4_IRQn)
+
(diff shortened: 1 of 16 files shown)


stm32-modules/thermocycler-gen2/firmware/motor_task/motor_hardware.c
```diff
@@ -110,6 +117,9 @@ typedef struct seal_hardware_struct {
     // Current direction of the seal stepper.
     // True = forwards, False = backwards
     bool direction;
+    // Bool check for whether the next switch interrupt
+    // should trigger a callback
+    bool limit_switch_armed;
```
sfoster1: does this want to be a c atomic?

sfoster1 (approved):
Looks good overall but does that state bool need to be atomic?

stm32-modules/thermocycler-gen2/firmware/motor_task/motor_hardware.c
fsinapi: Yes indeed it does!

PR #293, opened 2022-05-06
chore(thermocycler-gen2): use gen1 PID and VID for EVT

fsinapi:
Updating the robot to recognize the Gen2 USB PID/VID will include a buildroot update. For now we will use the the values from the Gen1 board so a buildroot update won't be required for EVT testing. Once the buildroot changes are applied + tested + rolled out, the values can revert to the Gen2 ones.

Verified that the PID/VID is updated as expected in firmware and seen by the host computer.


sfoster1 (approved):
Blargh

PR #294, opened 2022-05-09
chore(thermocycler-gen2): Fixed cmake lint settings

fsinapi:
`clang-tidy` wasn't actually linting the `thermocycler-gen2-core` target. This PR fixes that, and also adds fallthrough checking to `gcc` because clang-tidy doesn't support it as a direct check and the workaround we were using doesn't seem to work correctly.

- Adjusted cmake linter settings to properly lint `core` sources
- Added -Wimplicit-fallthrough to gcc options because clang-tidy was not picking it up.
- Fixed some linting errors that were not being detected before


sfoster1 (approved):
hooray

PR #295, opened 2022-05-10
feat(thermocycler-gen2): seal switch safety checks

fsinapi:
Adds some safety checks for the seal motors. Attempting to open or close the lid while the seal is already triggering one of the two (indistinguishable) endstop switches will result in an error. Future versions of the PCB will be able to account for this, but on Rev2 it is important to not risk damaging the switches since we can't distinguish whether a movement will result in driving the switch into a hardstop.

- Addressed a bug where the response ID isn't cleared after Debug Seal/Lid commands
- Added the seal switch status to the M901.D response for debugging purposes
- Adjusted distances for lid hinge based on testing with EVT unit

Tested on an EVT unit


sfoster1 (approved):
Nice

PR #296, opened 2022-05-10
fix(heater-shaker): fix message passing for bootloader

pmoegenburg:
No plate-latch-closed detection was preventing H-S from running the enter-bootloader process. This fix also adds clearing the NVIC prior to entering the bootloader.

### DIFF
--- a/stm32-modules/include/heater-shaker/heater-shaker/system_task.hpp
+++ b/stm32-modules/include/heater-shaker/heater-shaker/system_task.hpp
@@ -110,7 +110,7 @@ class SystemTask {
         }
 
         auto cool_message = messages::SetTemperatureMessage{
-            .id = 1, .target_temperature = 0, .from_system = true};
+            .id = 0, .target_temperature = 0, .from_system = true};
         auto cool_id = prep_cache.add(cool_message);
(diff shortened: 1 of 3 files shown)


stm32-modules/include/heater-shaker/heater-shaker/system_task.hpp
```diff
@@ -110,15 +110,15 @@ class SystemTask {
         }
 
         auto cool_message = messages::SetTemperatureMessage{
-            .id = 0, .target_temperature = 0, .from_system = true};
+            .id = 1, .target_temperature = 0, .from_system = true};
```
fsinapi: Do these `id` values need to be nonzero before adding to the cache? If not, I think that the intention is a bit clearer if they're left as 0. Doesn't really matter either way tbh.

fsinapi (approved):
Nice catch, this looks good to me! One suggestion but it looks good to go as-is.

stm32-modules/include/heater-shaker/heater-shaker/system_task.hpp
pmoegenburg: Changed for debugging, but can change back

PR #297, opened 2022-05-10
feat(heater-shaker): added combo binary target

pmoegenburg:
Combined binary (startup + application) needed for dfu process


sfoster1 (changes requested):
We should build this in the github action also, but this looks good

sfoster1 (approved):
Nice! At some point we're going to want those files getting generated and pushed to storage somewhere but we're good for now.

PR #299, opened 2022-05-11
feat(thermocycler-gen2): add volumetric overshoot

fsinapi:
Adds support for volumetric overshoot to the Thermocycler Gen2. This copies the behavior from the TC Gen1, including the coefficients for calculating the offset.

- Added a readme for the thermal subsystem to explain some of the less straightforward parts of the plate control
- Added overshoot to the `Platecontrol` class. The full state machine of it is shown in the readme.
- Added tests for the overshoot/undershoot calculations, and updated tests for the `PlateControl` class

Tested in the simulator, as well as on an EVT unit by setting the volume very high (400 uL) and moving to a few temperatures. The software targets the overshoot/undershoot temperature for 10 seconds as intended, and then moves to the real target.

### DIFF
--- a/stm32-modules/thermocycler-gen2/src/plate_control.cpp
+++ b/stm32-modules/thermocycler-gen2/src/plate_control.cpp
@@ -20,7 +20,7 @@ auto PlateControl::update_control(Seconds time) -> UpdateRet {
     switch (_status) {
         case PlateStatus::INITIAL_HEAT:
             // Check if we crossed the TRUE threshold temp
-            if (plate_temp() > _setpoint) {
+            if (plate_temp() >= _setpoint) {
                 _status = PlateStatus::OVERSHOOT;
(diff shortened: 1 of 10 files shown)


stm32-modules/thermocycler-gen2/src/plate_control.cpp
```diff
@@ -98,20 +127,32 @@ auto PlateControl::set_new_target(double setpoint, double hold_time,
     return temp * IDLE_FAN_POWER_SLOPE;
 }
 
+[[nodiscard]] auto PlateControl::calculate_overshoot(double setpoint,
```
sfoster1: these can be `const` and maybe even `static` since it looks like they just use static member values and arguments

sfoster1 (approved):
Looks good; could make the new over/undershoot functions const but that's about it.

PR #300, opened 2022-05-12
fix(thermocycler-gen2): fixed LED power issues

fsinapi:
Erroneous LED behavior was noticed on EVT units, most notably blinking lights while the seal motor was being driven. This issue was not seen in units being run in the NY office. Reviewing the code + schematic revealed that the LED Enable Pin was never being set. In theory the LED's never should have worked without this pin being set, but in practice they just worked at a reduced capacity.

After enabling this pin, the LED performance is more in-line with what is expected. Even on units that did not show erratic blinking, the colors are more representative of what's expected (in particular, the orange had been very red-ish and is now clearly orange).

This PR also increases the PWM frequency from 400kHz to approximately 800kHz. The latter showed erratic performance in the past, but testing with the power pin enabled indicates that this frequency works exactly as expected now.

### DIFF
--- a/stm32-modules/thermocycler-gen2/firmware/system/system_led_hardware.c
+++ b/stm32-modules/thermocycler-gen2/firmware/system/system_led_hardware.c
@@ -21,14 +21,17 @@
 // ----------------------------------------------------------------------------
 // LOCAL DEFINITIONS
 
-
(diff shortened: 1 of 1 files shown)


sfoster1 (approved):
:electron:

PR #302, opened 2022-05-13
feat(thermocycler-gen2): evaporation test script

fsinapi:
Adds a testing script for evaporation testing on EVT units. This script copies the current protocol in use for evaporation testing, but removes the need for an OT2 to control the thermocycler.

The script is fully contained, using only standard Python libraries (with the exception of the Serial library), in order to provide portability outside of this repository.

The script has been tested running for low cycle counts (1 and 2) on an EVT unit.


sfoster1 (approved):
Nice, love it.

PR #304, opened 2022-05-17
chore(thermocycler-gen2): motor speed tuning

fsinapi:
A few small changes to the motor behavior on the system
- Lid motor velocity is increased from 75rpm to 125rpm for smoother performance and shorter open/close times
- Lid motor hold current is increased from 0 to a very small current to eliminate motor driver hum when holding
- Seal motor velocity is bumped from 90k to 200k steps/second. With the acceleration curve it still ends up quite smooth, and dramatically decreases the time to open/close.
- During plate lift movements, the lid closes a little less to avoid potential crashes.

All changes tested extensively on EVT hardware, and velocities were agreed on with hardware engineering.


sfoster1 (approved):
Nice

PR #305, opened 2022-05-17
feat(thermocycler-gen2): thermal calibration and tuning

fsinapi:
### Summary

- Added in a new calibration offset for the thermal system which takes the current heat sink temperature into account when calculating the offset from thermistor to plate, copied from the Thermocycler Gen1.
- Modified default PID values for peltiers to be less aggressive in order to reduce oscillation
- Added default thermal calibration values based off of initial population testing on EVT units

Tested on an EVT unit. The offset values were verified by testing the steady-state temperature in the well plates using a multiprobe tool.

### DIFF
--- a/stm32-modules/include/thermocycler-gen2/thermocycler-gen2/eeprom.hpp
+++ b/stm32-modules/include/thermocycler-gen2/thermocycler-gen2/eeprom.hpp
@@ -32,7 +32,7 @@ namespace eeprom {
  *
  */
 struct OffsetConstants {
-    // The value of the constants B and C
+    // The value of the constants A, B and C
     double a, b, c;
 };
 
@@ -49,8 +49,10 @@ class Eeprom {
      * @brief Get the offset constants from the EEPROM
      *
      * @tparam Policy for reading from EEPROM
+     * @param defaults OffsetConstants containing default values to return
+     *                 in the case that the EEPROM is not written.
      * @param policy Instance of Policy
-     * @return OffsetConstants containing the B and C constants, or the
+     * @return OffsetConstants containing the A, B and C constants, or the
      * default values if the EEPROM doesn't have programmed values.
      */
     template <at24c0xc::AT24C0xC_Policy Policy>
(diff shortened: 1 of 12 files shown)


stm32-modules/include/thermocycler-gen2/thermocycler-gen2/eeprom.hpp
```diff
@@ -23,7 +23,7 @@ namespace eeprom {
  * readings. Using two constants, B and C (for legacy purposes), the
```
sfoster1, line 23: add an explanation about A?

stm32-modules/include/thermocycler-gen2/thermocycler-gen2/eeprom.hpp
```diff
@@ -118,12 +130,17 @@ class Eeprom {
...
+        // See \ref EEPROMFlag
+        CONST_FLAG = 2,
+        CONST_A = 3,  // Value of the A constant
     };
 
     // Enumeration of the EEPROM_CONST_FLAG values
-    enum class EEPROMFlag { WRITTEN_NO_CHECKSUM = 1, INVALID = 0xFF };
+    enum class EEPROMFlag {
+        WRITTEN_NO_A = 1,    // Values of B and C are written
+        WRITTEN_WITH_A = 2,  // Values of all constants are written
```
sfoster1: you are a hero for trying to keep this backward compatible across proto versions but maybe we should also consider just considering eeproms without A invalid?

stm32-modules/include/thermocycler-gen2/thermocycler-gen2/thermal_plate_task.hpp
```diff
@@ -661,7 +672,9 @@ class ThermalPlateTask {
 
     auto handle_temperature_conversion(uint16_t conversion_result,
                                        Thermistor& thermistor,
-                                       bool apply_offset) -> void {
+                                       bool apply_offset,
+                                       const double heatsink_temp = 0.0F)
```
sfoster1: [nit] I personally don't love const value params

sfoster1 (approved):
Looks good but we're gonna want to update some comments.

stm32-modules/include/thermocycler-gen2/thermocycler-gen2/thermal_plate_task.hpp
fsinapi: image

PR #306, opened 2022-05-19
feat(thermocycler-gen2): thermistor drift check

fsinapi:
### Summary

Adds the thermistor drift error from TC Gen1. Similar to the Gen1 firmware, this error is persistent until the device resets. A "clear error" command may be appropriate in a future PR.

This has not been tested on hardware yet, and should not be merged until that testing can occur.

EDIT: Tested on hardware by unplugging a thermistor after reaching temperature. After doing this, any attempt to set plate temperature returned an `ERR408` message.


stm32-modules/include/thermocycler-gen2/thermocycler-gen2/plate_control.hpp
```diff
@@ -159,6 +163,21 @@ class PlateControl {
      */
     [[nodiscard]] auto temp_within_setpoint() const -> bool;
 
+    /**
+     * @brief Check for thermistor drift.
+     * @return true if the thermistors are \b within spec, false if the
```
sfoster1, line 168: `\b` ?

sfoster1 (approved):
Looks good to me, though like you say we should test it.

Another thing to think about - should this be latching until power cycle? I know that's what the gen1 does but that doesn't mean it's the right thing.

stm32-modules/include/thermocycler-gen2/thermocycler-gen2/plate_control.hpp
fsinapi: I like using [Doxygen formatting](https://www.doxygen.nl/manual/commands.html#cmdb)

fsinapi:
> Looks good to me, though like you say we should test it.
> 
> Another thing to think about - should this be latching until power cycle? I know that's what the gen1 does but that doesn't mean it's the right thing.

Per outside discussion, the plan will be to add a command that will clear this error.

PR #307, opened 2022-05-19
feat(thermocycler-gen2): simulate motors

fsinapi:
Adds simple simulation of the lid & seal motors on the thermocycler gen2. The lid is more involved just because most of the seal logic is already handled by the task itself.

Tested on the simulator by sending the commands to open/close the lid, and checking the lid & seal status via gcode after each action. Also tested sending the commands to move the lid by microsteps and confirmed the logic limits the lid appropriately.


sfoster1 (approved):
Nice!

PR #308, opened 2022-05-20
chore(tempdeck-gen3): create project

fsinapi:
Adds the `tempdeck-gen3` project!

Currently compiles useless placeholder executables for all targets, but this PR adds in the skeleton of the project structure and adds the tempdeck to github actions for all future development.

In particular, the usb firmware code is included because it's necessary to compile the HAL. This has been entirely copy+pasted from the Thermocycler but it isn't active at the moment.


stm32-modules/tempdeck-gen3/firmware/host_comms_task/usbd_desc.c
```diff
@@ -0,0 +1,281 @@
...
+  */
+
+/* Includes ------------------------------------------------------------------*/
+#include "usbd_core.h"
+#include "usbd_desc.h"
+#include "usbd_conf.h"
+
+/* Private typedef -----------------------------------------------------------*/
+/* Private define ------------------------------------------------------------*/
+// TODO(frank 05-06-2022): Uncomment these VID/PID definitions and replace
```
sfoster1, line 29: probably want to change these to something vaguely like the tempdeck - might forget otherwise.

sfoster1 (approved):
Looks awesome! Very excited.

Probably want to do a `grep -ri thermocycler ./tempdeck-gen3 ./include/tempdeck-gen3` and fix anything that pops up though - always better to get ahead of things in that regard.

fsinapi:
> Probably want to do a `grep -ri thermocycler ./tempdeck-gen3 ./include/tempdeck-gen3` and fix anything that pops up though - always better to get ahead of things in that regard.

Just USBD stuff popped up 👍

PR #309, opened 2022-05-25
feat(heater-shaker): create release with assets on tag push

pmoegenburg:
Creates a workflow to create a release for a tag upon pushing to that tag. Workflow also uploads assets to that release.


.github/workflows/heater-shaker-create-release.yaml
```diff
@@ -0,0 +1,109 @@
...
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ github.ref }}
+          release_name: ${{ github.ref }}
+          body: |
+            Changes in this Release
+            - change list
```
sfoster1: probably should either actually generate a change list, possibly even by subshelling out to `git log ./stm32-modules/heater-shaker`, or get rid of this

.github/workflows/heater-shaker-create-release.yaml
```diff
@@ -0,0 +1,109 @@
...
+        run: cmake --build ./build-stm32-cross --target heater-shaker-image-hex
+      - name: 'Build Combo Image Binary'
+        run: cmake --build ./build-stm32-cross --target heater-shaker-image-bin
+      - name: 'Create Release'
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ github.ref }}
```
sfoster1: looking at https://docs.github.com/en/actions/learn-github-actions/contexts#github-context I think you want `github.ref_name` here - `github.ref` will look like `refs/tags/tagname` while `github.ref_name` will just be `tagname`

.github/workflows/heater-shaker-create-release.yaml
```diff
@@ -0,0 +1,109 @@
...
+          release_name: ${{ github.ref }}
```
sfoster1: ditto

.github/workflows/heater-shaker-create-release.yaml
sfoster1: In fact I think you absolutely *must* fix this before trying to run it, otherwise
- You run this workflow on a tag, `heater-shaker@whatever`
- It creates a release with a tag called `refs/tags/heater-shaker@whatever`, with the full ref `refs/tags/refs/tags/heater-shaker@whatever`
- that's a new tag, so this action gets run again,
- you can see where this is going

sfoster1 (changes requested):
Definitely test this with some `workflow_dispatch`, even if that means there's spurious tags - gotta check it somehow, after all.

I think it's kind of redundant to do the lint, format, and host-build steps here; those should happen in other actions so this one can be slimmed down

sfoster1 (approved):
Nice! Let's give it a go

.github/workflows/heater-shaker-create-release.yaml
```diff
@@ -0,0 +1,91 @@
...
+      - if: github.event_name != 'pull_request'
+        name: 'Deploy'
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.S3_MODULES_DEPLOY_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.S3_MODULES_DEPLOY_SECRET }}
+          AWS_DEFAULT_REGION: us-east-2
+        run: |
+          . ./scripts/set_build_vars.sh
+          aws s3 sync --acl=public-read ${RELEASE_LOCAL_DIR} s3://opentrons-modules-builds/${RELEASE_UPLOAD_DIR}
+      # - name: 'Create Release'
```
sfoster1: let's remove the commented out sections

.github/workflows/heater-shaker-create-release.yaml
```diff
@@ -0,0 +1,91 @@
...
+      - name: 'Prep for install'
+        run: cmake --install ./build-stm32-cross
+      - if: github.event_name != 'pull_request'
+        name: 'Deploy'
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.S3_MODULES_DEPLOY_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.S3_MODULES_DEPLOY_SECRET }}
+          AWS_DEFAULT_REGION: us-east-2
+        run: |
+          . ./scripts/set_build_vars.sh
```
sfoster1, line 48: I think you'll have to edit https://github.com/Opentrons/opentrons-modules/blob/edge/scripts/set_build_vars.sh to have an entry for the heater-shaker to make sure that this actually goes to the right place in s3. you should be able to access it at `https://opentrons-modules-builds.s3.us-east-2.amazonaws.com/heater-shaker/$VERSION/filename.zip` after the build completes. i think this will upload nothing (because `RELEASE_LOCAL_DIR` isn't set because there's no `set_build_vars` entry for heater-shaker) to the root of the bucket (because `RELEASE_UPLOAD_DIR` isn't set)

sfoster1 (changes requested):
Almost there, just have to make sure real quick that this gets to the right place in aws.

sfoster1 (approved):
Looks good to me!

PR #310, opened 2022-05-25
feat(thermocycler-gen2): per-channel peltier tuning

fsinapi:
Initial thermal characterization testing on EVT units shows higher uniformity numbers than expected. Some estimated calculations based on early testing data suggest that tuning the devices with a different B & C coefficient for each channel will greatly improve the uniformity on outlier units. This PR adds the functionality to the firmware for having these separate coefficients.

- The thermal plate task has been modified to apply a different set of coefficients to the left/center/right channels, and the EEPROM driver is updated to store all 7 constants.
- The command to set the offset coefficients (M116) has been modified to accept a suffix with a channel to set, e.g. `M116.L` to set the left channel's values. The A value is always shared, however.

Tested on a simulator by _greatly_ changing the B and C constants per channel and verifying that the temperatures change on each side of the plate. Also tested on hardware that:
- The defaults are the same on each channel
- The updated values are kept over power cycles
- The updated values affect each channel independently

All while sitting at ambient temperature.

### DIFF
--- a/stm32-modules/include/thermocycler-gen2/thermocycler-gen2/eeprom.hpp
+++ b/stm32-modules/include/thermocycler-gen2/thermocycler-gen2/eeprom.hpp
@@ -32,8 +32,11 @@ namespace eeprom {
  *
  */
 struct OffsetConstants {
-    // The value of the constants A, B and C
-    double a, b, c;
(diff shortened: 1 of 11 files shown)


sfoster1 (approved):
Great!

PR #312, opened 2022-05-31
fix(thermocycler-gen2): system shutdown turns off thermal elements

fsinapi:
- deactivation messages have a new field to specify whether ack should be sent to system task instead of host task
- system task disables LED's before shutdown
- system task disables lid heater and thermistors before shutdown

Tested on simulator and on hardware.


sfoster1 (approved):
Nice. Love to shut down

PR #313, opened 2022-05-31
chore(thermocycler-gen2): Fan PWM at 1kHz

fsinapi:
The TC1 Arduino firmware uses `analogWrite` for fan PWM, which research shows should be around the magnitude of 1kHz. The TC2 has been  using the suggested 25kHz from the fan datasheet, but testing shows that reducing the PWM frequency to 1kHz will allow a lower slow-end fan velocity. A lower fan velocity is beneficial for uniformity at high target temperatures. 

Scope captures below show the fan velocity (captured from the tachometer pin) at 15% PWM driven at 25kHz and 1kHz respectively. 

image

image


sfoster1 (approved):
There's nothing worse than low speed BLDC control, whatever works

PR #314, opened 2022-06-02
fix(heater-shaker): M241 return semantics

pmoegenburg:
Previously returned `M241 STATE:xyz OK`. Changed to `M241 STATUS:xyz OK` to agree with module driver.

Opentrons/opentrons#10249


sfoster1 (approved):
Love it

laviera:
Thank you @pmoegenburg 😎

PR #315, opened 2022-06-02
feat(thermocycler-gen2): read fan tachometers

fsinapi:
Adds support to read the fan tachometers.

The tachometers use TIM4 channels 1 and 2 configured as Input Capture lines. TIM4 is set to One Pulse Mode and counts at 100kHz for 0.25 seconds. _This period was chosen based on the low end of the pulse rate being ~40Hz_. Whenever the timer is reset, DMA for each line is configured to capture the first 3 input capture readings into a buffer. When TIM4 overflows (and disables itself), the period for each tachometer is calculated based off of the latter 2 readings in its respective buffer. If there are not enough valid readings, the tachometer period is set to 0 to indicate that it is not moving. The timer is then restarted to refresh the reading.

The fan datasheet cites a max speed around 17,500 RPM. Testing on hardware shows expected scaling based on power (power of 30% gives around 5500 RPM).

- Simulator and test policy just scale the power by 17,500 to match the datasheet
- M103.D now reports the two tachometer readings. Currently there's no actual integration of tachometer speed into control.

### DIFF
--- a/stm32-modules/include/thermocycler-gen2/firmware/thermal_fan_hardware.h
+++ b/stm32-modules/include/thermocycler-gen2/firmware/thermal_fan_hardware.h
@@ -28,6 +28,15 @@ bool thermal_fan_set_power(double power);
  */
 double thermal_fan_get_power(void);
 
+double thermal_fan_get_tach_1_rpm(void);
+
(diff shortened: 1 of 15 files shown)


sfoster1 (approved):
Makes sense to me.

PR #316, opened 2022-06-06
feat(tempdeck-gen3): add queue aggregator

fsinapi:
This PR adds a "queue aggregator" type, intended to replace the hard-coded central task list for TD3 firmware. The class maintains an internal tuple of queue handles (which are just wrappers around pointers to the queue types), and provides a few different ways to send a message to a specific queue. 

- Tests for QueueAggregator are added
- Consolidated TestMessageQueue to be in the `common` include folder instead of being individually defined for Heater-Shaker and Thermocycler-Gen2
- Modified the TestMessageQueue type in order to fit with the requirements of the QueueAggregator
- The QueueAggregator is not used in anything outside of tests so far


stm32-modules/include/common/core/queue_aggregator.hpp
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    /**
+     * @brief Get the index of a queue, as generated by the Task List
+     *
+     * @tparam Message The type of the message
+     * @return size_t A unique index for this message type
+     */
+    template <typename Queue>
+    constexpr static auto get_task_idx() -> size_t {
```
sfoster1: can we use `consteval` here, which will cause a compilation failure if this can't be evaluated during compilation

stm32-modules/include/common/core/queue_aggregator.hpp
```diff
@@ -0,0 +1,274 @@
...
+        using Tag = typename Queue::Tag;
+        return TagTypes((Tag())).index();
+    }
+
+    /**
+     * @brief Send a message and automatically deduce the mailbox
+     * to forward it to.
+     *
+     * @tparam Tag The tag type of the queue to send to. This must be
+     * provided by each queue type delcared for the QueueAggregator.
```
sfoster1:
```suggestion
     * provided by each queue type declared for the QueueAggregator.
```

stm32-modules/include/common/core/queue_aggregator.hpp
```diff
@@ -0,0 +1,274 @@
...
+     * @tparam Message The type of message to send
+     * @param msg The message to send
+     * @param tag An instance of the tag type
+     * @return true if the message could be sent, false otherwise
+     */
+    template <typename Tag, typename Message>
+    auto send(const Message& msg, const Tag& tag = (Tag())) -> bool {
+        static_cast<void>(tag);
+        constexpr auto idx = get_task_idx<Tag>();
```
sfoster1: Isn't `get_task_idx`'s tparam the `Queue` and it derives the tag from there? How does this call work?

sfoster1 (changes requested):
The functionality looks good but I'm a bit confused by how `get_task_idx` works when you give it a `Tag`-type as a template param instead of a `Queue` type. Either there's a bug in the code that somehow doesn't prevent compilation or we really gotta change some of the names and the docs to match the functionality.

sfoster1 (approved):
Awesome! Can't wait to see this in practice.

PR #317, opened 2022-06-08
fix(thermocycler-gen2): increase seal backoff distance

fsinapi:
At the end of a seal movement to a limit switch, we have to back off of the limit switch because there is no way to distinguish which switch is pressed. Some EVT units were not backing off of the switch with the original distance of 0.5mm. Regardless of the mechanical reasons for this difference, a distance of 1.0mm seems to fix the issue.

The backoff requirement will not be present on DVT units and onwards (each switch will have its own line), so it is low risk to switch this for EVT.

Tested on an EVT unit that had been getting "stuck" every 3-4 open/close cycles. The unit can work for 10+ cycles now without problem.


sfoster1 (approved):
📈

PR #321, opened 2022-06-13
fix(thermocycler-gen2): I2C error mitigation

fsinapi:
Cycle testing with simultaneous thermal control & lid open/close control shows that the code for reading from the thermistor ADC's periodically fails. This results in an error state that ruins an active thermocycler run. The following changes appear to mitigate the effect:

- Increased I2C interrupt priority to match motor interrupts
- Added more detailed ADC error reporting through the raw ADC reading 
- Reduced the I2C bus speed from an excessive 1MHz to 400kHz
- Added a retry system, wherein a failed reading can be repeated

Tested with a 12 hour cycle test on EVT-06.

### DIFF
--- a/stm32-modules/include/thermocycler-gen2/thermocycler-gen2/thermal_plate_task.hpp
+++ b/stm32-modules/include/thermocycler-gen2/thermocycler-gen2/thermal_plate_task.hpp
@@ -93,7 +93,7 @@ class ThermalPlateTask {
     using Queue = QueueImpl<Message>;
     using Milliseconds = std::chrono::milliseconds;
     using Seconds = std::chrono::duration<double, std::chrono::seconds::period>;
-    static constexpr const uint32_t CONTROL_PERIOD_TICKS = 50;
+    static constexpr const uint32_t CONTROL_PERIOD_TICKS = 1;
     static constexpr double THERMISTOR_CIRCUIT_BIAS_RESISTANCE_KOHM = 10.0;
     static constexpr uint16_t ADC_BIT_MAX = 0x5DC0;
     static constexpr uint8_t PLATE_THERM_COUNT = 7;
--- a/stm32-modules/include/thermocycler-gen2/thermocycler-gen2/thermal_plate_task.hpp
+++ b/stm32-modules/include/thermocycler-gen2/thermocycler-gen2/thermal_plate_task.hpp
@@ -93,7 +93,7 @@ class ThermalPlateTask {
     using Queue = QueueImpl<Message>;
     using Milliseconds = std::chrono::milliseconds;
     using Seconds = std::chrono::duration<double, std::chrono::seconds::period>;
-    static constexpr const uint32_t CONTROL_PERIOD_TICKS = 1;
+    static constexpr const uint32_t CONTROL_PERIOD_TICKS = 50;
     static constexpr double THERMISTOR_CIRCUIT_BIAS_RESISTANCE_KOHM = 10.0;
     static constexpr uint16_t ADC_BIT_MAX = 0x5DC0;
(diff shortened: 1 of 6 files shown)


stm32-modules/include/thermocycler-gen2/thermocycler-gen2/thermal_plate_task.hpp
```diff
@@ -93,7 +93,7 @@ class ThermalPlateTask {
     using Queue = QueueImpl<Message>;
     using Milliseconds = std::chrono::milliseconds;
     using Seconds = std::chrono::duration<double, std::chrono::seconds::period>;
-    static constexpr const uint32_t CONTROL_PERIOD_TICKS = 50;
+    static constexpr const uint32_t CONTROL_PERIOD_TICKS = 1;
```
sfoster1: inadvertent from debugging? seems like a big change

stm32-modules/thermocycler-gen2/firmware/thermal/freertos_thermal_plate_task.cpp
```diff
@@ -82,7 +82,7 @@ static auto read_thermistor(const ADCPinMap &pin) -> uint16_t {
     auto result =
         _adc.at(static_cast<uint8_t>(pin.adc_index)).read(pin.adc_pin);
     if (std::holds_alternative<ADS1115::Error>(result)) {
-        return 0;
+        return 0xFFFF - (uint16_t)std::get<ADS1115::Error>(result);
```
sfoster1: What's your thought process on this one? Why not just the cast?

sfoster1 (changes requested):
Some other changes that probably need comments or removals

stm32-modules/include/thermocycler-gen2/thermocycler-gen2/thermal_plate_task.hpp
fsinapi: Haha yes absolutely a debugging change... will remove

stm32-modules/thermocycler-gen2/firmware/thermal/freertos_thermal_plate_task.cpp
fsinapi: This was in order to disambiguate an error response from just a faulty reading. I think that just using a cast and updating the error enumeration to start at 1 would be more appropriate.

sfoster1 (approved):
Looks good to me!

stm32-modules/thermocycler-gen2/firmware/thermal/ads1115.cpp
```diff
@@ -102,7 +102,8 @@ auto ADC::read(uint16_t pin) -> ADC::ReadVal {
 
     static_cast<void>(release_lock());
     if ((!i2c_ret) || (notification_val == 0)) {
```
sfoster1: [nit] this is pretty annoying to read, i think a series of ifs would be nicer, like
```
if (!i2c_ret) {
    return ReadVal(Error::ADCTimeout);
}
if (notification_val == 0) {
    return ReadVal(Error::I2CTimeout);
}
```

stm32-modules/thermocycler-gen2/firmware/thermal/thermal_hardware.c
```diff
@@ -56,7 +56,7 @@
  * On the higher end (low-priority) because timing
```
sfoster1: [nit] probably want to change this comment given that it's now on the lower end (high priority)

sfoster1 (approved):
Looks good, a couple small nits though

PR #322, opened 2022-06-15
refactor(thermocycler-gen2): made ADS1115 driver portable

fsinapi:
### Summary 
The ADC driver for the Thermocycler was originally written such that it could _only_ compile for the STM32 Cross build target, and it was not very portable (containing hard coded links to the low level firmware code). The ADC driver will be used on the Temp Deck Gen3 project, and in general will be useful because the ADS1115 is an extremely common ADC IC.

This refactor makes the core ADS1115 driver portable, with the hardware implementation defined through a template policy just like much of the rest of the Modules code. The PR also adds automated tests on the ADS1115 driver.

### Testing
- Ran the ADC driver on a board and read back the ADC readings with `M105.D`. Also disocnnected individual thermistors and confirmed that the firmware would enter an error state.
- Checked with a debugger that the memory address of the Policy instance is the same between both thermistor threads (plate & lid); see below
image

### Review Requests

- [x] Does the singleton behavior of the `AdcPolicy `class make sense? The intention is to be able to share the policy instantiations between threads (one for the plate thermistors, one for the lid thermistor).
- [ ] Double check that the logic of `ADC::Read` is equivalent to the implementation it replaces


stm32-modules/include/common/core/ads1115.hpp
```diff
@@ -0,0 +1,178 @@
...
+    using ReadVal = std::variant<uint16_t, Error>;
+
+    ADC() = delete;
+    /**
+     * @brief Construct a new ADS1115 ADC
+     * @param[in] addr The I2C address of the ADC
+     * @param[in] id The ID of the ADC, which will link to one of the
+     * interrupts defined in \ref thermal_hardware.h
+     */
+    ADC(Policy& policy) : _policy(policy) {}
```
sfoster1: not really critical but single-argument ctors should be marked `explicit` so they don't participate in implicit conversions

sfoster1 (approved):
Yeah, I think this is the right approach - hardware handling should be a singleton since it represents a single instance of something physical. Nice!

PR #324, opened 2022-06-16
fix(thermocycler-gen2): changes to uniformity error check

fsinapi:
Per internal discussion, it seems that the uniformity error checking is excessively aggressive - the Thermocycler tends to flag a uniformity error while thermistors are still settling. This PR adjusts the behavior so that the Thermocycler gives an extra 30 seconds _after_ the end of the Overshoot phase until it starts checking for the error. 

Tested on hardware by setting target temperature and unplugging thermistors. Unplugging before the 30 seconds and then plugging the thermistor back in shows that no uniformity error is flagged. If a thermistor is unplugged _after_ the 30 seconds, the uniformity error is triggered and stays active.


sfoster1 (approved):
Nice

PR #325, opened 2022-06-17
feat(tempdeck-gen3): add thermistor table

fsinapi:
The Tempdeck Gen3 uses a different thermistor part than the HS and TC2 modules, so it had to be added to the `generate_thermistor_table.py` script & cmake. As a part of this, I took a little time to refactor the thermistor generation script to be a little friendlier to future additions so that adding new thermistors will be less painful. Now each thermistor type has an optional argument to pass its CSV file, and any files passed to the script get included in the generated header/source files.


stm32-modules/common/src/nxft15xv103fa2b030.csv
```diff
@@ -0,0 +1,171 @@
+              R25          10  k ohm +/- 1%
```
sfoster1, line 1: this is an incredibly cursed format my lord

sfoster1 (approved):
Looks good. Ouch that format though

stm32-modules/common/src/nxft15xv103fa2b030.csv
fsinapi: 🙃

PR #326, opened 2022-06-21
feat(tempdeck-gen3): parameterized gcode parser

fsinapi:
Adds a parameterized gcode parser to simplify writing code for parsing gcode commands & arguments.

The parser accepts a parameter pack of structure types that represent individual arguments in a command, and each argument can be marked as either 1) optional or not optional 2) with a value (a member variable named `value`) or without a value. Every argument struct must also have a member struct that marks whether it's present after parsing.

If successful, the parser returns a tuple containing an instance of each argument type to make it easy for the caller to act on the values passed into the command. See the added tests for the use of the parser functions.

Not currently used outside of tests, but I will plan on integrating this for the Temp Deck Gen3 firmware.


stm32-modules/include/common/core/gcode_parser.hpp
```diff
@@ -121,6 +122,201 @@ auto parse_value(const Input& start_from, Limit stop_at)
...
+    {std::is_same_v<decltype(Arg::required), bool>};
+    // Variable within the argument - flag for if it is present
+    {std::is_same_v<decltype(a.present), bool>};
+};
+
+// Gcode argument that has a prefix
+template <typename Arg>
+concept GCodeArgumentWithPrefix = requires(Arg& a) {
+    // Prefix needs to be iterable
+    {Arg::prefix.cbegin()};
```
sfoster1:
```suggestion
    {std::is_same_v<std::remove_cvref<decltype(*Arg::prefix.cbegin())>, char>}
```
?

sfoster1: oh it also might be better to use `std::begin` and `std::end` rather than the member fns

stm32-modules/include/common/core/gcode_parser.hpp
```diff
@@ -121,6 +122,201 @@ auto parse_value(const Input& start_from, Limit stop_at)
...
+    arg.present = true;
+    return start_from;
+}
+
+/**
+ * @brief This integer-templated struct allows us to recurse through all
+ * of the arguments present in a command, checking each individual arg
+ * for validity and then passing along to the next one.
+ */
+template <size_t N>
```
sfoster1: I see what you're doing where but I personally don't love this pattern and find it tough to read - the whole thing with the two layers of templates and all. I'd much prefer something that [analogizes much better to normal runtime recursion](https://godbolt.org/z/sdMee3P96) - should let you drop the structs.

sfoster1 (changes requested):
Looks fundamentally good, but can we use the other style of recursion instead (as in the inline)? I think that'll make it a bit easier to udnerstand.

sfoster1 (approved):
Looks great! I like the use of the `if constexpr (sizeof...` to avoid the explicit specialization base case.

PR #329, opened 2022-06-27
feat(tempdeck-gen3): USB communication setup

fsinapi:
### Summary
The tempdeck-gen3 can now talk over USB. It only responds with error messages, but everybody has to start somewhere.

This PR adds USB by adding:
* The low-level callback code for USB, largely copy + pasted from the Thermocycler Gen2
* A small set of initial message definitions, as well as definitions of the queue types & aggregators for firmware and tests
* The initial code for the Host Comms task, with a single gcode defined
  * This gcode tries to send to the System Task, which isn't defined, so an error always comes back

Tested on hardware that sending M115\n results in an error code `ERR002:internal queue full`, and any other string results in `ERR003:unhandled gcode`


sfoster1 (approved):
Looks good to me! Give it a test on a windows machine too, although it probably works fine given that it looks like you copied over the descriptor twiddle hack

fsinapi:
> Looks good to me! Give it a test on a windows machine too, although it probably works fine given that it looks like you copied over the descriptor twiddle hack

Tested on a windows machine, works fine 👍

PR #330, opened 2022-06-29
fix(startup): exception handling is more resilient 

fsinapi:
In the module-startup program, the processor has the opportunity to become locked in an endless loop of the `Fault_Handler` if the Link Register holds an invalid EXC_RETURN value, because the program uses this value to define the proper exception return behavior. Sometimes `LR` seems to be loaded with an invalid value when the built-in bootloader enters an exception. This handler _always_ wants to use Return to Handler Mode when jumping to the `Post_Fault_Handler`, so this PR makes sure it does that.

Tested with WIP TD3 DFU code. Without this change, the CPU cycles through the Fault Handler forever. With this change, it ends up entering DFU mode correctly.

### DIFF
--- a/stm32-modules/common/module-startup/module-startup.S
+++ b/stm32-modules/common/module-startup/module-startup.S
@@ -88,6 +88,17 @@ Fault_Handler:
     ldr     r0, =Post_Fault_Handler
     str     r0, [sp,#0x18]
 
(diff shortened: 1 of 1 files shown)


Laura-Danielle (approved):
lgtm.

sfoster1 (approved):
a) yuck b) awesome

PR #332, opened 2022-06-30
feat(tempdeck-gen3): add system task

fsinapi:
### Overview

Adds a System Task and two commands to set the serial number (M996) and jump to the bootloader (dfu). Both commands are largely copied from Thermocycler-Gen2.

This PR _does_ test out the "return address" functionality of the message system! The message to the Host Comms task to disable USB specifies a return address for the reply, in case another task (such as in an error case) has to send the same message.

Tested on a TD3 PCB. The serial number is persistent across power cycles, and the DFU command is able to enter the bootloader and the PCB enumerates as a DFU device.


stm32-modules/tempdeck-gen3/firmware/system/stm32g4xx_hal_timebase_tim.c
```diff
@@ -0,0 +1,140 @@
+/* USER CODE BEGIN Header */
```
sfoster1: is this autogenerated? where'd it come from? is it just the user bindings or is it a copy of a hal file?

sfoster1 (approved):
Looks good, confused about that hal file though

stm32-modules/tempdeck-gen3/firmware/system/stm32g4xx_hal_timebase_tim.c
fsinapi: This is based on an autogen file, but I think renaming it + reformatting it a bit would make a lot of sense

PR #337, opened 2022-07-01
feat(tempdeck-gen3): add ui task & heartbeat led

fsinapi:
Adds a UI task driven by a FreeRTOS soft timer, and for now that task just sets a heartbeat LED. This task will be expanded in the future when an actual UI is relevant.

Tested on a TD3 board, the LED pulsing is consistent and pleasant.


sfoster1 (changes requested):
Looks good, but let's set a good precedent by having a test for the heartbeat class since it's so testable

sfoster1 (approved):
Nice!

PR #338, opened 2022-07-02
fix(heater-shaker): add post homing delay

### LINKED ISSUE #10858: bug: heater-shaker errors out on commands immediately after deactivate shaker
# Overview

Heater-Shaker's `deactivate_shaker` method sends a home command to stop shaking ~and returns immediately~ . This results in issues when a shake command immediately follows a deactivate command- the shake command errors out ~because the module is still in motion~. Upon testing it's been observed that home command in fact waits until homing completes, but there still seems to be something that causes a communication lock-up when a shake command immediately follows a deactivate command.

Found by @ecormany 

# Steps to reproduce

Run this protocol
```
from opentrons import protocol_api, types
​
metadata = {'apiLevel': '2.12'}
​
def run(protocol):
​
	# labware
	plate = protocol.load_labware('nest_96_wellplate_100ul_pcr_full_skirt', location='2')
	tiprack = protocol.load_labware('opentrons_96_tiprack_300ul', location='1')
​
	# pipettes
	pipette = protocol.load_instrument(
		 'p300_single', mount='left', tip_racks=[tiprack])
​
	# modules
	hs = protocol.load_module('heaterShakerModuleV1', 5)
	hs.close_labware_latch()
	hs.set_and_wait_for_shake_speed(500)
	hs.deactivate_shaker()
	hs.set_and_wait_for_shake_speed(500)

```

# Current behavior

Results in this runtime error during execution of the second `set_and_wait_for_shake_speed`:
```
/dev/ot2_module_heatershaker0: No response to 'M3 S500'
```
The issue does not occur if there's a small delay between the deactivate and the next shake command. Tested with a 10sec delay.

# Expected behavior

`deactivate_shaker` should wait to return until homing is complete so that subsequent, immediate shake commands can be executed. 

# Thoughts

Please make sure that my conclusion that the homing is causing the next set shake command to error out is correct. Assuming that it is, ~here's some thoughts on my suggestion to wait until homed:~ Removed the thoughts since it isn't clear why homing is causing the next command to lock up. Testing revealed that the home command waits until homing is observably complete. Whether the firmware is silently busy for longer is not known.

pmoegenburg:
Homing gcode immediately followed by SetRPM produced a `motor unable to move` error.

This fix introduces a delay at the end of the homing sequence to ensure motor control is fully deactivated before subsequent SetRPM commands may be executed.

Closes Opentrons/opentrons#10858

### DIFF
--- a/stm32-modules/include/heater-shaker/heater-shaker/motor_task.hpp
+++ b/stm32-modules/include/heater-shaker/heater-shaker/motor_task.hpp
@@ -169,20 +169,24 @@ class MotorTask {
         } else {
             policy.homing_solenoid_disengage();
             error = policy.set_rpm(msg.target_rpm);
-            setpoint = msg.target_rpm;
-            state.status = State::RUNNING;
-            policy.delay_ticks(MOTOR_START_WAIT_TICKS);
-            if ((msg.target_rpm != 0) &&
-                (policy.get_current_rpm() < MOTOR_START_THRESHOLD_RPM)) {
-                error = (error == errors::ErrorCode::NO_ERROR)
-                            ? errors::ErrorCode::MOTOR_UNABLE_TO_MOVE
-                            : error;
-                policy.stop();
-                state.status = State::ERROR;
-                setpoint = 0;
+            if (error == errors::ErrorCode::NO_ERROR) {  // only proceed if
+                                                         // target speed legal
+                setpoint = msg.target_rpm;
+                state.status = State::RUNNING;
+                policy.delay_ticks(MOTOR_START_WAIT_TICKS);
+                if ((msg.target_rpm != 0) &&
+                    (policy.get_current_rpm() < MOTOR_START_THRESHOLD_RPM)) {
+                    error = errors::ErrorCode::MOTOR_UNABLE_TO_MOVE;
+                    policy.stop();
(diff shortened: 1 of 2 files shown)


fsinapi (approved):
LGTM

stm32-modules/include/heater-shaker/heater-shaker/motor_task.hpp
```diff
@@ -171,7 +174,9 @@ class MotorTask {
             policy.delay_ticks(MOTOR_START_WAIT_TICKS);
             if ((msg.target_rpm != 0) &&
                 (policy.get_current_rpm() < MOTOR_START_THRESHOLD_RPM)) {
-                error = errors::ErrorCode::MOTOR_UNABLE_TO_MOVE;
+                error = (error == errors::ErrorCode::NO_ERROR)
```
sfoster1: why do we have to do this? if it's the underlying driver doing something weird, let's drop a comment to that effect here and add a test.

sfoster1 (changes requested):
Solution looks good, but that error handling thing really needs a comment and probably a test.

sfoster1 (approved):
Looks good to me!

sanni-t:
Thanks for fixing this!

PR #339, opened 2022-07-06
feat(tempdeck-gen3): add thermal tasks

fsinapi:
Adds the Thermal Task and Thermistor Task for the tempdeck-gen3, as per the architecture documentation.

The tasks don't really _do_ anything yet - the thermistor task sends messages with empty thermistor data and valid timestamps, while the thermal task doesn't do anything but hold the latest thermistor update data.

* Thermistor task policy just requires a way to get the current time (for attaching a timestamp to the message)
* Thermal task policy doesn't have _any_ real requirements, just has a placeholder expression to allow a placeholder policy for now
* Added some very basic tests for the tasks as placeholders

Tested on a board with a debugger that the Thermal Task keeps getting messages from the Thermistor Task, and that the timestamp keeps incrementing properly.


sfoster1 (approved):
Nice!

PR #340, opened 2022-07-11
feat(tempdeck-gen3): read thermistors

fsinapi:
Adds the required code to read from the external ADC on the Temp Deck Gen3 board and convert the readings to temperature values.

- Adds in code to read/write 16 bit registers over I2C, as well as wait for the ADC Ready interrupt to trigger before reading the result from each channel
- Reads NTC0 and NTC1 as placeholders for now, final thermistor/channel mapping is TBD
- There isn't any code to get the readings off the board yet, that will be done with future PR's

Tested with a debugger by sticking a 10k resistor into each channel and checking the resulting temperature in the `thermal task` instance. Each channel converts a 10k resistor to around 25ºC, which matches well with the nominal thermistor values.


sfoster1 (approved):
Looks good to me!

PR #341, opened 2022-07-11
feat(tempdeck-gen3): add GetTemperatureDebug command

fsinapi:
Adds command M105.D to get the temperature of the system. This is a debug command that includes both the temperature readings for the thermistors _and_ the raw ADC counts.

Tested on hardware by sending the command through Python with 10k resistors in the NTC0 and NTC1 slots.


sfoster1 (approved):
Looks good to me!

PR #342, opened 2022-07-13
feat(tempdeck-gen3): add SetPeltierDebug command

fsinapi:
Adds a command M104.D to set the PWM level of the peltiers. **This PR does not include hardware drivers to control the peltiers**, that will be implemented in a subsequent PR. This PR _does_ include the logic for the SetPeltierDebug command, as well as a policy implementation for the tests & matching tests.

Tested that sending M104.D succeeds on a board (but nothing happens other than getting an ACK back).


sfoster1 (approved):
looks good to me, big steps!

PR #343, opened 2022-07-13
feat(tempdeck-gen3): added a python class for testing the tempdeck

fsinapi:
Adds a simple wrapper Python class to test the Tempdeck Gen3 over serial


sfoster1 (approved):
noice

PR #346, opened 2022-07-15
feat(tempdeck-gen3): peltier driver firmware

fsinapi:
Adds the necessary code to control the peltier drive circuitry.

* Channels 0 and 1 are used for cooling and heating, respectively.
* A 12v Rail Enable line is set to  high on startup
* A Peltier Enable line is set high or low based off of the Thermal Task
* TIM1 runs with a period of ~500kHz, and PWM output is clamped in the range [0.1, 0.9] based off of the constraints of the constant current topology.

Checked with an oscilloscope that:
* The PWM runs at 500kHz
* The PWM is always between [0.1,0.9]
* The enable line is set/disabled properly
* The correct channel is set based off of positive/negative power settings with M104.D
* Sending `M104.D S0` disables the peltier entirely

### DIFF
--- a/stm32-modules/tempdeck-gen3/firmware/thermal_control/thermal_hardware.c
+++ b/stm32-modules/tempdeck-gen3/firmware/thermal_control/thermal_hardware.c
@@ -7,7 +7,7 @@
 
 // RTOS includes
 #include "FreeRTOS.h"
-
+#include "task.h"
 
 // ***************************************************************************
 // Local constants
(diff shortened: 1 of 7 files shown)


stm32-modules/tempdeck-gen3/firmware/thermal_control/thermal_hardware.c
```diff
@@ -0,0 +1,269 @@
...
+
+// ***************************************************************************
+// Local constants
+
+#define PULSE_WIDTH_FREQ (500000)
+#define TIMER_CLOCK_FREQ (170000000)
+
+// Given a desired frequency of 500kHz, we do not need to prescale the timer
+#define TIM1_PRESCALER (0)
+#define TIM1_RELOAD ((TIMER_CLOCK_FREQ / (PULSE_WIDTH_FREQ * (TIM1_PRESCALER + 1))) - 1)
```
sfoster1, line 21: drop a comment with the value here?

sfoster1 (approved):
Nice. 500khz, wow!

PR #347, opened 2022-07-18
fix(heater-shaker): heatpad circuit checking

pmoegenburg:
The HAL_TIM_OC_DelayElapsedCallback attempts to check for heatpad open and short circuit conditions via the heatpad current sensing pin once per second. TIM4 channels 2 and 4 are used to trigger this callback at 10% and 90% of the heatpad pwm period. After one second has elapsed, on the first 90% period callback, the DAC and comparator settings are updated and the pwm pulse is prevented from updating. This is done to ensure the peripherals have sufficient time to update and checking conditions are undisturbed. The following period, if the heatpad pwm pulse is greater than 20% of the pwm period, the current sensing pin will be checked to be greater than 0.2V during the 10% callback to confirm there is no open circuit, and if the heatpad pwm pulse is less than 80% of the pwm period, the current sensing pin will be checked to be less than 0.2V during the 90% callback to confirm there is no shorted circuit. Following the last check, the update_lock is removed and the comparator threshold is returned to 3.2V to detect an overcurrent condition during the remainder of the one second checking period.

Locking the pwm pulse updating eliminated false positives.

If there is an error, the specific error will be reported back by heater_hardware_power_set and an error gcode will be produced.


stm32-modules/include/heater-shaker/heater-shaker/heater_task.hpp
```diff
@@ -402,24 +406,28 @@ class HeaterTask {
...
+            if (error) {
                 state.system_status = State::ERROR;
                 setpoint = std::nullopt;
-                state.error_bitmap |= State::CIRCUIT_ERROR;
-                auto error_message =
-                    messages::HostCommsMessage(messages::ErrorMessage{
-                        .code =
-                            errors::ErrorCode::HEATER_HARDWARE_ERROR_CIRCUIT});
+                auto error_message = messages::ErrorMessage{};
+                if (error == 9){
```
sfoster1: let's get some names for these values instead of just integer constants please. ideally they could be in a header shared by both c and c++.

sfoster1 (changes requested):
Definitely need to get names for some of these magic numbers and ideally make that `uint8_t` return something with a better name.

Also, this is a pretty complex and hardware-focused PR. Would you mind adding an in-depth PR message, and thus commit message when this is merged, about what this is and what it does?

stm32-modules/include/heater-shaker/systemwide.h
```diff
@@ -18,3 +18,5 @@ typedef enum {
     WHITE_AMBER
 } LED_COLOR;
 typedef enum { SOLID_HOT, SOLID_HOLDING, PULSE, MODE_OFF } LED_MODE;
+
+typedef enum { NONE, SHORT, OPEN, OVERCURRENT } HEATPAD_CIRCUIT_ERROR;
```
fsinapi: I think it would make sense to prefix these names since they're pretty generic and they're in the global namespace

stm32-modules/heater-shaker/firmware/heater_task/heater_hardware.c
```diff
@@ -285,39 +288,47 @@ void heater_hardware_power_disable(heater_hardware* hardware) {
...
-        HAL_TIM_PWM_Start_IT(&internal->pad_tim, HEATER_PAD_SHORT_CHECK_TIM_CHANNEL);
-        internal->pwm_config.Pulse = HEATER_PAD_OPEN_CHECK_PULSE;
-        HAL_TIM_PWM_Stop_IT(&internal->pad_tim, HEATER_PAD_OPEN_CHECK_TIM_CHANNEL);
-        HAL_TIM_PWM_ConfigChannel(
-            &internal->pad_tim, &internal->pwm_config, HEATER_PAD_OPEN_CHECK_TIM_CHANNEL);
-        HAL_TIM_PWM_Start_IT(&internal->pad_tim, HEATER_PAD_OPEN_CHECK_TIM_CHANNEL);
-        internal->heater_started = true;
-    } else {
-        HEATER_PAD_LL_SETCOMPARE(internal->pad_tim.Instance, setting);
+    if (!internal->update_lock) {
```
fsinapi, line 313: My understanding is that, if the update lock is enabled, this function will silently return without updating the power. If we need to lock out PWM updates for period of time, I think it makes sense to actually cache the setting and then update the timer compare value once the lock is cleared.

fsinapi (commented):
The fix looks good, nice debugging! I had a couple of pretty minor things.

stm32-modules/heater-shaker/firmware/heater_task/heater_hardware.c
```diff
@@ -35,6 +36,7 @@ typedef struct {
     heatpad_cs_state heatpad_cs_status;
     uint16_t pwm_pulse_duration;
     uint16_t period_count;
+    bool update_lock;
```
sfoster1: this is accessed from threadmode and from an ISR. Since this is c, you should include `<stdatomic.h>`and make this an [`_Atomic bool`](https://en.cppreference.com/w/c/language/atomic) or `atomic_bool` which should be the same.

sfoster1 (commented):
This is awesome! Agree with @fsinapi and also think we should use an atomic bool for the update lock but overall, great find

stm32-modules/heater-shaker/firmware/heater_task/heater_hardware.c
pmoegenburg: Should I make all the shared vars _Atomic? heatpad_cs_status, pwm_pulse_duration, cached_pulse_setting

sfoster1: Oh, yeah definitely. It probably isn't a problem (and all it's doing is inserting memory barriers) because [arm loads and stores of word-aligned words (uint32), half-word aligned half-words (uint16), and all bytes  are strongly ordered](https://developer.arm.com/documentation/ddi0406/cb/Application-Level-Architecture/Application-Level-Memory-Model/Memory-types-and-attributes-and-the-memory-order-model/Atomicity-in-the-ARM-architecture?lang=en)

stm32-modules/heater-shaker/firmware/heater_task/heater_hardware.c
pmoegenburg: ramp with cached_pulse_setting
ramp with cached_pulse_setting

pmoegenburg: ramp without cached_pulse_setting
ramp without cached_pulse_setting

pmoegenburg: Not using the cached_pulse_setting will prevent pwm pulse from updating for one cycle. The following cycle, the PID will calculate power update with the new, increased error. I think this approach seems correct, and the ramp profile seems to agree

fsinapi (approved):
LGTM as long as a ramp with these changes looks good!

sfoster1 (approved):
Nice! Looks good to me

PR #348, opened 2022-07-18
fix(heater-shaker): add ack to error responses

pmoegenburg:
For serial_connection::_send_data processing, which includes `read_until(match=self._ack)`, which looks for an `OK` in a response in order to process it, whether that's a success or error response. TC1 appends error responses with `OK`

Supports Opentrons/opentrons#10765


sfoster1 (commented):
Couldn't we change the python side to not look for the ok anymore? terminate on a `\n` instead? this is kinda gross

sfoster1 (approved):
If everything else does ok for errors, i guess we have to also, so it goes

PR #349, opened 2022-07-19
feat(thermocycler-gen2): peltier fine tuning

fsinapi:
This PR makes a number of changes to the Thermal Plate control firmware for better thermal performance.

- Added the ability to configure the integrator reset threshold in the PID driver.
- Made some significant changes to how the firmware deals with integral windup. Previously, it would always reset the integral term once a peltier reached its target temperature; this ended up causing very noticeable oscillations around target temperatures. There are now a few different methods of compensation described in the updated Thermal README file:
  - Conditional Integration when moving away from ambient. The proportional band is defined as 1.0/Kp
  - Integral term reset at 3º from target when moving towards ambient. Conditional Integration showed too much overshoot in this case, as did resetting the integral term _at_ the target
  - When **heating up** to a temperature that is below the heatsink temperature, the initial target temperature is shifted 2º down. Testing showed that there is a consistent over-overshoot of ~2º, and this brings the max plate temperature within reasonable bounds.
- Added new default PID constants to the peltiers. These were iteratively developed off of PID Tuner values.

### DIFF
--- a/stm32-modules/common/src/pid.cpp
+++ b/stm32-modules/common/src/pid.cpp
@@ -59,6 +59,7 @@ auto PID::compute(double error, double sampletime) -> double {
 auto PID::reset() -> void {
     _last_error = 0;
     _last_iterm = 0;
+    _reset_trigger = NONE;
 }
 
 auto PID::arm_integrator_reset(double error) -> void {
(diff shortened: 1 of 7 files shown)


sfoster1 (approved):
Looks good to me, nice tests and documentation

PR #351, opened 2022-07-19
feat(tempdeck-gen3): add initial simulator

fsinapi:
Adds an initial working version of the simulator. It doesn't have any actual simulated temperatures yet, but each task gets spun up and message passing works as intended.

Tested by running the simulator with `--stdin` and sending some gcodes, responses come back as expected.


sfoster1 (approved):
love it

PR #353, opened 2022-07-20
feat(tempdeck-gen3): add fan driver

fsinapi:
Adds code to drive the fan on the Tempdeck. Uses TIM16 as a PWM generator on pin A6.

- Adds M106 SetFanManual to set the power of the fan from 0-1
- Adds M107 SetFanAutomatic to set the fan to automatic mode - right now this just sets the power to 0
- _For now,_ the PWM to turn the fan off is 0.01% instead of 0%. This is because we cannot turn off the 12v rail (because it is also used for the peltier driver), and a PWM of 0% causes the fan driver to think there is no control line connected and set the output to 100%.
  - Testing hasn't been done with the final fan part yet so this may change!

Verified on hardware that M106 sets the power accordingly, and sending `M106 S0` or `M107` turns off the fan.

### DIFF
--- a/stm32-modules/tempdeck-gen3/firmware/thermal_control/thermal_hardware.c
+++ b/stm32-modules/tempdeck-gen3/firmware/thermal_control/thermal_hardware.c
@@ -41,11 +41,25 @@
 // Peltier drive circuitry cannot support higher PWM than 0.9
 #define MAX_PELTIER_POWER (0.9)
 
+#define FAN_PWM_Pin (GPIO_PIN_6)
+#define FAN_PWM_GPIO_Port (GPIOA)
+#define FAN_PULSE_WIDTH_FREQ (1000)
+
+#define TIM16_PRESCALER (67)
+// Calculates out to 2499
+#define TIM16_RELOAD ((TIMER_CLOCK_FREQ / (FAN_PULSE_WIDTH_FREQ * (TIM16_PRESCALER + 1))) - 1)
+
+// PWM should be scaled from 0 to FAN_MAX_PWM, inclusive
+#define FAN_MAX_PWM (TIM16_RELOAD + 1)
+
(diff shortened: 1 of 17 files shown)


stm32-modules/tempdeck-gen3/firmware/thermal_control/thermal_hardware.c
```diff
@@ -153,15 +174,33 @@ bool thermal_hardware_set_peltier_cool(double power) {
...
+        return false;
+    }
+    if(power == 0.0) {
+        // The fan controller will default to full power if it thinks the
+        // control line is disconnected, and unfortunately it thinks a 0% PWM
+        // is a disconnection. So the lowest allowable PWM is 0.01, which 
+        // still results in the fan staying still.
+        power = 0.01;
+    }
+    uint32_t pwm = power * (double)FAN_MAX_PWM;
```
sfoster1, line 193: it makes me a little nervy to do this check this way, for two reasons:
- exact float comparison; what if somebody commands 0.005 power?
- what if for some reason in the future we change the period to a value not a multiple of 100, and this results in a 0 value in the compare register?

Recommend making the check more like
```
uint32_t pwm = power * (double)FAN_MAX_PWM;
if (pwm == 0) { pwm = 1; }
```

sfoster1 (changes requested):
Looks good (and lol at the fan thing) but let's change the way we do the minimum power setting to be more robust.

stm32-modules/tempdeck-gen3/firmware/thermal_control/thermal_hardware.c
fsinapi: @sfoster1 tested this change on hardware and it works 👍

sfoster1 (approved):
Nice

PR #354, opened 2022-07-22
fix(heater-shaker): prevent set_rpm during homing

### LINKED ISSUE #352: bug: heater-shaker: gcode cache seems to get full unexpectedly and stay full
While testing [this home command bug](https://github.com/Opentrons/opentrons/issues/10858), I received a `gcode cache full` error. According to [this doc](https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/512819201/Interaction+GCodes+and+Errors#Errors), this error is expected to be raised when a lot of gcode commands are received at once such that the input buffer is filled and the OS hasn't had a chance to address them yet. The error is supposed to go away once the commands get handled and the buffer is freed of space. But during my testing, it seemed like the cache was not being emptied as I kept receiving `gcode cache full` error on every gcode sent.

I will be re-testing this to add logs and more details

pmoegenburg:
Closes #352


sfoster1 (changes requested):
Definitely need a test for this one unfortunately

pmoegenburg:
You beat me to it

sfoster1 (approved):
Haha fair enough - just got the ping. Looks good to me!

PR #355, opened 2022-07-23
fix(heater-shaker): return EMPTYSN to M115 if serial number not programmed

pmoegenburg:


fsinapi (approved):
LGTM!

stm32-modules/heater-shaker/tests/test_m115.cpp
```diff
@@ -23,6 +23,36 @@ SCENARIO("GetSystemInfo (M115) response works", "[gcode][parse][m115]") {
                 REQUIRE_THAT(buffer, Catch::Matchers::EndsWith(suffix));
             }
         }
+        WHEN("filling response with an unset serial number (all 0xFF)") {
+            std::array<char, SYSTEM_WIDE_SERIAL_NUMBER_LENGTH> TEST_SN;
+            memset(TEST_SN._M_elems, 0xFF, TEST_SN.size());
```
sfoster1: whoa whoa we definitely should not be poking STL internals. arrays don't have convenient ctors but what they do have is [fill](https://en.cppreference.com/w/cpp/container/array/fill) which should do what you want

stm32-modules/include/heater-shaker/heater-shaker/gcodes.hpp
```diff
@@ -595,8 +597,29 @@ struct GetSystemInfo {
         if (written == write_to_limit) {
             return written;
         }
-        written = write_string_to_iterpair(written, write_to_limit,
-                                           serial_number.begin());
+        // If the serial number is unwritten, it will contain 0xFF which is
+        // an illegal character that will confuse the host side. Replace the
+        // first instance of it with a null terminator for safety.
+        constexpr uint8_t invalid_ascii_mask = 0x80;
```
sfoster1, line 603: Let's not overcomplicate this - let's just look for the first exact match for `0xff`.

sfoster1 (changes requested):
Let's definitely do a quick change to avoid poking stl internals. I think the actual logic would be better off checking for `0xff` explicitly rather than "anything that's not a 7 bit ascii" but if it works it's ok to not change it.

sfoster1 (approved):
Looks good to me!

PR #359, opened 2022-07-26
fix(heater-shaker): use non-malloc snprintf implementation

pmoegenburg:
Same as #275

### DIFF
--- a/stm32-modules/CMakeLists.txt
+++ b/stm32-modules/CMakeLists.txt
@@ -9,6 +9,7 @@ else()
 endif()
 
 find_package(Clang)
+find_package(MpalandPrintf)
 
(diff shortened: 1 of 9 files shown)


sfoster1 (approved):
Looks good once you get it working. I think the reason it's not is that another cmake project creates the library. 

There's two ways to fix that:
1. Change the find project implementation in cmake-utils to provide a function to add the library with a name altered by some input (i.e. so there's `Mpalaprint-thermocycler` and `Mpalaprint-heater-shaker`
2. As long as all the architectures match, do the find project and add library in the `stm32-modules/CMakeLists.txt` and just add the dep here

fsinapi (approved):
Nice

PR #362, opened 2022-07-26
feat(thermocycler-gen2): support for second seal switch

fsinapi:
Adds in support for the second seal switch line that was added in the Revision 3 PCB. The second line is connected to pin C12, and is used for the **retraction** switch. The original line is used for the **extension** switch.

* M901.D now returns both switches, `E:` and `R:`, separately
* Seal movements now only look at one of the switches. Movements with positive steps (aka retraction) only look at the retraction switch, while movements with negative steps (aka extension) only look at the extension switch.
* The hardware policy is configured to accommodate earlier boards (for EVT unit support).

Tested on a Rev2 board to check that the backwards compatibility works as intended, and tested on a Rev3 board to check the following:
* The two switches are read correctly (M901.D)
* Movements only fail to start if the one specific switch for that direction is triggered
* Movements only stop when the one specific switch for that direction triggers an interrupt


sfoster1 (approved):
Looks good to me!

PR #363, opened 2022-07-26
fix(heater-shaker): remove sha from firmware version

pmoegenburg:


fsinapi (commented):
I agree with this, but my strong preference is to make this change in the cmake and keep the firmware agnostic to the formatting of its version strings. We can just add the flag `--abbrev=0` in the call to `git describe` and it'll take out the hash.

sfoster1 (changes requested):
Yeah you know I recommended this earlier but looking at it again, if you edit this: https://github.com/Opentrons/opentrons-modules/blob/e4454125e5120bd382fd17e7badde3d6730b0ad1/stm32-modules/heater-shaker/src/CMakeLists.txt#L25-L43 to only do the git describe bit (remove lines 38-43 inclusive) you should get the exact behavior we're looking for here

sfoster1:
> I agree with this, but my strong preference is to make this change in the cmake and keep the firmware agnostic to the formatting of its version strings. We can just add the flag `--abbrev=0` in the call to `git describe` and it'll take out the hash.

I don't think we want to do this - we want just straight up `git describe`, as implemented, because release builds should be exactly the tag commit

fsinapi (approved):
LGTM once it's tested!

sfoster1 (approved):
Looks good to me, but let's test locally before trying it out

PR #364, opened 2022-07-26
feat(thermocycler-gen2): front button LED pulsing and blinking

fsinapi:
Adds pulsing and blinking behaviors to the front button LED. The LED isn't connected to any timer line, so it has to be pulsed manually with a periodic timer (in this case a FreeRTOS soft timer).

The callback for the front LED _could_ send a message to the system task that then updates the message when the task runs, but this ends up causing very noticeable jitter to the LED blinking whenever the seal motor runs (which is a high priority and high frequency interrupt). Keeping the LED handling to a single callback invoked by the FreeRTOS timer task keeps jitter invisible to the naked eye.

The LED behavior is as follows:
* When the lid is opening or closing (including seal motor movements), the LED pulses
* When the lid begins a Plate Lift action, the LED blinks two times
* At all other times, the LED is just left on

Video of the updated LED behavior on a physical unit was approved by Design.


sfoster1 (approved):
Looks good to me!

PR #365, opened 2022-07-28
feat(thermocycler-gen2): run plate lift with long button press

fsinapi:
Updates the behavior of the front button to activate a plate lift action when the button is held for 3 seconds. This includes a complete refactor in how the button is controlled; rather than using a software timer, there is a dedicated task that handles monitoring the button. This task waits for an IRQ from the button being pressed, and then starts polling the state of the button (including debouncing for press & release) to check for when it is released. A dedicated class handles when to send a message to the Motor Task, and whether it should classify that message as a short or long button press.

Tested on a unit that:
* Pressing the button and releasing it before 3 seconds results in the lid opening or closing as it did before
* Pressing the button for more than 3 seconds results in nothing happening if the lid is closed
* Pressing the button for more than 3 seconds results in a plate lift action if the lid is open
* If the button is held down to trigger a plate lift, and isn't released until the plate lift is over, the debounce handling prevents a false positive button press.

### DIFF
--- a/stm32-modules/include/thermocycler-gen2/firmware/system_hardware.h
+++ b/stm32-modules/include/thermocycler-gen2/firmware/system_hardware.h
@@ -8,6 +8,8 @@ extern "C" {
 
 // Front button can only be pressed at 200ms increments
 #define FRONT_BUTTON_DEBOUNCE_MS (200)
+// Front button should be queried at this frequency after debouncing
+#define FRONT_BUTTON_QUERY_RATE_MS (50)
 
 // Type for callback when the button has been pressed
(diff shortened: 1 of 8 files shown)


sfoster1 (approved):
This pretty much looks good, but does the thermocycler issue anything out via the usb when this happens? does the host robot know when the button is pressed?

fsinapi:
> does the thermocycler issue anything out via the usb when this happens? does the host robot know when the button is pressed?

The Thermocycler does not send any unsolicited messages to the host after a button press, nor does the host expect this. The host is constantly querying the position of the lid, which updates to "between" as soon as any motors are moving.

PR #366, opened 2022-07-28
feat(tempdeck-gen3): closed loop thermal control

fsinapi:
Adds basic support for closed loop thermal control. Tested on a board connected to the hardware of an old Tempdeck.

* Added GCode M18 to deactivate the thermal system
* Added GCode M104 to set the temperature target of the peltier
* Added GCode M301 to update the PID constants
* Added PID control of the peltiers based off of the target temperature
* Updated the upper limit of the Peltier PWM based off of testing (90% overworks the low side FET on the H-bridge)
* Added some Python utilities for testing closed loop control


Laura-Danielle (approved):
looks good to me

stm32-modules/tempdeck-gen3/firmware/thermal_control/thermal_hardware.c
```diff
@@ -39,7 +39,7 @@
 // Peltier drive circuitry cannot support lower PWM than 0.1
 #define MIN_PELTIER_POWER (0.1)
 // Peltier drive circuitry cannot support higher PWM than 0.9
```
sfoster1: update the comment too

sfoster1 (approved):
Looks good but let's update the comment.

PR #367, opened 2022-07-29
feat(thermocycler-gen2): lid heater tuning

fsinapi:
The lid heater originally dealt with the issue of integral windup by resetting the integral term of the PID controller once reaching the target temperature. Unfortunately, this results in a noticeable drop of the temperature from the target which the small Integral constant takes a long time to recover from.

This PR makes some changes to the control logic of the lid heater, opting to perform conditional integration (only integrating within the proportional band of the controller) and using a new set of default PID constants. The result is a much smoother approach to target temperatures, as seen in the attached image where the Lid Heater is targeting 105° using the new control system.

image


Laura-Danielle (approved):
nice!

PR #368, opened 2022-08-01
feat(thermocycler-gen2): add lid fan control

fsinapi:
The lid fan output driver is unused in the current Thermocycler Gen2 design, but incoming PCB tests will test that the output can be driven correctly. This PR adds a command (M903.D) to set the fans enabled or disabled, as well as the firmware required to enable or disable the lid fan LDO.

Tested on a board that the output is 0.0V on startup, can be set to 3.0V (the configured output of the LDO), and can be reset back to 0.0V with the M903.D command.


sfoster1 (approved):
Looks good to me!

PR #369, opened 2022-08-02
feat(thermocycler-gen2): light strip debug command

fsinapi:
Adds command M904.D, which is intended for integration with the ICT fixture to verify LED strip connectivity. When the command enables the LED Debug Mode, the lights stay on in the White setting until the command disables this mode. The debug mode takes precedence over any other statuses in order to ensure the setting on the LED's will be predictable.

Tested on a unit that, when the debug mode is enabled, setting plate temperature or having an error doesn't change the LED color.


sfoster1 (approved):
Looks good to me!

PR #372, opened 2022-08-03
feat(thermocycler-gen2, tempdeck-gen3): make assertions configurable

fsinapi:
In order to avoid any potential failures due to assertions in release code, this PR adds a CMake Option to configure whether assertions are enabled.

Tested on a Thermocycler board by adding in a definitely-failing assertion, and compiling both with and without the option. The firmware compiled with the option got stuck in an assert loop, and the other firmware image continued past it fine.


sfoster1 (approved):
Good call

PR #374, opened 2022-08-04
feat(thermocycler-gen2, startup): better hardfault handling

fsinapi:
This PR adds functionality for the Thermocycler Gen2 firmware to jump to the bootloader after any fatal exceptions that occur during runtime.

If assertions are enabled (presumably for debugging purposes), the firmware locks up permanently in the fault handlers. Otherwise, an inline assembly function adjusts the exception stack frame & the exception handling registers and then returns to the function for entering the bootloader.

This PR also adds a small change to the startup shim; the Program Status Register execution mode wasn't being checked, and if it was in an invalid state then an endless loop of hardfaults would ensue.

Tested the application fault handling by adding a null-dereference to the code. With the new code, the firmware ends up enumerating as a DFU device over USB.

### DIFF
--- a/stm32-modules/thermocycler-gen2/firmware/system/system_hardware.c
+++ b/stm32-modules/thermocycler-gen2/firmware/system/system_hardware.c
@@ -93,6 +93,9 @@ void system_debug_led(int set)
     HAL_GPIO_WritePin(GPIOE, GPIO_PIN_6, state);
 }
 
+// WARNING: it is VITAL that this function never returns! It has the chance to
+// be invoked via PC overwrite in an exception handler, and if the function
+// ever returns then catastrophic errors will ensue.
 void system_hardware_enter_bootloader(void) {
 
     // We have to uninitialize as many of the peripherals as possible, because the bootloader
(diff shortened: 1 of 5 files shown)


stm32-modules/thermocycler-gen2/firmware/system/system_hardware.c
```diff
@@ -163,3 +163,46 @@ void system_front_button_callback(void) {
...
+        "ldr r0, bootloader_address_const\n"
+        "str r0, [sp,#0x18]\n"
+        /* In case the PSR is in invalid state, force to user mode*/
+        "ldr r1, [sp,#0x1C]\n"
+        "and r1, r1, #0xFFFFFFF0\n"
+        "orr r1, r1, #0x10\n"
+        "str r1, [sp,#0x1C]\n"
+        /* Leave the exception handler */
+        "ldr lr,=0xFFFFFFF1\n"
+        "bx  lr\n"
```
sfoster1, line 205: i'm a little bit dubious about this because `system_hardware_jump_to_bootloader` isn't naked, but i guess it's a noreturn function anyway so nothing bad can really happen. maybe add a comment note there of like "do not ever change this function to occasionally return because it is called from a pc overwrite in an exception handler and it will go really badly"

sfoster1 (approved):
Looks good, although it does mean if anybody ever has the bright idea to make `system_hardware_jump_to_bootloader` return it will catastrophically break things

PR #375, opened 2022-08-08
chore(all): add image-bin and zip targets, speed up CI

fsinapi:
* Adds targets `-image-bin` and `-zip` for thermocycler-gen2 and tempdeck-gen3
* Parallelizes compilation for host & cross builds for all stm32 targets to speed up CI. This doesn't help much with the cross compilation since most of the time is actually spent in the linter, but it helps a bit with host compilation


sfoster1 (approved):
nice

PR #376, opened 2022-08-09
feat(thermocycler-gen2): slower plate lift 

fsinapi:
Added variable lid motor speeds and slowed down plate lift commands. The higher lid speed is good for the open/close user experience, but the plate lift was a little jerky at 125rpm.

Tested to make sure the firmware changes the velocity back to the higher one for opening/closing the lid.


Laura-Danielle (approved):
lgtm

PR #377, opened 2022-08-10
chore(startup): allocate more space for startup app

fsinapi:
Changes the linker settings for the startup app to allocate 32K, or the first 16 pages of memory. Also changes the linkage for the applications and accordingly changes the vector table relocation to match.

Tested that programming the new combined .bin still lets the startup app and the main application work fine. Linker output confirms that the region size for the startup app is increased and the main app is decreased.


Laura-Danielle (approved):
lgtm!

sfoster1 (approved):
Nice

PR #378, opened 2022-08-10
feat(all): add checksum to stm32 binaries

fsinapi:
### Overview
Calculates a CRC32 checksum for all STM32 module targets and injects it into the .hex target. All other images (excluding the base target) are derived from the .hex in order to ensure the integrity information is included in all targets.

The updates to the README file explain the details of the changes, but the short version is that a small chunk of flash (0x200 bytes) is now reserved for "integrity information." This information includes:
* A CRC32 of the rest of the firmware image (excluding the vector table)
* The number of bytes that the CRC32 includes
* The address where the calculation of the CRC begins (this should always be the same exact address, but is included just in case it needs to be variable)
* A null-terminated string containing the name of the module

None of this information is actually used by the startup app yet.

The CRC is calculated by a Python script. 
* The script emulates the STM32 CRC calculation from a .hex or .bin file:
  * For a .hex file, the script allocates a bytearray the same length as the firmware, initialized to 0xFF; populates it from the hex file; and runs a CRC32 using the same starting value as the STM32 (0xFFFFFFFF).
  * For a .bin file, the script skips an initial chunk of the file (based off of the starting location for where the CRC should begin) and runs a CRC on the rest of the file. The file must be generated with gaps filled to 0xFF.
  * In practice, the script is always run with a .bin input
* The docs for the library used _claim_ to use the same polynomial as the STM32 default (ethernet polynomial), but this can be changed to a custom implementation if it turns out to be incompatible in practice.

Tested by uploading to both Heater-Shaker and Thermocycler-Gen2, and then checking in the debugger that the integrity info is loaded at address 0x08008200 and that the application can still run fine.


scripts/calculate_checksum.py
```diff
@@ -0,0 +1,236 @@
...
+INITIAL_CRC = 0xFFFFFFFF
+
+class HexRecord:
+    """Represents a single record in a hex file."""
+    PATTERN = re.compile(r"^:([A-F0-9]{2})([A-F0-9]{4})(\d{2})([A-F0-9]*)([A-F0-9]{2})$")
+
+    _SEGMENT_ADDR_TYPE = 2
+    _LINEAR_ADDR_TYPE = 4
+
+    def __init__(self, line: str):
```
sfoster1, line 25: I think this works, but I will say that you could run `stm32-tools/gcc-arm-embedded/${PLATFORM}/bin/arm-none-eabi-objcopy -Iihex -Obinary --gap-fill=0xff` and then checksum that rather than writing your own hex decoder

stm32-modules/common/STM32F303/STM32F303RETx_FLASH.ld
```diff
@@ -67,8 +67,18 @@ SECTIONS
     . = ALIGN(4);
   } >FLASH
 
+  /* Contains firmware authentication information */
+  .integrity_info BLOCK(0x200) :
+  {
+    . = ALIGN(4);
+    /* placeholder for CRC */
+    LONG(0);
```
sfoster1, line 76: I _think_ you  could do something like 
```
.integrity_info BLOCK(0x200) :
  {
    . = ALIGN(4);
    _integrity_crc = .;
    LONG(0);
    _integrity_length=.;
    LONG(0);
} >FLASH
```

and then reference those symbols with extern in the actual code

sfoster1: in fact, you might be able to get the linker to put the overall length in there

sfoster1 (approved):
This looks great; there's some little things I think we could improve to make it a little more automatic, but this is gonna work.

scripts/calculate_checksum.py
fsinapi: [Pain](https://en.wikipedia.org/wiki/Sunk_cost).

stm32-modules/common/STM32F303/STM32F303RETx_FLASH.ld
fsinapi: One issue is that this is the linker for the runtime app, not the startup app; but adding this to the startup linker as a NOLOAD should do the trick I think?

scripts/calculate_checksum.py
fsinapi: Switched over to checking .bin files, and verified that the CRC and length stay the same between the two methods

sfoster1 (approved):
Looks good! Nice.

PR #379, opened 2022-08-11
chore(heater-shaker): update cmake with new F303 macros

fsinapi:
* Updated the subtree for `cmake/`
* Updated the Heater-Shaker firmware CMakeLists.txt to work with new F303 BSP macros


Laura-Danielle (approved):
lgtm!

PR #380, opened 2022-08-11
feat(all): changed startup app to a C program

fsinapi:
This PR rewrites the startup app as a C program, rather than an assembly program. The functionality is equivalent; the app checks that the vector table of the app exists, checks that it points to a Thumb function, and then jumps to the app. If those checks fail, it jumps to the built-in bootloader.

As a part of this change, the build process is a bit more complicated for each microcontroller target. The common folder for each MCU target:
* Provides a `startup_hal.h` file that provides any platform-specific values
* Links to the respective BSP driver target
* Provides the startup assembly file, which is the same as the one for the main applications

Tested on a heater-shaker, thermocycler-gen2 and a tempdeck-gen3. Confirmed that:
* With the entire image provided, the startup app can jump to the main application
* When the main application is deleted (or just the vector table is deleted), the startup app jumps to the bootloader and the microcontroller enumerates over USB as a DFU device


sfoster1 (approved):
This looks correct to me!

PR #381, opened 2022-08-12
feat(all): startup app CRC validation

fsinapi:
Adds functionality to the startup app to check the CRC of installed firmware images. If the CRC in the integrity region matches the CRC calculated by the startup app, it will boot into the app. If there's a mismatch, it will jump to the bootloader.

Changelist:
* Added debug targets for the startup apps
* Added explicit gap-fill of 0xFF to all bin targets in order to match the hex targets
* Added CRC check to startup app
* Removed unnecessary Initial Value from crc calc in python script
* Added clock/HAL initialization to the startup apps instead of relying on the clock config at startup. Also added deinitialization before jumping to the main app.

Tested on all STM32 modules by:
* Uploading the combined image-bin to the module and restarting to verify it jumps to the application
* Erasing a single page in the middle of the app to ensure the CRC check fails and the DFU bootloader activates
* Rewriting the base .hex (no startup app) through DFU and verifying the module starts up fine on reboot

### DIFF
--- a/scripts/calculate_checksum.py
+++ b/scripts/calculate_checksum.py
@@ -13,8 +13,6 @@ import zlib # for crc calc
 import sys
 from pathlib import Path
 
-INITIAL_CRC = 0xFFFFFFFF
-
 class HexRecord:
     """Represents a single record in a hex file."""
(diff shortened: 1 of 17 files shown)


sfoster1 (approved):
Looks good to me, though I for one always miss `~` and it might be a good idea to comment it or put it on its own line with a temporary.

PR #383, opened 2022-08-12
feat(heater-shaker): hardfaults jump to bootloader

fsinapi:
Copied fix from thermocycler-gen2. Tested by adding a divide-by-zero and confirming the module enumerates as a DFU device.


sfoster1 (approved):
I love this hack so much

PR #384, opened 2022-08-15
feat(all): added module name check to startup app

fsinapi:
Adds a check to the startup app to verify the name of the module that the firmware is intended for. This means that, in the case that a user erroneously sends the firmware intended for a different module (or completely random firmware that happens to pass the CRC check), the startup app will reject it.

Changelist:
- Updated the CMake for the startup app to create a unique startup app target for each module, in order to pass in module-specific values (i.e. the name)
- Added a check in the startup app to verify the name in the firmware matches the module it is intended for

Tested on heater-shaker and tempdeck-gen3 by:
- Uploading valid firmware, confirming the startup app boots to the main application
- Uploading firmware with the wrong name, confirming the startup app jumps to DFU
- Uploading firmware for a different module, confirming the startup app jumps to DFU

### DIFF
--- a/.github/workflows/heater-shaker.yaml
+++ b/.github/workflows/heater-shaker.yaml
@@ -62,6 +62,8 @@ jobs:
         run: cmake --build --preset cross --target heater-shaker-format-ci
       - name: 'Lint'
         run: cmake --build --preset cross --target heater-shaker-lint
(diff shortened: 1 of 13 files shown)


sfoster1 (approved):
Nice, good thinking

PR #385, opened 2022-08-15
feat(all): application backup and recovery

fsinapi:
This PR adds functionality to the startup app to recover from a bad firmware update (primarily intended for failures during transfer).
- The linkerfiles for the main applications have been edited to use _half_ of the available flash memory, leaving room for an identical copy of the firmware. 
- When the firmware is updated through DFU, it will ovewrite the _first_ image in memory. However, the second image will be left alone.
- On system restart, the startup app checks if the main image and the backup image exist:
  - If they both exist, and they match, just jump to the application firmware
  - If they exist and there's a mismatch, overwrite the backup with the main firmware and then jump to the application
  - If only one exists (i.e. there is an error with the integrity of **one** of the images), overwrite the faulty one and then jump to the main application
  - If neither image exists, jump to the DFU bootloader

In practice, any normal failure during the operation of dfu-util will only result in an extended reboot time while the system recovers its firmware from a backup, instead of bricking the module or getting it stuck in the bootloader.

Tested on heater-shaker and tempdeck-gen3 boards by replicating each of the above scenarios by manually erasing sectors, and then restarting the system to verify that it recovers in the intended fashion.

Startups where an image must be duplicated take a noticeable amount of time before jumping to the application, but normal startups take less than half a second.

### DIFF
--- a/stm32-modules/common/STM32F303/CMakeLists.txt
+++ b/stm32-modules/common/STM32F303/CMakeLists.txt
@@ -64,7 +64,8 @@ function(stm32f303_startup TARGET FW_NAME)
 
     target_sources(${TARGET} PUBLIC 
         ${CMAKE_CURRENT_FUNCTION_LIST_DIR}/startup_stm32f303xe.s
-        ${CMAKE_CURRENT_FUNCTION_LIST_DIR}/startup_system_stm32f3xx.c)
+        ${CMAKE_CURRENT_FUNCTION_LIST_DIR}/startup_system_stm32f3xx.c
+        ${CMAKE_CURRENT_FUNCTION_LIST_DIR}/startup_hal.c)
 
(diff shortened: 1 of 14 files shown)


sfoster1 (approved):
This looks good to me.

is the plan that dfu updates will only ever write the application slot, never the backup slot, and these mechanisms are there to get the backup slot updated?

fsinapi:
> is the plan that dfu updates will only ever write the application slot, never the backup slot, and these mechanisms are there to get the backup slot updated?

Yep, that's the idea - we provide the combined image-hex or image-bin file for updates, so dfu updates will only write to the startup region + the first application slot. 

Locking the backup slot is also possible (at least the MCU's we're using, since the G491 has two customizable regions and the F303 locks in groups of 2 pages).

PR #387, opened 2022-08-16
feat(all): lock startup region flash

fsinapi:
### Summary

Adds functionality for the startup app to lock its own flash region, preventing firmware updates from overwriting it. In practice, this also makes it pretty much impossible to update the startup app without erasing the entire chip - see comments in `startup_main.c` for details.

Adds a specialized function for each MCU target to update the flash region protection bytes, which is now the first step taken by the startup app. The option bytes work pretty differently in the F303 and G491 so these functions have to implement pretty much the entire process.

Tested on heater-shaker and tempdeck-gen3. Confirmed that the region gets locked, and if you try to unlock it (e.g. with a debugger and STM32CubeProgrammer) it gets locked again as soon as the system restarts. 

### Review Requests

Based on testing, it seems that moving the lock to be the last thing done by the startup app would actually let a debugger have enough time to update the option bytes, restart it to load the new bytes, and then halt the device before the startup app locks the region again. Would it make sense to move the lock action to the end of the startup app, and maybe even put a small wait in to make sure there's some time before the lock occurs?

### DIFF
--- a/stm32-modules/common/STM32F303/startup_hal.c
+++ b/stm32-modules/common/STM32F303/startup_hal.c
@@ -16,3 +16,46 @@ bool startup_erase_flash_pages(uint32_t start_page, uint32_t page_count) {
     }
     return false;
 }
+
+bool startup_lock_pages(uint32_t start_page, uint32_t page_count) {
+    // On the STM32F303, write protection is configured with a bitmask. Each
+    // bit in the bitmask represents 2 pages, and each page is 2k. This means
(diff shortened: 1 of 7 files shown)


sfoster1 (approved):
The point of the locking is to prevent the dfu app from writing that part of flash, right? So as long as the locking happens before we jump to the bootloader, it doesn't matter exactly where it occurs. So I think that yeah we should do it at the end.

PR #388, opened 2022-08-17
feat(all): added flash-erase targets

fsinapi:
Adds a target for each stm32 module to erase the flash contents. This is required because the startup app automatically locks its own flash, which means that a simple mass-erase won't work.

Simply clearing the option bits and then erasing isn't an option because, in the case that no firmware is loaded to either slot on the device, the startup app will re-lock before the host is able to halt it and start the erase. Therefore, the process is as follows:
- Enable the read protection, and then disable it to trigger a mass-erase (which affects even write-protected regions)
- Disable write protection
- On the STM32F303 target which clears the write-protect by erasing all of the option bytes, re-unlock the read protection

Tested on Heater-Shaker and Thermocycler-Gen2. Running the corresponding targets results in the board memory getting wiped and the write protection being fully disabled.

### DIFF
--- a/stm32-modules/common/STM32F303/CMakeLists.txt
+++ b/stm32-modules/common/STM32F303/CMakeLists.txt
@@ -92,6 +92,11 @@ endfunction()
 
 # Adds a target to clear an STM32F303-based module
 function(stm32f303_clear_target TARGET)
+    # Each step in this target is forced to succed (|| true) because, when
+    # the target reboots to load the option bytes, OpenOCD exits with a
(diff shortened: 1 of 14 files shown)


sfoster1 (approved):
This looks good, though I would love it if this could all be folded in to an openocd script - the multiple shell commands are kinda gross

fsinapi:
> I would love it if this could all be folded in to an openocd script - the multiple shell commands are kinda gross

Merging as-is after external discussion. Implementation isn't ideal but openOCD can't deal with the flash-option-reset process required by each step.

PR #389, opened 2022-08-17
chore(all): flash and debug targets cleanup

fsinapi:
Changelist:
* The flash targets for the modules were incompatible with some of the recent updates related to the startup app. This PR changes the flash targets to upload the .hex targets instead of the .elf targets. OpenOCD is able to discern which pages need to be erased before uploading, whereas before it was trying to erase the entire startup region as well.
* Added an `-image-flash` target for each module, which uploads the combined .hex with the startup app and the main application. This will fail if the startup region is locked, which is why the basic `-flash` target is still required
* Changed the generation of integrity info to actually update the base application output, which is what the debugger uploads to the module. This means that debug builds will be able to restart without issue.
* Changed the name of the combined image targets for the thermocycler-gen2 and tempdeck-gen3 to match what is generated for the heater-shaker, as well as the format that's expected by the Python side of things


sfoster1 (approved):
Looks good!

PR #390, opened 2022-08-18
fix(all): regenerate hex and bin targets more consistently

fsinapi:
There were some issues with the way dependencies between targets were defined.
* Some custom commands didn't have dependencies on all of the targets needed to cause regeneration (e.g. the hex targets need to depend on the base module target)
* Some custom targets were trying to define target-level dependencies in their definition, which is not allowed for some reason unbeknownst to me. This needs to happen in a separate `add_dependencies` call.

In general, we want to make the dependencies for custom _commands_ be custom targets when possible, because those will rerun unconditionally.


sfoster1 (approved):
Sounds good!

PR #391, opened 2022-08-18
fix(heater-shaker): kickstart for low rpms

pmoegenburg:
We've seen SetRPM(165) fail to start motor on some PVT units. As a precaution, with a factor of safety of roughly 2, this applies a SetRPM(300) for 1 second to SetRPM(<300) commands.

This also increases initial homing speed to 300 from 225rpm.


sfoster1 (approved):
Looks good to me

PR #392, opened 2022-08-18
fix(cmake): added component specifiers to stm32 install targets

fsinapi:
Installation step was trying to install everything in ./build-stm32-cross. We need to specify a component so that there are unique install targets.


sfoster1 (approved):
Very convenient

PR #393, opened 2022-08-19
fix(thermocycler-gen2): reduce thermocycler interrupt overhead

fsinapi:
The introduction of the front button LED flashing appears to have introduced issues with CPU overload when many actions are occurring at once. In particular, situations where the Thermocycler is moving the seal motor were causing other actions on the system to effectively stall. 

This PR adds in a few performance-oriented changes:
- Decreased the frequency of the seal motor interrupt. Testing a bunch of cycles of open/close with this lower speed shows that the acceleration curve still works correctly and the max velocity is the same.
- The Front Button LED callback only writes to the front button GPIO if there is an update to be made
- The timer for the front button LED is now wrapped into TIM7, which is used as the STM32 HAL Tick timer. The overhead of the FreeRTOS timer daemon with a 1ms period was excessive.
- Added flags to ensure max optimization on high-frequency interrupts (seal motor ticks and the TIM7 system tick handler)
  - This one is pretty minor, but our default configuration of MinSizeRel means some optimizations are disabled by default, and it barely affects flash size to have a select group of functions optimized for speed over space

Tested on an EVT unit. Before this update, the light blinking would become jittery and the ADC readings would sometimes start to fail after opening/closing the lid. After the update, dozens of cycles work fine.


sfoster1 (approved):
Looks good.

The choice of `MinSizeRel` and thus `-Os` was made for the original heater-shaker part that had a very small amount of stack. It might be worth trying to build with just `RelWithDebInfo` (and `-O2`) to see just how big it makes.

I do agree with leaving `-O3` to specific parts of the code though.

PR #394, opened 2022-08-19
fix(heater-shaker): force USB disconnect on startup

fsinapi:
The heater-shaker PCB has a hardwired pullup on the D+ line to indicate that it is a valid Full Speed device. Unfortunately, with the addition of the startup app, sometimes it takes ~10 seconds until the peripheral is actually initialized. During that time, the USB host expects a valid USB device to be connected, and will attempt enumeration.

This PR fixes the issue by forcing the D+ line on the USB low for a second, right before initializing the USB peripheral. This forces the USB host to re-enumerate the bus even if it failed previously, and it successfully detects the heater-shaker.

Tested with by my Mac and with an OT-2 by updating from v1.0.1 to v1.0.2. When the device restarts, it is detected as a USB CDC device correctly once it starts the application firmware.


sfoster1 (approved):
That'll (horrifyingly) do it. Good find!

PR #395, opened 2022-08-19
chore(heater-shaker): save zip output when creating release

fsinapi:
Store the output of the heater-shaker-zip target as an artifact in github actions. Will test by triggering the workflow on this branch.

**EDIT** tested by manually running the github action, and confirmed that it stores the zipfile artifact for download.


sfoster1 (approved):
🤐

PR #396, opened 2022-08-23
chore(thermocycler-gen2): deploy releases to aws

fsinapi:
* Added explicit naming of artifacts for heater-shaker releases
* Added workflow to deploy TC2 firmware to AWS

- [x] Test that this succeeds when run manually against this branch
- [x] Does the target directory in AWS have to be set up manually before this can succeed?


sfoster1 (approved):
Looks good. The target directory does not have to be set up because directories on s3 aren't real they're just common prefixes of object names

PR #397, opened 2022-08-23
feat(thermocycler-gen2): overshoot/undershoot tuning

fsinapi:
This PR makes some big changes to how the overshoot/undershoot behavior is handled on the Thermocycler Gen2.

* Default Kp has been increased quite a bit, results still have minimal unwanted overshoot/oscillation
* Added a factor of 2x when looking for the Proportional Band to allow slightly more integrator windup before the P term starts to damp
* The amplitude of the goal overshoot/undershoot is now 2º for every 50µL of liquid
* Instead of a 10-second "hold period," the TC2 switches its target to the user-programmed temperature as soon as the overshoot/undershoot target is reached.
  * There is a 1º margin here, which is accounted for by increasing the overshoot/undershoot magnitude by a flat 1º. This ensures that the PID controller errs on the side of aggressive ramping to the temperature, rather than requiring a separate tune.
  * The center channel gets an extra boost after testing showed that the temperature in the wells responds slower than other wells.
  * The block is only considered "at" this target temperature once all three columns independently reach their overshoot/undershoot target
* Overshoot/undershoot is **not** applied in the following circumstances:
  * Sample volume is less than 20µL
  * Current temperature is less than 5º away from the target temperature
  * Target temperature is below ambient room temperature
* Hold time countdown is active when the following conditions are met:
  * The plate temperature has reached the 2º threshold for the overshoot/undershoot, and is now in the "steady state" phase targeting the programmed target
  * The plate temperature is within 2º of the programmed target temperature

Merging this PR pends on some test data using an external temperature probe.

Some captures of test data with various programmed volumes:
image

image

image


sfoster1 (approved):
This looks good to me, nice graphs!

sfoster1 (approved):
Changes look good to me.

PR #398, opened 2022-08-23
fix(heater-shaker): detect faulty plate latch sensors

pmoegenburg:
Safeguards failure mode seen with latest firmware on outdated EVT units, where units' end stop sensor logic was flipped.


sfoster1 (approved):
Nice!

PR #400, opened 2022-08-24
fix(all): mark HAL structs in RESET state

fsinapi:
Thermocycler-Gen2 firmware was exhibiting a bug where the seal motor interrupt would not fire if the following happened:
* `dfu` command sends the MCU to the system memory
* A dfu leave command causes the MCU to jump back to the application memory _without resetting power_
* A button press or a GCode starts a seal motor movement 

Running the firmware with a debugger and replicating this setup showed that the HAL Structure for the seal motor timer had an uninitialized State variable, which was set to `READY` instead of `RESET`. In this case, the HAL does not call the MSP initialization function and thus the interrupt for the timer is never enabled. This seems to be a side effect of the fact that the RAM does not get reset after leaving the system memory bootloader, because in any other situation (i.e. after a power cycle) the RAM is fully cleared.

This PR fixes the issue by manually setting the `State` variable in any relevant HAL structures to the `RESET` setting, which ensures that the initialization functions will configure the hardware for each peripheral.

### DIFF
--- a/stm32-modules/heater-shaker/firmware/heater_task/heater_hardware.c
+++ b/stm32-modules/heater-shaker/firmware/heater_task/heater_hardware.c
@@ -146,6 +146,7 @@ static void adc_setup(ADC_HandleTypeDef* adc) {
 }
 
 static void tim_setup(TIM_HandleTypeDef* tim) {
+    tim->State = HAL_TIM_STATE_RESET;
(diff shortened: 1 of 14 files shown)


Laura-Danielle (approved):

sfoster1 (approved):
Well that's awful.

PR #401, opened 2022-08-25
fix(heater-shaker): differentiate sync and async error responses

pmoegenburg:
Currently, all gcode messages sent by firmware contain the ack "OK". The hardware controller class doesn't have a way of differentiating responses to gcode commands from asynchronous error messages. This prepends the latter with an identifier "async", the firmware-side fix to RET-1178. The api-side fix is Opentrons/opentrons #11400.

Examples:
- successful response: `M3 OK`
- unsuccessful response: `ERRxxx:...`
- asynchronous error message: `async ERRxxx:...`

To-do: implement in TC2 and TD3 firmware

### DIFF
--- a/stm32-modules/heater-shaker/tests/test_errors.cpp
+++ b/stm32-modules/heater-shaker/tests/test_errors.cpp
@@ -12,12 +12,14 @@ SCENARIO("testing error writing") {
                 errors::write_into(buffer.begin(), buffer.end(),
                                    errors::ErrorCode::USB_TX_OVERRUN);
             THEN("the error is written into the buffer") {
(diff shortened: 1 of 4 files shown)


sfoster1 (changes requested):
Not sure I get what this is doing. Would you mind putting a larger description of the point of the changes in the PR body?

If I understand correctly, responses to normal gcodes will now look like, for instance,
`gcode response M105 C10 T15 OK\r\n` ?

sfoster1 (approved):
Looks good to me!

fsinapi (approved):
Nice

PR #403, opened 2022-08-26
feat(heater-shaker): rpm data filtering

fsinapi:
The Heater-Shaker uses a Hall effect sensor to detect the current RPM of the motor. The actual speed of the motor _as measured by an external tachometer_ is consistent and stays within the device's specified accuracy spec, but the speed as reported by the internal sensor varies severely. Therefore, it is reasonable to state that the values reported by the sensor work for the application and allow the motor control driver to control the motor accurately. However, it makes sense to create an RPM that is more representative of the motor speed over USB.

This PR adds in a circular buffer to keep track of the last 32 readings from the Hall sensor. When any of the higher level code needs to get the current RPM, it is calculated by averaging the contents of the buffer. The size of the filtering buffer was chosen by testing a few values - smaller values result in higher variability. 

The new script `rpm_data.py` was used to collect the min & max RPM reported at setpoints for both the original firmware and the updated firmware. The results are summarized in the table below; the updated firmware drastically decreases the variability of these numbers.

Setpoint (rpm) | Min rpm (No Filter) | Max rpm (No Filter) | Min rpm (With Filter) | Max rpm (With Filter)
-- | -- | -- | -- | --
200 | 126 | 246 | 180 | 210
1000 | 978 | 1020 | 984 | 996
2000 | 1968 | 2034 | 1992 | 1998
3000 | 2958 | 3054 | 2994 | 3000

### DIFF
--- a/stm32-modules/heater-shaker/firmware/motor_task/mc_tasks.c
+++ b/stm32-modules/heater-shaker/firmware/motor_task/mc_tasks.c
@@ -279,10 +279,22 @@ void TSK_MediumFrequencyTaskM1(void)
 {
   State_t StateM1;
   int16_t wAux = 0;
+  
+  static volatile uint8_t last_SpeedFIFOIdx = 0;
+  static volatile int16_t last_wAux = 0;
 
   bool IsSpeedReliable = HALL_CalcAvrgMecSpeedUnit( &HALL_M1, &wAux );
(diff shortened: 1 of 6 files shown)


sfoster1 (approved):
[hands shaking trying to prevent suggesting a million micro optimizations about bitmasks and remainders] Looks good!

PR #405, opened 2022-08-29
fix(thermocycler-gen2): update LED color to match Gen1

fsinapi:
The white color on the Gen1 is actually generated by turning on the Red, Green, and Blue LED's instead of the White LED. This results in a much different color than the "soft white" from the actual White LED.

This PR updates the shade of white on the TC2 to match the TC1.


Laura-Danielle (approved):
🛋️

PR #407, opened 2022-08-30
feat(thermocycler-gen2): decrease transition time between close temperatures

fsinapi:
Tests with standard PCR cycles of [94º, 70º, 72º] showed a very long transition time from 70 to 72. The root of the cause seems to be that the PID integral term gets wiped out when setting a new target. If two targets are very close, the P term will only result in a small amount of power, while the I term will take a significant amount of time to saturate. 

It seems reasonable to assume that the steady-state integral windup should be similar for two temperatures within a few degrees of each other. This PR removes the PID reset call when a new target temperature is very close to the last target temperature. Testing shows that the curve between 70 and 72 is faster with this change (charts below, using an external multiprobe for temperature data).

image

image


sfoster1 (approved):
Makes sense to me, good data

PR #408, opened 2022-08-31
feat(tempdeck-gen3): add EEPROM driver

fsinapi:
This PR adds a driver for the EEPROM IC on the Tempdeck Gen3 PCB. The EEPROM is on the same I2C bus as the external ADC, and is used to store thermal offset constants just like on the Thermocycler-Gen2. Fittingly, most of this code is carried over from TC2.

One caveat here is that the EEPROM is DNP'd on the Revision 1 PCB's. A ticket has been added to test integration once the Rev2 PCB arrives, but in the meantime the code allows users to set the constants and read them back fine.

Tested on a Rev 1 PCB by setting and getting the constants over USB. They persist until the power cycles, but are reset on startup.


sfoster1 (approved):
Looks good to me!

PR #409, opened 2022-09-07
feat(tempdeck-gen3): add internal ADC driver

fsinapi:
This PR adds code to read the internal ADC on the Tempdeck Gen3 board, used for the Peltier Current readback. This reports the total current drawn from the 36V rail to the peltier drive subsystem.

The ADC conversion is kicked off before the delay for each batch of thermistor ADC measurements, which provides more than enough time for the conversions to complete. When the thermistor task reads back the ADC readings, it calculates an average with the minimum and maximum ADC readings discarded for increased accuracy.

Because of the heavy filtering of the peltier control output, the timing of these ADC measurements relative to the thermal control PWM does not need to be considered.

Tested by checking the readings buffer and the Thermal Task's average value with a debugger while the peltiers were activated. At a power of 25%, the ADC value reported was 112 counts. Converting based off of the PCB schematic gives an overall current of 180.5mA for that value.


sfoster1 (approved):
Nice

PR #410, opened 2022-09-08
feat(tempdeck-gen3): get thermal power command

fsinapi:
Adds a command M103.D to get the power of the thermal system. The command returns the actual PWM of the peltiers & fan, as well as the current measured on the Peltier subsystem in milliamperes. The ADC readings are converted to current based off of the schematic of the Non Form Factor board, this may change with the next rev of the board.


sfoster1 (approved):
Looks good to me.

PR #412, opened 2022-09-12
feat(tempdeck-gen3): get fan RPM measurements

fsinapi:
Adds functionality to read the RPM of the fan on the tempdeck-gen3. Uses TIM15 in Input Capture Mode to read pulses. See `tachometer_hardware.c` for detailed explanation.

Tested with a Thermocycler Gen2 fan since the current fans we have don't have the tachometer installed. RPM measurements match that of the TC2 at given PWM values, and turning the fan properly shows the RPM drop down to 0.


sfoster1 (approved):
Looks good to me!

PR #413, opened 2022-09-14
feat(tempdeck-gen3): support multiple i2c buses

fsinapi:
The I2C code was originally lifted from Thermocycler Gen2, where only one I2C bus is used. In the Tempdeck we are going to have two buses, with one bus dedicated to the LED board to avoid any interference with the thermal I2C bus. This code sets up the framework to allow the second bus to be seamlessly added when bringing up the LED board.

* Moves the I2C code from `thermistor_hardware` to `i2c_hardware`
* Each I2C bus gets its own struct with a unique buffer, semaphore, and task notification for signalling the end of transmission.

Tested on hardware that hooking up a thermistor still shows the correct temperature being polled.


sfoster1 (approved):
Looks good to me!

PR #414, opened 2022-09-14
fix(thermocycler-gen2): update async error responses

pmoegenburg:
Prepends async error responses with `async`


fsinapi (approved):
Looks good! While we're at it we should update the default setting for `ALLOW_ASYNC_ERRORS` to `ON` in `thermocycler-gen2/src/CMakeLists.txt`, right now the code defaults to not sending async error messages.

I also think we should hold off on merging this until the monorepo driver is updated to match.

PR #416, opened 2022-10-12
fix(thermocycler-gen2): front button ESD filtering

fsinapi:
ESD testing showed that voltage spikes can generate spurious inputs on the front button GPIO. This PR adds some logic to filter out short spikes such as this, while allowing actual button presses.

- If the button is not pressed after the initial debounce period, we now assume it was a voltage spike rather than a valid press.
- Decreased debounce time to 50ms to avoid filtering out valid presses. This is still a generous debounce period based on testing, and especially will be expected to filter out transient ESD spikes.


pmoegenburg (approved):
LGTM

sfoster1 (approved):
Nice

PR #417, opened 2022-10-13
feat(thermocycler-gen2): optimize seal logic for DVT boards

fsinapi:
Closes RET-1209 and RET-1228

The seal motor has two limit switches on the EVT units, but those switches share a single GPIO. On DVT and later thermocyclers, there are independent lines for each switch. This means that the control logic for the seal motor during lid open/close movements can be simplified.

- When given an open/close lid command, check whether both switches are triggered and return an error if so.
- When the next step is retracting or extending the seal, skip the step if the correct limit switch is already triggered
- When the next step is a "backoff" from the limit switch at the end of a movement, skip the step if we are on a board with multiple GPIO for the seal switches.

Tested that the lid still works as expected on an EVT unit (and still won't move when the seal limit switches are triggered), and tested on a DVT unit that the motor task uses the new logic and that the behavior works as expected.

### DIFF
--- a/stm32-modules/include/thermocycler-gen2/firmware/motor_policy.hpp
+++ b/stm32-modules/include/thermocycler-gen2/firmware/motor_policy.hpp
@@ -177,6 +177,16 @@ class MotorPolicy {
     [[nodiscard]] auto seal_read_retraction_switch() const -> bool;
 
     /**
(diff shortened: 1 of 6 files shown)


sfoster1 (approved):
This looks good to me.

PR #419, opened 2022-10-27
feat(tempdeck-gen3): update thermal control

pmoegenburg:
Different PID values were determined to be necessary for heating versus cooling. Step response PID tuning utilized the thermal_profiling script and subsequent thermal testing utilized the test_temperature script. Test results can be found here: https://docs.google.com/spreadsheets/d/1L9o-yjL69LEtBNwnRX4wdr1X6_NRCenyyLH5wbUfED8/edit#gid=0

Closes RET-1186


fsinapi (approved):
This looks good, nice data!

The one thing I'd like is a way to configure the PID values - we have a command to set them, but now any custom values will get overwritten by the SetTemperatureMessage handler. It doesn't have to be in this PR (can just set up a ticket to do this later), but I'd like to have a flag in the thermal task that tracks whether a user set custom PID values and then keeps them indefinitely.

sfoster1:
Let's do that but not right now - this one's gonna be on the backburner for a bit

PR #420, opened 2022-10-28
fix(thermocycler-gen2): add ack to error responses

pmoegenburg:
Same fix as #348


fsinapi (approved):
LGTM once you format it! Gave this a test against the robot software by modifying the firmware to throw an error when setting lid temperature. Confirmed that the error is read and the `OK\n` is all interpreted as the ack and dropped.

PR #421, opened 2022-11-01
feat(thermocycler-gen2): add script to test temperature divergence

fsinapi:
This PR adds a script for simple testing of temperature divergence. It accepts an argument for a temperature and logs all of the thermal information until the script is cancelled. The script tries to stick to minimal dependencies so it can be run on any target with Python3 and the pyserial library.

Tested that this works + logs information from a TC2 running the latest firmware.


sfoster1 (approved):
looks good to me though you can specify `re.I` to ignore case in the regex rather than leaving off the `t`

PR #423, opened 2022-11-04
fix(thermocycler-gen2): throttle simulator periodic data

fsinapi:
When running the TC Gen2 simulator in opentrons emulation, errors would bubble up indicating message queue overflow. This seems to be due to the periodic data thread running too fast for its own good and filling the other task's queues with tons of temperature update messages.

This PR adds atomic flags for the periodic data thread to track when it sends a temperature update to another thread. The lid and plate threads will clear their respective flags whenever a temperature conversion message is consumed.

Tested on my machine that the simulator still lets you set a temperature target and simulates the temperature reaching that target on both the lid and plate.


DerekMaggio (approved):
Tested with emulation and verified that errors about the input queue being full no longer happen

PR #425, opened 2022-11-15
chore(cmake): pull latest cmake-utils

fsinapi:
Just a subtree-pull to get the codecov stuff


sfoster1 (approved):
Looks pretty good but uh what's the `bar` tree there doing

fsinapi:
> Looks pretty good but uh what's the `bar` tree there doing

It's some example code from the repo I pulled this module from. It's included because I pulled _that_ whole repo into `cmake-utils` as a subtree to preserve the license and such

PR #426, opened 2022-11-15
feat(tests): add code coverage tracking

fsinapi:
This PR adds support to generate code coverage information for the firmware.

* This takes advantage of a cmake module that easily sets up a target to generate `lcov`  coverage info
* The lcov output is converted to an xml file with a python module to be compatible with `codecov`
* The resulting code coverage XML is pushed to codecov and displayed as a message with the PR

This PR also addresses an issue with a failing unit test that had not been failing in CI before. Weird!

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,8 @@
+
+comment:
+  # add flags to `layout` configuration to show up in the PR comment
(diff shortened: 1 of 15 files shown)


sfoster1 (approved):
Nice, love to get covered

PR #427, opened 2022-11-17
fix(thermocycler-gen2): always disable if thermistors are bad

fsinapi:
The TC2 firmware includes protection against shorted & disconnected thermistors. The status of each thermistor is stored as a bit flag for both the Thermal Plate and Lid Thermistor tasks. When a flag goes from 0 to 1, the firmware detects it and makes sure to disable any related hardware to prevent a dangerous thermal error case.

Unfortunately, some commands were resetting the task state to IDLE. Because the flags were only checked if they _changed_, the thermal tasks would then happily let users set a target temperature. If a thermistor is broken, the thermal control will probably lead to a runaway overheat situation.

Steps to reproduce:
* Turn on TC with a faulty  resistor (e.g. put a 2.2Ω resistor in place)
* Send M140 to set the lid on or M104 to set plate on --> get an error back
* Send M18 to deactivate everything
* Send M140 or M104 again --> this time, the lid or plate will start heating

After this PR, the last step will return an error and NOT control the temperature.

### DIFF
--- a/stm32-modules/include/thermocycler-gen2/thermocycler-gen2/lid_heater_task.hpp
+++ b/stm32-modules/include/thermocycler-gen2/thermocycler-gen2/lid_heater_task.hpp
@@ -160,7 +160,8 @@ class LidHeaterTask {
         auto old_error_bitmap = _state.error_bitmap;
         auto current_time = Milliseconds(msg.timestamp_ms);
         handle_temperature_conversion(msg.lid_temp, _thermistor);
-        if (old_error_bitmap != _state.error_bitmap) {
+        if ((old_error_bitmap != _state.error_bitmap) ||
+            (_state.error_bitmap != 0)) {
             if (_state.error_bitmap != 0) {
                 // We entered an error state. Disable power output.
                 _state.system_status = State::ERROR;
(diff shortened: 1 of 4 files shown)


stm32-modules/include/thermocycler-gen2/thermocycler-gen2/lid_heater_task.hpp
```diff
@@ -328,7 +330,9 @@ class LidHeaterTask {
             messages::DeactivateAllResponse{.responding_to_id = msg.id};
 
         static_cast<void>(policy.set_heater_power(0.0F));
-        _state.system_status = State::IDLE;
+        if (_state.system_status != State::ERROR) {
```
Laura-Danielle, line 333: why do we need to even set the system status in this message ?

sfoster1 (approved):
Looks good to me.

Laura-Danielle (approved):
Fix looks great to me. I am curious as to why the visit functions that receive messages are responsible for controlling the system status.

fsinapi:
> I am curious as to why the visit functions that receive messages are responsible for controlling the system status.

Because that's when the tasks receive their temperature data. There are separate tasks that run on a periodic basis to read the thermistors over I2C, and they send messages (`ThermalPlateTempReadComplete` and `LidTempReadComplete`) at fixed frequencies, thus implicitly driving the control frequency of the thermal control tasks. The `visit` method for those two messages is the point where the task is running the actual control loop, so the state gets updated if e.g. an error arises during a closed-loop temperature control

stm32-modules/include/thermocycler-gen2/thermocycler-gen2/lid_heater_task.hpp
fsinapi: If the status is CONTROLLING here, we need to reset it to IDLE in order to deactivate closed loop control. If it gets left as CONTROLLING then the task will keep holding temperature.

PR #428, opened 2022-11-17
feat(thermocycler-gen2): ignore thermistor uniformity checks below 7.5 degrees

fsinapi:
A potential use case of the thermocycler is holding for an extremely long time (multiple days) at 4 degrees at the end of an experiment. There are some cases where condensation from the well plate can result in skewed thermistor readings over this long time period, and the thermocycler may fail the thermistor uniformity check and disable itself. As long as all temperatures are below 8º, it is preferable for the TC to continue holding rather than stop itself due to the error. 

After discussions with the hardware team, we settled on a threshold of 7.5 degrees to ensure that, if thermistors in a column are diverging (but holding their average to 4º), the TC will shut off before the lower value reaches sub-zero temperatures.


sfoster1 (approved):
lgtm

PR #432, opened 2022-12-13
fix(thermocycler-gen2): handle systick overflow in thermal control tasks

fsinapi:
The thermistor data update messsages contain a timestamp in milliseconds, derived from the systick timer. Systick overflows once every ~50 days, and the Plate Control and Lid Control tasks were not properly handling this overflow.

This PR adds handling to make sure that, if the time delta between updates is negative (aka overflowed), the tasks compensate appropriately.


sfoster1 (approved):
chrono makes this all so much easier

PR #433, opened 2022-12-13
fix(thermocycler-gen2 script): fix order of thermistor data

fsinapi:
The script `hold_temp.py` prints out a csv header, and then was printing out the actual data in a different order. This PR just swaps the order of the data to match the header.


sfoster1 (approved):
lgtm

PR #435, opened 2023-03-03
feat(thermocycler-gen2): add filtering to peltier output

fsinapi:
This PR adds a slight amount of filtering to the output of the peltiers for the TC Gen2 plate control.

Transitioning immediately from full cooling to full heating (or vice-versa) may put undue stress on the peltier elements due to the thermal shock. Having a small amount of "dead time" is recommended to lessen that thermal shock when transitioning between directions; in this implementation, adding a constraint on the maximum power delta (such that the total transition time between full cooling and full heating takes a few control cycles) is intended to add a transition period to help provide a transition period.

Closed loop thermal performance will be verified through hardware testing before merging in order to ensure that this filtering will not affect the performance of the thermal control.


sfoster1 (approved):
This looks good to me if the hardware tests show equivalent functionality. I'd be a little worried about the PIDs winding while the power is clamped, but if they don't then they don't.

Laura-Danielle (approved):
lgtm

PR #436, opened 2023-03-08
fix(heater-shaker): bifurcated homing speed

pmoegenburg:
# Overview

This bifurcates final homing speed for units with old and new solenoids. We've seen repetitive stress failures of the old solenoid, so this reduces the speed of the main motor during homing from 300rpm to 225 for those units. Bifurcation is determined by serial number, with new solenoids in units made after the 2022-11-30 7th unit.

To ensure homing doesn't fail to start due to static friction, the main motor spins at 300rpm for 1 second before shifting to its final homing speed and the solenoid engages.

# Test Plan

- [x] Verify homing speed 300rpm with serial numbers after 2022-11-30_07
- [x] Verify homing speed 225rpm with serial numbers before 2022-11-30_07
- [x] Verify homing speed 225rpm with invalid serial numbers

# Changelog

# Review requests

# Risk assessment

Low, final homing speed determination via serial number fails safely, and kickstart process has been thoroughly tested previously.

### DIFF
--- a/stm32-modules/include/heater-shaker/heater-shaker/motor_task.hpp
+++ b/stm32-modules/include/heater-shaker/heater-shaker/motor_task.hpp
@@ -95,9 +95,14 @@ class MotorTask {
     static constexpr const uint16_t STARTUP_HOMING_WAIT_TICKS =
         200;  // needed to ensure motor setup complete at startup before homing
     static constexpr const uint16_t MOTOR_START_WAIT_TICKS = 1000;
+    static constexpr const uint16_t SOLENOID_ENGAGE_WAIT_TICKS = 1000;
     static constexpr const uint16_t POST_HOMING_WAIT_TICKS =
         500;  // needed to ensure motor control deactivated before subsequent
               // SetRPM commands
+    static constexpr const uint8_t SERIAL_NUMBER_PREFIX_OFFSET =
+        5;  // skips first 5 chars ("HSV01")
+    static constexpr const uint32_t SERIAL_NUMBER_SOLENOID_SWITCH_TIMESTAMP =
+        2022113007;  // switched from old to new solenoid on 2022-11-30 7th unit
 
   public:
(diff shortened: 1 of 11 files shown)


stm32-modules/include/heater-shaker/heater-shaker/motor_task.hpp
```diff
@@ -360,9 +366,37 @@ class MotorTask {
         }
     }
 
+    // unit test this
+    // comments
+    auto set_homing_speed(
+        std::array<char, SYSTEM_WIDE_SERIAL_NUMBER_LENGTH> serial_number)
+        -> void {
+        int result{};
+        auto [ptr, ec]{std::from_chars(serial_number.begin() + 5,
```
sfoster1: Oh wow I've never seen this syntax before and I love it. for others in the same boat, this is [structured binding](https://en.cppreference.com/w/cpp/language/structured_binding) e.g. destructuring, and has been in since c++17!

stm32-modules/include/heater-shaker/heater-shaker/motor_task.hpp
```diff
@@ -360,9 +366,37 @@ class MotorTask {
...
+                                       serial_number.end(), result)};
+        if (ec == std::errc()) {
+            if (result > 2022113007) {  // make define. is this always true?
```
sfoster1: this logic leaves the `else` for this if out - could make it `if (ec == std::errc() && result > 2022113007)` and it'd be fine

sfoster1 (changes requested):
Looking solid, agree with the stuff you want to do, couple extra things I didn't see comments for

fsinapi (commented):
This looks really solid to me! Only thing is I'd like to see one of the CI tests set the serial number to before & after the solenoid switch timestamp and check the homing speeds.

fsinapi (approved):
Nice!!!

sfoster1 (approved):
Looks good to me, nice work!

PR #437, opened 2023-03-08
feat(tempdeck-gen3): support pcb rev2

fsinapi:
This PR makes hardware configuration changes to port the Tempdeck Gen3 firmware to the Revision 2 board. For the sake of expediency, there is __no backwards compatibility__ to the non-form-factor board.

The following changes were made:
- Moved from I2C2 to I2C1 for ADC/EEPROM
- Moved ADC alert pin to C12
- Moved tachometer input to TIM17 CH1
- Moved TEC PWM output to TIM2 CH3 and CH4
- Moved EEPROM Write Protect to C11
- Moved Current Sense input from ADC2 CH2 to ADC1 CH5 on pin B14
- Moved Heartbeat LED to A3
- Updated the thermistor model to the model that will be used on the final TD3, the same as the Thermocycler Gen2

Tested on hardware that:
* The ADC still reads the thermistors, and connecting the new thermistor model correctly reads ambient temperature
* Heating/cooling the peltiers goes in the right direction
* Fan can be controlled with power %, and the tachometer data is updated as the fan speeds up/slows down
* Heartbeat still works
* Current sense responds to peltier power changes - _the transfer function is not updated yet!_
* USB comms still work fine
* Tried uploading firmware using the third ADC channel to make sure that works correctly as well (will be the heatsink thermistor)

Followup PR(s) will also:
- Modify control to use two plate thermistors instead of one
- Update the transfer function for current measurement based on the updated schematic

### DIFF
--- a/stm32-modules/include/tempdeck-gen3/tempdeck-gen3/thermal_task.hpp
+++ b/stm32-modules/include/tempdeck-gen3/tempdeck-gen3/thermal_task.hpp
@@ -498,7 +498,7 @@ class ThermalTask {
     Queue& _message_queue;
     Aggregator* _task_registry;
     ThermalReadings _readings;
-    thermistor_conversion::Conversion<lookups::NXFT15XV103FA2B030> _converter;
+    thermistor_conversion::Conversion<lookups::KS103J2G> _converter;
     Fan _fan;
     Peltier _peltier;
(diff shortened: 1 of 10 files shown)


sfoster1 (approved):
okay, looks good!

PR #439, opened 2023-03-08
feat(tempdeck-gen3): use two plate thermistors

fsinapi:
The rev2 PCB supports two plate thermistors and a heatsink thermistor. This PR modifies the allocation of the ADC channels to match the PCB updates.

Tested on a PCB with all 3 thermistors attached. The Python driver reports all three thermistors for command M105.D. All thermistors report ambient temperature, and pressing one between your fingers raises the reported temperature as expected.

Closes RET-1329


sfoster1 (approved):
Looks good to me!

PR #440, opened 2023-04-28
feat(thermocycler-gen2): add plate lift wiggling

fsinapi:
This adds a "wiggling" motion to the Thermocycler plate lift action in order to dislodge plates more gently.

* The speed for plate lift is decreased a bit to reduce the sudden force on the plate
* The plate lift moves up a little bit, retreats a bit, and then moves up a little more; it does this a few times until the plate is certainly dislodged
* The lifter runs to its limit and goes back to the normal "open" position as usual

The issue that this addresses does not happen on _every_ thermocycler. On affected units, the original plate lifter implementation would "pop" the plate very suddenly, potentially spilling liquid.

With the new implementation, you can leave a thermocycler closed for a long time, cycle temperatures, open the lid, and run a plate lift and the popping is gone (or, in the worst cases, very highly dampened)

### DIFF
--- a/stm32-modules/include/thermocycler-gen2/thermocycler-gen2/motor_task.hpp
+++ b/stm32-modules/include/thermocycler-gen2/thermocycler-gen2/motor_task.hpp
@@ -118,7 +118,7 @@ struct LidStepperState {
         motor_util::LidStepper::angle_to_microsteps(-30);
     // Velocity for plate lift actions. This provides a smoother lifting
     // action than the default open/close velocity.
-    constexpr static double PLATE_LIFT_VELOCITY_RPM = 80.0F;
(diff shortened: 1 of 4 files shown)


stm32-modules/include/thermocycler-gen2/thermocycler-gen2/motor_task.hpp
```diff
@@ -1281,7 +1292,41 @@ class MotorTask {
                 // TODO(Frank, Mar-7-2022) check if the lid didn't make it in
                 // all the way
                 break;
+            case LidStepperState::Status::LIFT_NUDGE:
+                // std::ignore = policy.lid_stepper_set_rpm(
```
sfoster1: remove commented-out code

sfoster1 (approved):
Looks good, let's pull the commented code though

PR #441, opened 2023-05-17
feat(tempdeck-gen3): support new eeprom

fsinapi:
Closes RET-1224

The Rev 3 version of the Tempdeck-Gen3 PCB uses a different eeprom than the previous revisions. The new EEPROM, STMicroelectronics' `M24128`, uses 64-byte pages instead of the previous model's 8-byte pages. Therefore, we are able to store the entirety of the offset data in a single page, reserving the rest of the space for potential future data.

This PR makes the following changes:
* The EEPROM model is updated to a new `M24128` class
* The EEPROM contents have been compressed into a single `__packed` struct that fits into a 64-byte page
* The offset values are not read back from the EEPROM after the first initialization
* Fixed a bug that the python driver wouldn't send offset calibration values of `0`

Tested on a Tempdeck-Gen3 Revision 3 PCB. Tried setting the value for each constant to a nonzero number, and then reset the board by removing power. After powerup, the constants could be read out over USB with the same values. Repeating the same test with different numbers was successful.

Given that the EEPROM was unpopulated on all prior revisions anyways, this PR does _not_ attempt to maintain backwards compatibility.

Also note that the offset constants aren't actually _used_ for anything yet. That will be a future PR for a different ticket.

### DIFF
--- a/stm32-modules/include/common/core/m24128.hpp
+++ b/stm32-modules/include/common/core/m24128.hpp
@@ -98,9 +98,9 @@ class M24128 {
         uint16_t start_addr = page * PAGE_LENGTH;
         // MSB is first, followed by LSB
 
-        //NOLINTNEXTLINE(cppcoreguidelines-avoid-magic-numbers)
+        // NOLINTNEXTLINE(cppcoreguidelines-avoid-magic-numbers)
         _buffer.at(0) = static_cast<uint8_t>((start_addr & 0xFF00) >> 8);
-        //NOLINTNEXTLINE(cppcoreguidelines-avoid-magic-numbers)
+        // NOLINTNEXTLINE(cppcoreguidelines-avoid-magic-numbers)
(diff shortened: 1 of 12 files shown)


stm32-modules/include/common/core/m24128.hpp
```diff
@@ -0,0 +1,113 @@
...
+  private:
+    static constexpr size_t ADDRESS_BYTES = 2;
+    // I2C address of the EEPROM, shifted 1 bit left from the
+    // datasheet.
+    static constexpr const uint8_t _address = ADDRESS << 1;
+
+  public:
+    static const constexpr uint16_t PAGES = 128;
+
+    template <typename T, M24128_Policy Policy>
```
sfoster1: might be nice to have the sizeof requirement in a concept for T

sfoster1 (approved):
Looks good, might want to put the sizeof requirement in a concept though

PR #442, opened 2023-05-22
feat(tempdeck-gen3): enable front LED driver

fsinapi:
This PR adds code to control the front LED strip on the Tempdeck Gen3.

The LED strip consists of three RGBW LED's, all controlled by a single LED driver IC controlled via I2C (PN IS31FL). The driver IC contains configuration registers to control both 1) the current of each LED 2) the PWM of each LED.

The driver for the IS31FL just operates on int-enumerated channels. The mapping of channels to their respective colors is kept in the UI Task.

For now, the firmware just sets the LED strip to be all-white on startup and never does anything else.


sfoster1 (approved):
Looks good to me, surprised we hadn't had any kind of i2c write before

fsinapi:
> Looks good to me, surprised we hadn't had any kind of i2c write before

All the other ones in the codebase either 1) wrote a prefixed register size, or  2) wrote an arbitrary array without a separate register-address field. This one worked a little more cleanly with the IC :)

PR #443, opened 2023-05-22
feat(tempdeck-gen3): include thermal calibration offsets when calculating temperature

fsinapi:
This PR adds functionality for the TD3 to include the calibration values when calculating temperature for the thermistors.

The algorithm is the same as most of our other thermal-control modules: `temp = (A * heatsink temp) + ((B+1)*plate temp) + C`

Tested setting the offsets on a board, reading back the temperatures, and verifying that the temperatures in fact changed.


sfoster1 (approved):
nice!

PR #449, opened 2024-02-27
feat(thermocycler): check closed lid sensor when a move finishes

caila-marashaj:
## Overview
Currently, the thermocycler's optical switches are not checked after opening or closing the lid to make sure that the action was performed successfully. In the case of closing the lid, we want to make sure that there isn't any extra labware obstructing the lid from closing before moving forward with a protocol.

This resolves https://opentrons.atlassian.net/browse/RQA-2303

# Changelog

- create an `UNEXPECTED_LID_STATE_ERROR`
- edit `motor_task.hpp` to throw said error after handling a `LidStepperComplete` with lid status `CLOSE_OVERDRIVE`, if the lid closed sensor is not triggered
- update the tests
- some CMake changes to stop gcc's yelling

### DIFF
--- a/stm32-modules/include/thermocycler-gen2/thermocycler-gen2/messages.hpp
+++ b/stm32-modules/include/thermocycler-gen2/thermocycler-gen2/messages.hpp
@@ -361,8 +361,7 @@ struct CloseLidMessage {
     uint32_t id;
 };
 
-struct LidStepperError {
-};
+struct LidStepperError {};
 
 struct PlateLiftMessage {
     uint32_t id;
--- a/stm32-modules/include/thermocycler-gen2/thermocycler-gen2/messages.hpp
+++ b/stm32-modules/include/thermocycler-gen2/thermocycler-gen2/messages.hpp
@@ -361,8 +361,6 @@ struct CloseLidMessage {
     uint32_t id;
 };
 
-struct LidStepperError {};
-
 struct PlateLiftMessage {
     uint32_t id;
 };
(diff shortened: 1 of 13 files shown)


stm32-modules/include/thermocycler-gen2/thermocycler-gen2/messages.hpp
```diff
@@ -361,6 +361,8 @@ struct CloseLidMessage {
     uint32_t id;
 };
 
+struct LidStepperError {};
```
pmoegenburg: Will this be used?

stm32-modules/include/thermocycler-gen2/thermocycler-gen2/motor_task.hpp
```diff
@@ -1289,8 +1289,10 @@ class MotorTask {
                     motor_util::LidStepper::Position::CLOSED;
                 // The overall lid state machine can advance now
                 error = handle_lid_state_end(policy);
-                // TODO(Frank, Mar-7-2022) check if the lid didn't make it in
-                // all the way
+                // if the lid isn't actually closed, overwrite error status
```
pmoegenburg, line 1293: Do we want to do this before the `handle_lid_state_end` call above?

stm32-modules/include/thermocycler-gen2/thermocycler-gen2/messages.hpp
caila-marashaj: nah, I realized I can just use the existing `ErrorMessage`

stm32-modules/include/thermocycler-gen2/thermocycler-gen2/motor_task.hpp
caila-marashaj: I figured it would be easier to overwrite the error value after that function is done, since inside of its scope it doesn't know the desired status of the lid, which in this case is `OVERDRIVE`

sfoster1 (approved):
Looks good to me, nice!

PR #450, opened 2024-07-15
chore(stacker): create project

ahiuchingau:
Project skeleton for the FLEX Stacker module. 

Closes PLAT-445


.github/workflows/stacker.yaml
```diff
@@ -0,0 +1,97 @@
+name: 'Tempdeck Gen3 build/test'
```
sfoster1:
```suggestion
name: 'stacker build/test'
```

stm32-modules/stacker/firmware/host_comms_task/usbd_desc.c
```diff
@@ -0,0 +1,279 @@
...
+
+/* Includes ------------------------------------------------------------------*/
+#include "usbd_core.h"
+#include "usbd_desc.h"
+#include "usbd_conf.h"
+
+/* Private typedef -----------------------------------------------------------*/
+/* Private define ------------------------------------------------------------*/
+// TODO(alise): Uncomment these VID/PID definitions and replace
+// the temporary values with the actual values for the shuttle
```
sfoster1: I'm pretty sure we can just pick a PID, and we can use the same VID because the logic is still "this is an ST part"

sfoster1 (approved):
Couple inline things about changing names but lookin good!

Laura-Danielle:
Nitpick: could we call the project `plate_stacker`. I know that it's technically not the product name but I feel like it's more descriptive for firmware.

Laura-Danielle (approved):
LGTM

PR #451, opened 2024-07-26
feat(flex-stacker): motor driver task set up 

ahiuchingau:
We're controlling 3 stepper motors using TMC2160 drivers on the same SPI bus. This PR creates a freeRTOS task for initializing the motor driver config for each motor, and allow the host to adjust the driver settings on the fly. 

TODO: The freeRTOS tasks do not seem to be set up properly, i'm suspecting it's the systick handler messing things up.


sfoster1 (approved):
All looks good if it works!

PR #452, opened 2024-07-30
chore(flex-stacker): Disable DMA spi for now

ryanthecoder:
Just want to get this working quick for now, keeping this as a solo commit so it can be easily reverted later


sfoster1 (approved):
:(

PR #453, opened 2024-07-31
feature(flex-stacker): bring up host comms task

ryanthecoder:
This PR sets up the host comms task for the stacker. You can now send a GCode to read from a register of one of the motor drivers. 
```
Tx ==> M920 X0
Rx <== M920 X0 1 OK
```

, where X is the axis [X | Z | L] and 0 is the register address in decimal and 1 is the data read from the register (decimal).

closes PLAT-427


vegano1 (approved):
This looks good, thank you for setting it up!

PR #455, opened 2024-08-16
feat(flex-stacker): add Gcodes to enable motor movements

ahiuchingau:
This PR adds the following GCodes:

#### M921: Write to motor driver register
As an example, if you want to write the following data 0xFFFF to register address 0x1C of the X motor driver:
```.cpp
M921 X28 65535
```
Note that the address and data values are in decimal.

#### M17/M18: Enable/Disable motor
```.cpp
M17 XZL   // enable all motors 
M18 X  // disable X motor 
```
#### G6: Motor motor at frequency
You can move a motor for a fixed number of steps at a specific frequency.
```
G6 X<direction> S<number of steps> F<step rate in Hz> 
```
`X<direction>`: 1 for positive, 0 for negative
`Z<direction>`: 1 for positive, 0 for negative
`L<direction>`: 1 for open, 0 for close

#### M0: stop motors


stm32-modules/flex-stacker/firmware/motor_control/motor_hardware.c
```diff
@@ -285,7 +287,7 @@ void hw_enable_motor(MotorID motor_id) {
             port = Z_EN_PORT;
             pin = Z_EN_PIN;
             status = HAL_TIM_Base_Start_IT(&htim20);
-            HAL_NVIC_SetPriority(TIM20_UP_IRQn, 5, 0);
+            HAL_NVIC_SetPriority(TIM20_UP_IRQn, 10, 0);
             HAL_NVIC_EnableIRQ(TIM20_UP_IRQn);
             if (status == HAL_OK)
```
vegano1, line 292: can we remove these if we are not using them, we are also returning a bool based on the status so I don't think we need this.

stm32-modules/flex-stacker/firmware/motor_control/motor_hardware.c
```diff
@@ -305,19 +307,20 @@ void hw_enable_motor(MotorID motor_id) {
...
-            HAL_NVIC_SetPriority(TIM3_IRQn, 6, 0);
+            HAL_NVIC_SetPriority(TIM3_IRQn, 10, 0);
             HAL_NVIC_EnableIRQ(TIM3_IRQn);
             if (status == HAL_OK)
             {
             }
             break;
         default:
-            return;
+            return status == HAL_OK;
```
vegano1: This default case should probably explicitly return `false`, since we did not match any of the cases and thus did not `enable` a motor.

stm32-modules/flex-stacker/firmware/motor_control/motor_hardware.c
```diff
@@ -326,30 +329,22 @@ void hw_disable_motor(MotorID motor_id) {
...
             port = L_EN_PORT;
             pin = L_EN_PIN;
             status = HAL_TIM_Base_Stop_IT(&htim3);
-            if (!status == HAL_OK){
-                Error_Handler();
-            }
             break;
         default:
-            return;
+            return status == HAL_OK;
```
vegano1: Same here, maybe return `false` if there are no matches.

stm32-modules/include/flex-stacker/flex-stacker/motor_task.hpp
```diff
@@ -47,23 +48,61 @@ class MotorTask {
...
+            response, Queues::HostCommsAddress));
+    }
+
+    template <MotorControlPolicy Policy>
+    auto motor_enable(MotorID id, std::optional<bool> engage, Policy& policy)
+        -> bool {
+        if (!engage.has_value()) {
+            return true;
+        }
+        return engage ? policy.enable_motor(id) : policy.disable_motor(id);
```
vegano1: This will always return `true`, it should be `return engage.value() ? ...` instead.

stm32-modules/include/flex-stacker/flex-stacker/host_comms_task.hpp
```diff
@@ -310,6 +315,122 @@ class HostCommsTask {
...
+        std::sized_sentinel_for<InputLimit, InputIt>
+    auto visit_gcode(const gcode::DisableMotor& gcode, InputIt tx_into,
+                     InputLimit tx_limit) -> std::pair<bool, InputIt> {
+        auto id = ack_only_cache.add(gcode);
+        if (id == 0) {
+            return std::make_pair(
+                false, errors::write_into(tx_into, tx_limit,
+                                          errors::ErrorCode::GCODE_CACHE_FULL));
+        }
+        auto message = messages::MotorEnableMessage{
```
vegano1, line 376: should this not be `MotorDisableMessage` ?

stm32-modules/include/flex-stacker/flex-stacker/gcodes_motor.hpp
```diff
@@ -0,0 +1,380 @@
...
+
+#include "core/gcode_parser.hpp"
+#include "core/utility.hpp"
+#include "flex-stacker/errors.hpp"
+#include "systemwide.h"
+
+namespace gcode {
+
+auto inline motor_id_to_char(MotorID motor_id) -> const char* {
+    return static_cast<const char*>(motor_id == MotorID::MOTOR_X   ? "X"
```
vegano1, line 25: Maybe use a switch statement here so we know if there's an unknown MotorID value and not default it to "L".

vegano1 (approved):
Left a few comments, overall this is great!

stm32-modules/flex-stacker/firmware/motor_control/motor_hardware.c
ahiuchingau: Good call! I actually updated these in the next PR, so i'll leave these as such right now

stm32-modules/include/flex-stacker/flex-stacker/host_comms_task.hpp
ahiuchingau: I'm using the same message here, but the data will look different to differentiate between the two.
Enable will be true, disable will be false, and unset is null.

PR #457, opened 2024-08-22
fix(flex-stacker): enable spi dma

ahiuchingau:
You need to turn on the DMA clock before configuring SPI in order for it to work.


sfoster1 (approved):
ah, that'll do it!

PR #458, opened 2024-08-28
feat(flex-stacker): add motor interrupt handlers

ahiuchingau:
This PR implements Movement Profile in the motor interrupt handlers to allow the motors to move at specified frequency (microsteps/sec) and for a fixed distance (microsteps).


vegano1 (approved):
This is looking great, really liking how we consolidated the get_motor code.

PR #459, opened 2024-08-29
feat(flex-stacker): add open loop movement gcode

ahiuchingau:
This PR allows us to command one of the motors to move in a specific direction until the limit switch is reached. 

Must only be merged in after https://github.com/Opentrons/opentrons-modules/pull/458


vegano1 (approved):
👍

PR #460, opened 2024-08-29
feat(flex-stacker): add gcode with params in mm and mm/s

ahiuchingau:
This PR adds GCode messages that allow user to specific distance in mm, velocity in mm/s and other motion related params, instead of relying solely on number of microsteps.

### DIFF
--- a/stm32-modules/include/flex-stacker/flex-stacker/gcodes_motor.hpp
+++ b/stm32-modules/include/flex-stacker/flex-stacker/gcodes_motor.hpp
@@ -347,6 +347,83 @@ struct MoveMotorAtFrequency {
     }
 };
 
+struct MoveMotor {
+    MotorID motor_id;
+    bool direction;
+    uint32_t frequency;
+
+    using ParseResult = std::optional<MoveMotor>;
+    static constexpr auto prefix = std::array{'G', '5', ' '};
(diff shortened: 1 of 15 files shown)


stm32-modules/include/flex-stacker/flex-stacker/gcodes_motor.hpp
```diff
@@ -255,42 +255,42 @@ struct DisableMotor {
...
-struct MoveMotorAtFrequency {
+struct MoveMotorInSteps {
     MotorID motor_id;
-    bool direction;
     int32_t steps;
-    uint32_t frequency;
+    uint32_t steps_per_second;
+    uint32_t steps_per_second_sq;
 
-    using ParseResult = std::optional<MoveMotorAtFrequency>;
```
vegano1, line 264: do we no longer need this command and the 'G6' gcode?

vegano1 (approved):
This is looking real good!

stm32-modules/include/flex-stacker/flex-stacker/gcodes_motor.hpp
ahiuchingau: Keeping the implementation, just converting it to `G0.S`

PR #461, opened 2024-09-05
fix(flex-stacker): do not disengage X and L motors when a move finishes

ahiuchingau:
Unlike the ebrake-supported Z motor, the XL motors need to stay engaged to hold its position when they are no longer moving. Otherwise the springs in the mechanical system can cause the motors to drift unexpectedly.


sfoster1 (approved):
Love a surprise spring

PR #463, opened 2024-09-13
fix(flex-stacker): increase comms task stack size

ahiuchingau:
As it turns out, we were using the wrong variable for setting the stack size of the comms task. Now that it has increased from 256 to 2048, we are finally able to use floats in gcodes.


sfoster1 (approved):
Ah, that would do it

PR #464, opened 2024-09-13
feat(flex-stacker): make move params persist

ahiuchingau:
This PR does the following:
1. Add gcode that sets the microstep resolution for the motors - after motor driver task finishes, it passes the message down the motor task, before sending the ack back to the comms task 
2. Make move params optional and persist between moves


sfoster1 (approved):
Looks good but you should add a way to query the options that are currently set

ahiuchingau:
Good point! I will add that to this PR

PR #465, opened 2024-09-17
feat(flex-stacker): add deceleration phase in motor interrupt handler

ahiuchingau:
This PR does the following:
1. fix the acceleration conversion mistake (accidentally multiply `steps_per_mm` twice) 
2. add deceleration phase in motor interrupt handler!

Please sanity check my calculations.


sfoster1 (approved):
interesting, cool approach!

PR #466, opened 2024-09-30
feat(flex-stacker): add the motor stallguard functionality and gcodes.

vegano1:
# Overview

The Flex Stacker motors do not have encoders on them so we can't use the same method we use for the Flex firmware to detect stalls, fortunately, the TMC2160 motors have a stall detection mechanism called `StallGuard` which uses back EMF to detect load changes. This pull request exposes the stall guard feature and takes advantage of the `diag0` pin to stop the motors in case of a stall. We then also provide a new `M910: SetMotorStallGuard` to enable and set each motor's Stall Guard Threshold (SGT) value to allow hardware testing to tune the values.

Closes: [PLAT-509](https://opentrons.atlassian.net/browse/PLAT-509)

# Changelog

- Enable diag0 pin by changing the `TCOOLTHRS > TSTEP` value to `0x2FF`
- Enable diag0_stall for all motors in the GCONF (0x00) register
- Enable the Stall Guard Threshold sgt value in the COOLCONF (0x6D) register
- Added an interrupt to listen to diag0 pin, which sends a `GPIOInterruptMessage` to the `motor_task`
- Handle the `GIOInterruptMessage` in the `motor_task` which stops the motors from running.
- Add new gcode `M910: SetMotorStallGuard` to enable/disable and set each motor's Stall Guard Threshold (SGT) value
- Add new gcode `M911: GetMotorStallGuard` to get the SGT since the coolconf reg is write-only.

# Testing

- [x] Make sure that the motors do not stop when held if stallguard Is disabled with `M910`
- [x] Make sure that we can change the stallguard sensitivity with `M910` and that the motors stops when held
- [x] Make sure we can get the stallguard threshold with `M911` after setting it with `M910`

# Review Requests

- Makes sense?
- Anything I'm missing?

[PLAT-509]: https://opentrons.atlassian.net/browse/PLAT-509?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/stm32-modules/flex-stacker/firmware/motor_control/motor_hardware.c
+++ b/stm32-modules/flex-stacker/firmware/motor_control/motor_hardware.c
@@ -1,77 +1,15 @@
-
 #include "stm32g4xx_hal.h"
 #include "stm32g4xx_it.h"
+#include "motor_hardware.h"
 #include "systemwide.h"
 #include "FreeRTOS.h"
 
 #include <stdbool.h>
+#include <stdint.h>
 
 #ifdef __cplusplus
 extern "C" {
 #endif // __cplusplus
-
-/******************* Motor Z *******************/
-/** Motor hardware **/
-#define Z_STEP_PIN (GPIO_PIN_2)
-#define Z_STEP_PORT (GPIOC)
-#define Z_DIR_PIN (GPIO_PIN_1)
-#define Z_DIR_PORT (GPIOC)
-#define Z_EN_PIN (GPIO_PIN_3)
-#define Z_EN_PORT (GPIOA)
-#define Z_N_BRAKE_PIN (GPIO_PIN_7)
-#define Z_N_BRAKE_PORT (GPIOB)
-
-/** Limit switches **/
-/* Note: Photointerrupters limit switches */
-#define Z_MINUS_LIMIT_PIN (GPIO_PIN_3)
-#define Z_MINUS_LIMIT_PORT (GPIOC)
-#define Z_PLUS_LIMIT_PIN (GPIO_PIN_0)
-#define Z_PLUS_LIMIT_PORT (GPIOA)
-
-
-/******************* Motor X *******************/
-
-/** Motor hardware **/
-#define X_STEP_PIN (GPIO_PIN_7)
-#define X_STEP_PORT (GPIOA)
-#define X_DIR_PIN (GPIO_PIN_6)
-#define X_DIR_PORT (GPIOA)
-#define X_EN_PIN (GPIO_PIN_4)
-#define X_EN_PORT (GPIOA)
-#define X_N_BRAKE_PIN (GPIO_PIN_9)
-#define X_N_BRAKE_PORT (GPIOB)
-
-
-/** Limit switches **/
-/* Note: Photointerrupters limit switches */
-#define X_MINUS_LIMIT_PIN (GPIO_PIN_1)
-#define X_MINUS_LIMIT_PORT (GPIOA)
-#define X_PLUS_LIMIT_PIN (GPIO_PIN_2)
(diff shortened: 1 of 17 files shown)


stm32-modules/include/flex-stacker/flex-stacker/motor_driver_task.hpp
```diff
@@ -278,6 +278,32 @@ class MotorDriverTask {
...
 
+    template <tmc2160::TMC2160InterfacePolicy Policy>
+    auto visit_message(const messages::SetMotorStallGuardMessage& m,
+                       tmc2160::TMC2160Interface<Policy>& tmc2160_interface)
+        -> void {
+        auto response = messages::AcknowledgePrevious{.responding_to_id = m.id};
+        driver_conf_from_id(m.motor_id).gconfig.diag0_stall =
+            static_cast<int>(m.enable);
+        if (m.sgt.has_value()) {
+            if (m.sgt.value() >= -64 && m.sgt.value() <= 63) {
```
ahiuchingau: Can we create this as a function, something like `verify_sgt_value()`, in the tmc2160 module?

stm32-modules/flex-stacker/firmware/motor_control/motor_hardware.c
```diff
@@ -217,6 +163,16 @@ void motor_hardware_gpio_init(void){
...
+
+    /*Configure GPIO pins : INPUTs IRQ */
+    init.Mode = GPIO_MODE_IT_FALLING;
+    init.Pull = GPIO_PULLUP;
+    init.Speed = GPIO_SPEED_FREQ_LOW;
+
+    init.Pin = MOTOR_DIAG0_PIN;
+    HAL_GPIO_Init(MOTOR_DIAG0_PORT, &init);
+    HAL_NVIC_SetPriority(EXTI15_10_IRQn, 6, 0);
+    HAL_NVIC_EnableIRQ(EXTI15_10_IRQn);
```
ahiuchingau: Do we want to enable this interrupt all the time? Or should we only do it when we tell the motor driver to enable to the diag pin? I'm only concerned about this because i'm not sure how noisy this line is

stm32-modules/include/flex-stacker/firmware/motor_interrupt.hpp
```diff
@@ -67,13 +68,19 @@ class MotorInterruptController {
...
         set_direction(direction);
         _profile = motor_util::MovementProfile(
             TIMER_FREQ, steps_per_sec_discont, steps_per_sec, step_per_sec_sq,
             motor_util::MovementType::OpenLoop, 0);
         _policy->enable_motor(_id);
         _response_id = move_id;
     }
+    auto stop_movement(uint32_t move_id) -> void {
+        _stop = true;
+        _policy->disable_motor(_id);
```
ahiuchingau: Not entirely sure if we want to disable all the motors since we'll need the X and L to be holding as they don't have brakes. Can we call stop motor instead?

ahiuchingau (changes requested):
Implementing the `aggregator.send_from_isr` is ingenious! I didn't think of that.

Got a couple of questions. Mainly not sure if we want to disable the X (and L, but I'm not as concerned about that) motors since they don't have brakes and the axis has a pretty strong spring that might displace the motor.

stm32-modules/flex-stacker/firmware/motor_control/motor_hardware.c
vegano1: fixed, nice callout

ahiuchingau (approved):
💂💂💂

PR #467, opened 2024-09-30
fix(flex-stacker): g-code message parsing

ahiuchingau:
@vegano1 pointed out that the `SetTMCRegister` G-code no longer works. This PR fixes that and refactors the parser by adding argument templates so we don't have to re-write the structs so many times.


sfoster1 (approved):
Now that's very satisfying

vegano1 (approved):
Very nice, thats a lot cleaner!
I tested this branch on available gcodes and all are working as expected.

PR #468, opened 2024-10-04
feat(flex-stacker): stream motor driver stall guard value during move

ahiuchingau:
This PR creates a freertos task that handles the streaming of the stall guard value if requested in the move. The stall guard result from the motor driver will be reported via a `M900` message every 10 ms while the motor is in motion.

Any concerns regarding how the freertos task is created/executed?

### DIFF
--- a/stm32-modules/flex-stacker/firmware/motor_control/freertos_motor_driver_task.cpp
+++ b/stm32-modules/flex-stacker/firmware/motor_control/freertos_motor_driver_task.cpp
@@ -3,6 +3,7 @@
 #include "firmware/motor_driver_policy.hpp"
 #include "firmware/motor_hardware.h"
 #include "flex-stacker/motor_driver_task.hpp"
+#include "flex-stacker/tmc2160_interface.hpp"
 #include "ot_utils/freertos/freertos_timer.hpp"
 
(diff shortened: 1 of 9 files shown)


vegano1 (changes requested):
Thank you for setting this up and I'm glad it's working!

A few comments

1. It could get hard to manage multiple tasks from the same module, usually, we have one task per module. So if we want to go the new task route it would make sense to pull the `run_stallguard_task` into its own file that follows the `motor_task.hpp` and `motor_driver_task.hpp` framework.

2. Alternatively, instead of creating a new task, we could also use the `freertos_timer.hpp` to set up a timer from the `motor_driver_task` that calls `read_stallguard` and sends the result over USB. We use the timer with the Hepa/UV in the ot3-firmware repo, you can find that [here](https://github.com/Opentrons/ot3-firmware/blob/03a648a80decc770783d45c4fed0daa56c230ca2/include/hepa-uv/core/uv_task.hpp#L36).

vegano1:
@ahiuchingau 
Comments have been addressed.

sfoster1 (approved):
10ms feels pretty fast for a text-mode UART but hey the math says it works so we good

PR #469, opened 2024-10-15
feat(thermocycler): handle lid movement errors

caila-marashaj:
## Overview
We should have a check for the end of lid stepper movements where we expect the status of a limit switch to change. If an open or close movement fails, we'll now:

- set the lid position to `UNKNOWN`
- disengage the lid solenoid
- set the lid stepper state to `IDLE` 
- reset the lid stepper current to the hold current rather than the run current

## Changelog

- add a function for handling lid movement errors uniformly
- if an errant limit switch state is found, send an `ErrorMessage` and break out of the state machine


ahiuchingau (commented):
Great! This error handling implementation makes sense to me - I'm surprised we didn't have that before.

Could you (1) add new state machine tests and (2) let's make sure the changes here have as little impact on previous tests as possible (don't change old behavior unless it was wrong)?

PR #471, opened 2024-10-21
feat(thermocycler): make sure the lid ease off the latch before attempting to open

ahiuchingau:
## Overview
During ABR testing, we found that sometimes the thermocycler lid could get stuck on the latch while opening. 

Hardware advised that we close the lid further, to allow the probes on the lid to ease off the latch plate, before releasing the latch. This should ensure that the latch and probes are not locked up by friction during lid opening. 

## Changelog

- when opening the lid from the closed position:
              - first release the solenoid
              - drive the lid closed 1 degree
              - open the lid the first 10 degrees until the optical switch should be disengaged
              - check that the optical switch was successfully disengaged 
              - open the lid the rest of the way
- after opening the lid all the way, check that the 'lid open' limit switch is triggered and throw an error if it isn't
- after backing off the 'lid open' switch, check that it is no longer engaged
- after closing the lid, check that the 'lid closed' limit switch is triggered and throw an error if it isn't
- after overdriving while closing the lid, check that the 'lid closed' switch is still triggered
- no longer engage/disengage the solenoid during close lid to ensure the prongs have a chance to hook onto the latch

### DIFF
--- a/stm32-modules/include/thermocycler-gen2/test/test_motor_policy.hpp
+++ b/stm32-modules/include/thermocycler-gen2/test/test_motor_policy.hpp
@@ -43,10 +43,9 @@ class TestMotorPolicy : public TestTMC2130Policy {
     auto lid_solenoid_disengage() -> void { _solenoid_engaged = false; }
     auto lid_solenoid_engage() -> void { _solenoid_engaged = true; }
 
-    auto lid_read_closed_switch() -> bool {
-        return _lid_closed_switch; }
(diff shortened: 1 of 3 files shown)


sfoster1 (approved):
Okay, this looks good to me! THank you so much for dealing with this

vegano1 (approved):
Nice work!

ryanthecoder (approved):
LGTM Gonna miss CLOSING_CLOSE_HINGE and OPENING_OPEN_HINGE

PR #472, opened 2024-10-31
feat(flex-stacker): board revision handling & add system task

ahiuchingau:
This PR does two things:
1. allow us to specify the board revision from cmake by adding the `STACKER_REVISION` var
2. add the system task to handle `dfu: EnterBootloader`, `M996: Set Serial Number` and `M115: GetSystemInfo` gcodes


sfoster1 (approved):
I so dearly love that enter-bootloader code lol

PR #473, opened 2024-11-07
refactor(flex-stacker): linear motion system and motor task

ahiuchingau:
This PR simplifies things a bit in the motor task which helps make implementation of the homing routine a bit easier later.

### DIFF
--- a/stm32-modules/include/common/core/linear_motion_system.hpp
+++ b/stm32-modules/include/common/core/linear_motion_system.hpp
@@ -13,20 +13,21 @@ class BeltConfig {
 };
 
 class LeadScrewConfig {
-    public:
-    static constexpr auto mm_per_rev(float lead_screw_pitch, float gear_reduction_ratio) -> float {
+  public:
+    static constexpr auto mm_per_rev(float lead_screw_pitch,
+                                     float gear_reduction_ratio) -> float {
         return lead_screw_pitch / gear_reduction_ratio;
     }
 };
 
 class GearBoxConfig {
   public:
-    static constexpr auto mm_per_rev(float gear_diameter, float gear_reduction_ratio) -> float {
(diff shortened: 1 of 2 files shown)


stm32-modules/include/common/core/linear_motion_system.hpp
```diff
@@ -5,61 +5,39 @@
 
 namespace lms {
 
-struct BeltConfig {
-    // NOLINTNEXTLINE(misc-non-private-member-variables-in-classes)
-    float pulley_diameter;  // mm
-    [[nodiscard]] constexpr auto get_mm_per_rev() const -> float {
+class BeltConfig {
```
sfoster1: [nit] If everything is going to be public might as well keep it a struct

stm32-modules/include/flex-stacker/flex-stacker/motor_task.hpp
```diff
@@ -299,25 +295,8 @@ class MotorTask {
...
-            case MotorID::MOTOR_L:
-                _l_mech_conf.microstep =
-                    static_cast<float>(pow(2, m.microsteps_power));
-                _l_state.steps_per_mm = _l_mech_conf.get_usteps_per_mm();
-                break;
-            default:
-                break;
-        }
+        motor_state(m.motor_id).lms_config.microstep =
+            static_cast<float>(pow(2, m.microsteps_power));
```
sfoster1: I think that if you make this `pow(2.f, m.microsteps_power)` it'll remain a float without the static cast

sfoster1 (approved):
Looks good to me, two nits

stm32-modules/include/common/core/linear_motion_system.hpp
ahiuchingau: yes yes

vegano1 (approved):

PR #474, opened 2024-11-07
fix(flex-stacker): clear ack message cache when a stall is detected.

vegano1:
# Overview

This fixes an issue where the ack cache would get full when a stall occurred because the incoming move message was not getting the expected response. So let's clear the ack cache when we get a stall before we send an error message.

# Testing

- [x] Trigger multiple stalls and make sure we no longer get gcode cache full messages


sfoster1 (approved):
Looks good I think. This means that the python side is expected to cancel its tracking of any in-flight moves when it gets a stall, right?

ahiuchingau (approved):
Nice

vegano1:
> Looks good I think. This means that the python side is expected to cancel its tracking of any in-flight moves when it gets a stall, right?

yeah that's correct.

PR #478, opened 2024-11-14
feat(thermocycler): don't operate the solenoid while trying to close the thermocycler lid

caila-marashaj:
## Overview
One of the change from a [previous firmware fix](https://github.com/Opentrons/opentrons-modules/pull/471). 

Instead of opening the solenoid latch while the thermocycler lid closes, then closing the latch after the move is complete, just don't engage the solenoid while closing in order to give the lid's prongs a better chance to catch.


ahiuchingau (approved):
yep yep

PR #480, opened 2024-11-14
feat(thermocycler): overdrive tc lid before opening from closed position

caila-marashaj:
## Overview
In [this thermocycler commit](https://github.com/Opentrons/opentrons-modules/pull/471), we changed the procedure of opening the thermocycler lid to the following:

- engage the solenoid latch
- if the lid is closed, overdrive the lid further closed
- open the lid 8 degrees
- check the 'closed' limit switch. If it's engaged, throw an error
- proceed to opening the lid all the way 

This re-implements that change, but removes the 'latch release backoff' step, so opening the lid will now look like:

- engage the solenoid latch
- if the lid is closed, overdrive the lid further closed
- proceed to opening the lid all the way


sfoster1 (approved):
Nice. So complex

PR #481, opened 2024-11-19
feat(flex-stacker): implement platform sensors and door switch gpios

ahiuchingau:
This PR enables the reading of the hopper door switch and platform sensors from the Flex Stacker.

My logic for placing the platform sensors in the motor task is that the sensors will be primarily used for motor movements. The door switch logic is owned by the system task instead.

### DIFF
--- a/stm32-modules/include/flex-stacker/flex-stacker/messages.hpp
+++ b/stm32-modules/include/flex-stacker/flex-stacker/messages.hpp
@@ -262,7 +262,8 @@ using HostCommsMessage =
     ::std::variant<std::monostate, IncomingMessageFromHost, ForceUSBDisconnect,
                    ErrorMessage, AcknowledgePrevious, GetSystemInfoResponse,
                    GetTMCRegisterResponse, GetLimitSwitchesResponses,
-                   GetMoveParamsResponse, GetMotorStallGuardResponse>;
+                   GetMoveParamsResponse, GetMotorStallGuardResponse,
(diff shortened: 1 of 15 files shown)


stm32-modules/include/flex-stacker/flex-stacker/messages.hpp
```diff
@@ -239,6 +239,27 @@ struct GetMotorStallGuardResponse {
     int sgt;
 };
 
+struct GetDoorClosedMessage {
+    uint32_t id;
+};
+
+struct GetDoorClosedResponse {
+    uint32_t responding_to_id;
+    ;
```
vegano1: typo

stm32-modules/include/flex-stacker/flex-stacker/messages.hpp
```diff
@@ -239,6 +239,27 @@ struct GetMotorStallGuardResponse {
...
+};
+
+struct GetPlatformSensorsMessage {
+    uint32_t id;
+};
+
+struct GetPlatformSensorsResponse {
+    uint32_t responding_to_id;
+    ;
+    bool extend;   // Sensor located on the positive end of X axis
```
vegano1: remove ;
`extend` is a keyword

vegano1 (approved):
nice, this looks good.
left 2 small comments.

sfoster1 (approved):
agree with the semicolon and keyword things, looks good otherwise

PR #482, opened 2024-11-20
feat(flex-stacker): add estop

ahiuchingau:
This PR lets us read the estop state from Gcode and sets it up as an interrupt that notifies the motor task in the same way as the driver stall diagnostic pin.


sfoster1 (approved):
Looks good. We'll have to be very careful with our python-side comms handling now that this can send its own messages

stm32-modules/include/flex-stacker/flex-stacker/motor_task.hpp
```diff
@@ -406,7 +415,11 @@ class MotorTask {
         _z_controller.stop_movement(0, true);
         _x_controller.stop_movement(0, false);
         _l_controller.stop_movement(0, false);
-        send_error_message(Error::MOTOR_STALL_DETECTED);
+        if (policy.is_diag0_pin(m.pin)) {
+            send_error_message(Error::MOTOR_STALL_DETECTED);
+        } else {
```
vegano1: lets check the specific pin here with "if else" or switch case.

vegano1 (approved):
very nice, left  small comment.

PR #485, opened 2024-11-22
feat(thermocycler and heater-shaker): grab reset flags on startup

caila-marashaj:
## Overview
There have been a couple of instances in abr where the thermocycler and the heater-shaker's firmware have unexpectedly restarted. Since the reason this happened isn't obvious, let's grab the HAL's provided reset flags, and give the modules the ability to ask for them.

[According to the internet](https://stackoverflow.com/questions/34196663/stm32-how-to-get-last-reset-status), this value needs to be grabbed as soon as possible into hardware initialization. So, this code saves it to a public variable inside `motor_hardware_setup` where it can be passed up to the usb bus whenever.

## Changelog

- create `GetResetReason` gcode and task messages
- handle new messages
- get the value of the HAL RCC flag in `motor_harwdware.c`
- add the ability to pass that upward and send it in a flag over usb

## TODO
- Implement this for flex stacker and temp deck also
- In the Python side, send a `GetResetReason` message on module connection and log the response
- Add some tests in a follow-up pr

## Test Plan

- [X] Make sure each module sends a flag containing the last reset reason in response to a `M114` message 

Is there any way to test/simulate reasons other than a regular power cycle to see if the values show up in the response message?

### DIFF
--- a/stm32-modules/heater-shaker/firmware/motor_task/motor_hardware.c
+++ b/stm32-modules/heater-shaker/firmware/motor_task/motor_hardware.c
@@ -19,6 +19,39 @@ motor_hardware_handles *MOTOR_HW_HANDLE = NULL;
 
 static _Atomic int16_t motor_speed_buffer[MOTOR_SPEED_BUFFER_SIZE];
 static _Atomic size_t motor_speed_buffer_itr = 0;
+uint16_t reset_reason;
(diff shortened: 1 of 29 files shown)


sfoster1 (approved):
Excellent!

PR #486, opened 2024-11-25
fix(flex-stacker): reporting correct limit switch response

ahiuchingau:
We have been mistakenly reporting the x- limit switch result for z-, this fixes it.


vegano1 (approved):

sfoster1 (approved):
ope

PR #487, opened 2024-11-26
fix(flex-stacker): more evt qc fixes

ahiuchingau:
Changelog:
1. Update the R_Sense value for the motor driver (evt uses 150 ohm instead of 220)
2. Change some of the logic concerning disabling and enabling motor when an ebrake is involved - had to pull out the disabling motor out of the actual interrupt because we now have delays
3. Make sure StopMotor `M0` gcode message is handled by the host comm task


vegano1 (approved):
this looks great.

PR #488, opened 2024-11-27
build(flex-stacker): add flex-stacker build and deply github workflow.

vegano1:


djgomez-opentrons (approved):
Tested FW on stackers that passed QC checks

PR #489, opened 2024-12-03
feat(tempdeck): grab reason for last reset

caila-marashaj:
## Overview
Continuation of https://github.com/Opentrons/opentrons-modules/pull/485 
Enable the temp deck to grab the HAL RCC flag on startup, and pass it over usb whenever asked in response to an `M114` message.


sfoster1 (approved):
ty!

PR #490, opened 2024-12-03
fix(stacker): small bug fixes for EVT 

ahiuchingau:
This PR fixes the stop motor message handling and update values from the stacker EVT extents doc.


sfoster1 (approved):
Nice

PR #491, opened 2024-12-04
feat(flex-stacker): grab reset reason

caila-marashaj:
## Overview
Continuation of https://github.com/Opentrons/opentrons-modules/pull/485
Enable the flex stacker to grab the HAL RCC flag on startup, and pass it over usb whenever asked in response to an M114 message.


ahiuchingau (changes requested):
This is great! Implementation looks good - but could we move this out to the system task?

ahiuchingau (approved):

PR #492, opened 2024-12-04
fix(stacker): motor driver conf update

ahiuchingau:
We need to set the `bbmclks` for the motor drivers to 4 and let's lower the stallguard threshold for now (stallguard not enabled unless you tell it to).


sfoster1 (approved):
nice!

PR #493, opened 2024-12-11
fix(stacker): update tpowerdown and hold current delay for motor driver

ahiuchingau:
Lowering the tpowerdown and ihold delay to minimize the time it needs to change motor current; this should reduce the wobble when we turn on the z-brake before disabling the motor.


djgomez-opentrons:
Tested on bench setup. 

TPOWERDOWN matches the behavior I get when I manually edit the registers. A period of 180ms from the last step pulse to when the current starts to drop with IHOLDDELAY. 
This 180ms is:
+ 65ms standstill delay time 
+ 115ms TPOWERDOWN delay 

TPOWERDOWN seems to be 16ms (which is the period added by incrementing the register by one) longer than what I would expect from the datasheet. 

I would say this is fine for now, not worth spending the time trying to fix the offset error (this might even be the normal behavior of the driver) - but people should know it's there and we should figure out exactly what is going on there at some point. 

IHOLDDELAY is correct with 16ms per current step reductions. 

image

sfoster1 (approved):
Nice!

vegano1 (approved):

PR #494, opened 2024-12-12
fix(flex-stacker): fix invalid Latch motor gear ratio for EVT flex-stackers.

vegano1:
update the Latch motor gear ratio


ahiuchingau (approved):
Can't believe this is only surfacing now... 🤡 

Thanks for the fix. LGTM once CI checks all pass.

PR #497, opened 2024-12-19
fix(stacker): use actual serial number in usb serial number descriptor

ahiuchingau:
This PR updates the USB serial number string descriptor to correctly report the device's serial number stored in flash. This would allow us to easily identify a connected stacker without having to open up a serial connection.

We probably should consider fixing this descriptor for all of the stm32 modules at some point.


sfoster1 (changes requested):
We shouldn't #define an address like this if at all possible to avoid. We should either
- Traceably have the same address define be used in both writing and reading the serial, so it's impossible to get out of sync
- (preferred) define this as an extern; create a section in the linker script; set the value of the extern to the section address so it's harder to accidentally write something else on top of this

sfoster1 (approved):
If this is a quick change for the demo let's just do it this way and change it later

PR #498, opened 2024-12-19
fix(flex-stacker): standardize the gcode responses to use key and value pairs.

vegano1:
# Overview 

While working on the flex stacker hardware controller, I noticed that some gcode responses have different formatting, some have key value responses, and others don't. So to make parsing consistent let's standardize the gcode responses to return key:value pairs which are easier to parse.

# Changelog

- Change the gcode message responses to key:value pairs where needed

# Testing

- [x] Make sure that all the Gcodes still work and respond accordingly


sfoster1 (approved):
Thanks!

PR #499, opened 2025-01-22
fix(stacker): motor movement profile step detection

ahiuchingau:
This PR adds a test for the movement profile used in the Flex Stacker and fixes a motion bug.

I found a delay in motion control; when the motor speed is set to take exactly one step each time the motor timer ticks, the first tick wouldn't cause a step. It's because the bitwise comparison doesn't work very well while we're at the step boundary. 

By setting the step flag to `(static_cast<uint64_t>(1) << radix) - 1;`, we can make sure we're taking a step right when it happens.

The longer I look at this math, the more confused I get. So apologies in advance if none of this makes sense.


sfoster1 (approved):
Oh that's a great idea, nice!

PR #500, opened 2025-01-27
feat(stacker): Add statusbar animation patterns + enable and handle stallguard

vegano1:
# Overview

Now that the Flex Stacker is in ABR and Science, let's enable stallguard and add UI led to some states to inform the user what's happening with the device. This pull request goes in conjunction with this monorepo pull request [Opentrons/opentrons#17349](https://github.com/Opentrons/opentrons/17349)

# Changelog

- Enable stallguard motor error (diag0_stall = 1) and set the stallguard threshold to 2 for all motor axis
- Add debounce when handling GPIO interrupt messages
- check estop and diag0 pin in the motor interrupt so we can react while moving an axis
- change heartbeat led to blink after 500ms without setting the timer directly
- add animation patterns (Static, Pulse, Flash, Confirm) and expose duration, reps to the `SetStatusBarState` gcodes
- add the color yellow to the list of LED colors

# Testing

- [x] Make sure we can set the led status bar pattern, duration, and reps via repl
- [x] Make sure the flex stacker motors stop moving when there is a stall
- [x] Make sure the flex stacker motors stop moving when the estop is pressed
- [x] Make sure we only send an error message when `triggering` estop or stall

# Review requests

### DIFF
--- a/stm32-modules/include/flex-stacker/flex-stacker/ui_task.hpp
+++ b/stm32-modules/include/flex-stacker/flex-stacker/ui_task.hpp
@@ -197,7 +197,9 @@ class UITask {
     }
 
     template <UIPolicyIface Policy>
-    auto visit_message(const messages::UpdateUIMessage& m, Policy& policy)
+    auto visit_message(
+        const messages::UpdateUIMessage& m,
+        Policy& policy)  // NOLINT[readability-function-cognitive-complexity]
         -> void {
         static_cast<void>(m);
         static_cast<void>(policy);
--- a/stm32-modules/include/flex-stacker/flex-stacker/ui_task.hpp
+++ b/stm32-modules/include/flex-stacker/flex-stacker/ui_task.hpp
@@ -52,7 +52,7 @@ static auto color_to_channels(StatusBarColor color) -> const ChannelMapping& {
 }
 
 // The timer driving LED update frequency should run at this period
-static constexpr uint32_t UPDATE_PERIOD_MS = 1;
+static constexpr uint32_t UPDATE_PERIOD_MS = 10;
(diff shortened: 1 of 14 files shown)


stm32-modules/include/flex-stacker/flex-stacker/ui_task.hpp
```diff
@@ -40,34 +44,67 @@ static auto color_to_channels(StatusBarColor color) -> const ChannelMapping& {
...
+        case StatusBarColor::Yellow:
+            return yellow_channels;
         default:
             return white_channels;
     }
 }
 
 // The timer driving LED update frequency should run at this period
-static constexpr uint32_t UPDATE_PERIOD_MS = 1000;
+static constexpr uint32_t UPDATE_PERIOD_MS = 1;
```
sfoster1: that seems like quite a lot, it's the same frequency as the systick. can we make it 10 or something?

stm32-modules/include/flex-stacker/flex-stacker/ui_task.hpp
```diff
@@ -127,36 +164,158 @@ class UITask {
...
         static_cast<void>(m);
         static_cast<void>(policy);
     }
 
     template <UIPolicyIface Policy>
-    auto visit_message(const messages::SetStatusBarColorMessage& m,
+    auto visit_message(const messages::UpdateUIMessage& m, Policy& policy)
+        -> void {
+        static constexpr double TWO = 2.0F;
+        for (auto bar_id : {StatusBarID::Internal, StatusBarID::External}) {
```
sfoster1, line 206: ooh fancy i love it

sfoster1 (approved):
Looks good to me, though that's a pretty fast timer

PR #503, opened 2025-02-10
feat(flex-stacker): Enable the Time-Of-Flight sensors and add GCodes to interface with them.

vegano1:
# Overview

The TMF8820/21/28 is a dToF (direct time of flight) wide field of view optical distance sensor module achieving up to 5000 mm target detection distance and has 3x3, 4x4, 3x6, or 8x8 zones. It will allow the Flex Stacker to detect if there is labware on the X and Z axis. This pull request brings up both TOF sensors and exposes some functionality via new Gcodes. This pull request goes in conjunction with this monorepo pull request [Opentrons/opentrons#17478](https://github.com/Opentrons/opentrons/pull/17478).

```
M215 - Get TOF Sensor Status
M222 - Get TOF Driver Register
M223 - Set TOF Driver Register
M224 - Enable TOF Driver
```

Closes: [EXEC-971](https://opentrons.atlassian.net/browse/EXEC-971) [EXEC-1166](https://opentrons.atlassian.net/browse/EXEC-1166) [EXEC-1170](https://opentrons.atlassian.net/browse/EXEC-1170) [EXEC-1171](https://opentrons.atlassian.net/browse/EXEC-1171) [EXEC-1172](https://opentrons.atlassian.net/browse/EXEC-1172)

## Test Plan and Hands on Testing

- [x] Make sure both sensors are initialized to the "Measure" application on cold boot.
- [x] Make sure you can get the TOF sensor status for both sensors
- [x] Make sure you can set and get the registers for each TOF sensor
- [x] Make sure you can enable/disable the TOF sensor

## Changelog

- Adds a "TOFSensor" task that is responsible for interfacing with the tmf8820 driver
- Adds a "TMF8820" driver interface class to load the measure application on cold boot.
- Added `scripts/hex_to_c.py` script that generates c array files from hex files.
- Added `scripts/requirements.txt` to keep track of script python dependencies, only `intelhex` for now.
- Added a cmake mechanism to install the `scripts/requirements.txt` file
- Added a cmake mechanism for the flex-stacker that uses the scripts/hex_to_c.py to generate the tmf8820_image.c file from the tmf8820_image.hex file.
- Set the I2C device address to custom addresses for both sensors (0x39 for X) and (0x40 for Z)
- add M215 - Get TOF Sensor Status Gcode and handling
- add M222 - Get TOF Driver Register Gcode and handling
- add M223 - Set TOF Driver Register Gcode and handling
- add M224 - Enable TOF Driver Gcode and handling

## Reviewer

Feedback on the cmake changes would be great.

## Risk assessment

Low, unreleased

[EXEC-971]: https://opentrons.atlassian.net/browse/EXEC-971?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EXEC-1166]: https://opentrons.atlassian.net/browse/EXEC-1166?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EXEC-1170]: https://opentrons.atlassian.net/browse/EXEC-1170?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EXEC-1171]: https://opentrons.atlassian.net/browse/EXEC-1171?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EXEC-1172]: https://opentrons.atlassian.net/browse/EXEC-1172?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


stm32-modules/flex-stacker/firmware/system/i2c_hardware.c
```diff
@@ -1,5 +1,9 @@
 #include <stdint.h>
 
+#pragma GCC push_options
+#pragma GCC optimize("O0")
```
sfoster1: we sure about this one? What problem is it solving?

stm32-modules/flex-stacker/firmware/tof_sensor/tmf8820_image.c
```diff
@@ -0,0 +1,227 @@
+/* http://srecord.sourceforge.net/ */
```
sfoster1: this will work, but I would love if we could do something like the following:
- Have the image in the repo as a binary, straight up
- do one of the following:
  - generate a source file like this from the binary image using a python script or using `xxd` or something, integrated as a cmake task, in a similar way to how we handle versions
  - pack it into the linked binary directly

stm32-modules/flex-stacker/firmware/tof_sensor/tof_sensor_hardware.c
```diff
@@ -0,0 +1,72 @@
...
+    GPIO_InitStruct.Mode = GPIO_MODE_OUTPUT_PP;
+    GPIO_InitStruct.Speed = GPIO_SPEED_FREQ_LOW;
+    GPIO_InitStruct.Pull = GPIO_NOPULL;
+    HAL_GPIO_Init(EEPROM_WP_PORT, &GPIO_InitStruct);
+}
+
+/**
+ * @brief enable the TOF X/Z write protect pin.
+ */
+void tof_write_protect_init(void) {
```
sfoster1: this is an enable, not a write protect, right?

stm32-modules/flex-stacker/firmware/tof_sensor/tof_sensor_hardware.c
```diff
@@ -0,0 +1,72 @@
...
+ * enable/disable writing to the eeprom.
+ */
+void enable_eeprom_write(bool enable) {
+    HAL_GPIO_WritePin(EEPROM_WP_PORT, EEPROM_WP_PIN, enable ? GPIO_PIN_RESET : GPIO_PIN_SET);
+}
+
+/**
+ * enable/disable writing to a tof sensor.
+ */
+void enable_tof_sensor_write(TOFSensorID sensor_id, bool enable) {
```
sfoster1: similarly, this puts the sensor into/takes it out of standby mode, right? It doesn't control writing.

stm32-modules/flex-stacker/firmware/tof_sensor/tof_sensor_hardware.c
```diff
@@ -0,0 +1,72 @@
...
+    if (sensor_id == TOF_X) {
+        HAL_GPIO_WritePin(TOF_EN_X_PORT, TOF_EN_X_PIN, enable ? GPIO_PIN_SET : GPIO_PIN_RESET);
+    } else if (sensor_id == TOF_Z) {
+        HAL_GPIO_WritePin(TOF_EN_Z_PORT, TOF_EN_Z_PIN, enable ? GPIO_PIN_SET : GPIO_PIN_RESET);
+    }
+}
+
+// Initialize the eeprom and tof enable pins
+void tof_hardware_init(void) {
+    eeprom_write_protect_init();
```
sfoster1: why are we binding the eeprom stuff into the tof stuff? they're not really connected, right? they're not even on the same bus?

stm32-modules/include/flex-stacker/flex-stacker/host_comms_task.hpp
```diff
@@ -67,33 +69,31 @@ class HostCommsTask {
...
+    using GetTOFRegisterCache = AckCache<8, gcode::GetTOFRegister>;
 
   public:
     static constexpr size_t TICKS_TO_WAIT_ON_SEND = 10;
     explicit HostCommsTask(Queue& q, Aggregator* aggregator)
         : message_queue(q),
           task_registry(aggregator),
           // These nolints are because if you don't have these inits, host
-          // builds complain NOLINTNEXTLINE(readability-redundant-member-init)
+          // builds conplain
```
sfoster1:
```suggestion
          // builds complain
```

stm32-modules/include/flex-stacker/flex-stacker/tmf8821.hpp
```diff
@@ -0,0 +1,457 @@
...
+    STAT_ERR_RANGE,
+    STAT_ERR_MORE,
+};
+
+// Bootloader Payload constants
+constexpr uint8_t BL_DATA_LEN = 128;  // chunk len
+constexpr uint8_t BL_HEADER_LEN = 2;  // cmd + len
+constexpr uint8_t BL_FOOTER_LEN = 1;  // checksum
+constexpr uint8_t BL_BUFFER_LEN = BL_HEADER_LEN + BL_DATA_LEN + BL_FOOTER_LEN;
+uint8_t BUFFER[BL_BUFFER_LEN] = {0};
```
sfoster1: We don't want this in the header file; it's fine right now because I guess this header is only in one translation unit, but were we ever to include it in another translation unit we'd get linker duplicate-symbol errors because we're violating the one-definition rule. We can declare it here but we need to define it in a source file or inside a class def or something.

stm32-modules/include/flex-stacker/flex-stacker/tmf8821.hpp
```diff
@@ -0,0 +1,457 @@
...
+constexpr uint8_t STAT_ERR_UNKNOWN_CID = 0x09;
+constexpr uint8_t STAT_WARNING_CONFIG_SPAD_1_NOT_ACCEPTED = 0x0A;
+constexpr uint8_t STAT_WARNING_CONFIG_SPAD_2_NOT_ACCEPTED = 0x0B;
+constexpr uint8_t STAT_WARNING_OSC_TRIM_NOT_ACCEPTED = 0x0C;
+constexpr uint8_t STAT_WARNING_I2C_ADDRESS_NOT_ACCEPTED = 0x0D;
+constexpr uint8_t STAT_ERR_UNKNOWN_MODE = 0x0E;
+
+class TMF8821 {
+  public:
+    TMF8821(TOFSensorPolicy* policy) : _policy(policy) {}
```
sfoster1: why do we need the policy provided in both the ctor and the `initialize` function? 

also we should make any one-argument ctor `explicit`

stm32-modules/include/flex-stacker/flex-stacker/tmf8821.hpp
```diff
@@ -0,0 +1,457 @@
...
+        auto reg = registers.enable;
+        reg.padding_1 = 0;
+        return set_register(reg, sensor_id).has_value();
+    }
+
+    auto write(TOFSensorID sensor_id, uint16_t reg, uint32_t* data,
+               int size = 1) -> std::optional<RegisterSerializedType> {
+        using RT = std::optional<RegisterSerializedType>;
+        auto dev_address = get_sensor_i2c_address(sensor_id);
+        auto [res, _] =
```
sfoster1: this syntax is so good

stm32-modules/include/flex-stacker/flex-stacker/tmf8821.hpp
```diff
@@ -0,0 +1,457 @@
...
+            _policy->i2c_write(dev_address << 1, reg, (uint8_t*)data, size);
+        if (res != 0) return RT();
```
sfoster1: braces please

stm32-modules/include/flex-stacker/flex-stacker/tmf8821.hpp
```diff
@@ -0,0 +1,457 @@
...
+        return RT(res);
+    }
+
+    auto read(TOFSensorID sensor_id, uint32_t reg, int size = 1)
+        -> std::optional<RegisterSerializedType> {
+        using RT = std::optional<RegisterSerializedType>;
+        auto dev_address = get_sensor_i2c_address(sensor_id);
+        auto [res, data] = _policy->i2c_read(dev_address << 1, reg, size);
+        if (res != 0) return RT();
```
sfoster1: braces please

stm32-modules/include/flex-stacker/flex-stacker/tmf8821.hpp
```diff
@@ -0,0 +1,457 @@
...
+        auto appid = static_cast<tmf8821::AppID>(ret.value()).appid;
+        return static_cast<TOFSensorMode>(appid);
+    }
+
+    auto reset_custom_address() -> void { _custom_address = false; }
+
+    // Gets the sensor i2c address
+    auto get_sensor_i2c_address(TOFSensorID sensor_id, bool default_addr = true)
+        -> uint16_t {
+        if (default_addr && !_custom_address) return TOF_DEFAULT_ADDRESS;
```
sfoster1: braces for all of these please

stm32-modules/include/flex-stacker/flex-stacker/tmf8821.hpp
```diff
@@ -0,0 +1,457 @@
...
+            return TOFSensorMode::UNKNOWN;
+        }
+        auto appid = static_cast<tmf8821::AppID>(ret.value()).appid;
+        return static_cast<TOFSensorMode>(appid);
+    }
+
+    auto reset_custom_address() -> void { _custom_address = false; }
+
+    // Gets the sensor i2c address
+    auto get_sensor_i2c_address(TOFSensorID sensor_id, bool default_addr = true)
```
sfoster1: I'm a little confused by the logic with having both the `default_addr` param and the built-in `_custom_address` field - why do we need both?

stm32-modules/include/flex-stacker/flex-stacker/tmf8821.hpp
```diff
@@ -0,0 +1,457 @@
...
+        }
+
+        // Set change address I2C_ADDR_CHANGE (0x3E) to 0
+        if (!set_register(_registers.i2c_addr_change, sensor_id).has_value()) {
+            return false;
+        }
+
+        // Write the config page
+        auto len = prepare_cmd_frame(CMD_WRITE_CONFIG_PAGE, 0, 0);
+        if (!write(sensor_id, _registers.cmd_stat.address, (uint32_t*)BUFFER,
```
sfoster1: if this is what we're using `BUFFER` for then it definitely should be a class member. There's definitely going to be 2 of these class instances, which means that they could stomp on each other if they're sharing the same buffer. Putting it in the class will avoid that.

sfoster1: also the buffer size is defined in bytes and is `131` which definitely does not align with punning it to an array of uint32s, we'd get a guaranteed overflow if we ever wrote to that final word

stm32-modules/include/flex-stacker/flex-stacker/tmf8821.hpp
```diff
@@ -0,0 +1,457 @@
...
+        value &= Reg::value_mask;
+        return value;
+    }
+
+    template <tmf8821::TMF8821Register Reg>
+    requires ReadableRegister<Reg>
+    auto read_register(TOFSensorID sensor_id) -> std::optional<Reg> {
+        using RT = std::optional<Reg>;
+        auto ret = read(sensor_id, Reg::address, 1);
+        if (!ret.has_value()) return RT();
```
sfoster1: braces please

stm32-modules/include/flex-stacker/flex-stacker/tmf8821.hpp
```diff
@@ -0,0 +1,457 @@
...
+    }
+
+    template <tmf8821::TMF8821Register Reg>
+    requires WritableRegister<Reg>
+    auto set_register(Reg reg, TOFSensorID sensor_id) -> std::optional<Reg> {
+        using RT = std::optional<Reg>;
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
+        auto value = get_register_value(reg);
+        auto ret = write(sensor_id, Reg::address, &value, 1);
+        if (!ret.has_value()) return RT();
```
sfoster1: braces please

stm32-modules/include/flex-stacker/flex-stacker/tmf8821.hpp
```diff
@@ -0,0 +1,457 @@
...
+        }
+
+        // write firmware in 128 byte chunks to sensor
+        uint8_t chunk_size = 0;
+        for (uint32_t i = 0; i < tmf8820_image_length; i += BL_DATA_LEN) {
+            chunk_size = (i + BL_DATA_LEN <= tmf8820_image_length)
+                             ? BL_DATA_LEN
+                             : (tmf8820_image_length - i);
+            // send chunk
+            if (!bl_send_image_chunk(sensor_id, (uint8_t*)&tmf8820_image[i],
```
sfoster1: imo pointer arithmetic is a bit easier to follow than this but that's a light opinion

stm32-modules/include/flex-stacker/flex-stacker/tof_sensor_task.hpp
```diff
@@ -0,0 +1,202 @@
...
+    auto run_once(Policy* policy) -> void {
+        if (!_task_registry) {
+            return;
+        }
+
+        if (!_initialized) {
+            _policy = policy;
+
+            // Disable both sensors before initializing
+            if (!_tof_sensor_x.ok) _policy->enable_tof_sensor(TOF_X, false);
```
sfoster1: braces please

sfoster1 (changes requested):
Some more minor things, but while the structure of this looks good we really shouldn't use a file-scope global as a shared buffer between two objects and I would love to see the tof firmware be included a little more nicely.

stm32-modules/flex-stacker/firmware/system/i2c_hardware.c
vegano1: nope, this was for debugging, will remove it.

stm32-modules/include/flex-stacker/flex-stacker/tmf8821.hpp
vegano1: The policy in the ctor is initialized as a nullptr, then passed in the `initialize` function to set the pointer to the actual policy instance that we can reference within the class. Would it be better to pass in policy at compile time as a template parameter?

stm32-modules/flex-stacker/firmware/tof_sensor/tof_sensor_hardware.c
vegano1: removed

stm32-modules/include/flex-stacker/flex-stacker/tmf8821.hpp
vegano1: removed from ctor since initializing as nullptr.

stm32-modules/include/flex-stacker/flex-stacker/tmf8821.hpp
vegano1: Good call, `default_addr` is redundant.

sfoster1 (approved):
Looks good to me!

ahiuchingau (approved):
The sensor hardware, tasks and policy structure looks good to me! Thank you

PR #504, opened 2025-02-13
chore(stacker): add DVT board revision

ahiuchingau:
This PR makes the default stacker revision to be `b1` as we prepare for DVT testing and changes. Low risk as no other firmware changes are included in PR.


sfoster1 (approved):
Hooray

PR #508, opened 2025-03-03
build(fix): update actions/cache and actions/checkout to v4

vegano1:
actions/checkout@v2 and actions/cache@v2 are deprecated, update them to v4


ahiuchingau (approved):
Cool!

PR #510, opened 2025-03-04
fixes for new ubuntu

ryanthecoder:
Collection of fixes to get CI working again on ubuntu 24.04
Things that are now in Fashion:

**Concatenating namespaces:**
`namespace A {
namespace B {
} 
}`
Becomes 
`namespace A:B {
}`

**Using Const within functions when a variable is not changed (const-everywhere)**

**files should now directly `#include` the header where things are defined and not rely on indirect includes**

**remove includes that aren't directly used in that file.**

**You should no longer directly reference the return type in a `return` statement:**
`auto foo() -> Bar {
  return Bar("init-var");
 }` 
 Becomes
 `auto foo() -> Bar {
  return {"init-var"};
 }` 
 
 **enums need an explicit type, and it should be the smallest type that can contain that enum**
 `enum myEnum { A, B, C}` becomes `enum myEnum : uint8_t { A, B, C}`
 
 **functions that are already implicitly inlined should not be labeled inline**
 this is true for us mostly when you define a function within a class definition in the header file.
 
**Avoid C style pointer references in favor of class functions**
 `&myArray[0]` becomes `myArray.data()`

**don't initialize class member variables with default values in the constructor, do it in the definition**
```
class Foo {
  public:
    Foo(uint8_t val)
      : my_val(val),
      default_val(false) {} // Don't init defaults here
  private:
    uint8_t my_val;
    bool default_val;
 }
```

 becomes
```
class Foo {
  public:
    Foo(uint8_t val)
      : my_val(val) {}
  private:
    uint8_t my_val;
    bool default_val{false}; // init defaults in member definitions
 }
```
 
 **No more nested tertiary statements**
 
 **No more redundant instantiation of class members**
```
class Foo {
 public:
    Foo() {}
 private:
    Bar my_bar{}; // redundant call to constructor when {} is here
 }
```
 Becomes
```
class Foo {
 public:
    Foo() {}
 private:
    Bar my_bar;
 }
```
 
 **No more accessing optional values without checking**
 
`auto x = optional.value();`
becomes
```
if (optional.has_value()) {
    auto x = optional.value();
}
```


ahiuchingau:
🙏 🙏 🙏

ahiuchingau (approved):
Thank you!!

vegano1 (approved):
Heck yeah, thank you!

PR #511, opened 2025-03-06
feat(stacker): add a new error code and fix error message handling

ahiuchingau:
This PR introduces two improvements to the stacker:

1. Moves will now terminate immediately, and a UnexpectedLimitSwitch error will be raised if the limit switch in the direction of the move is triggered. This allows us to detect when we’ve reached the end of the axis right away, instead of waiting for a stall condition to occur.

2. Currently, when a stall or estop occurs, an asynchronous error message is sent to the host, and the move itself is acked separately as if it completed successfully. This PR ensures that the motor task responds directly to the move message with an ack containing the appropriate error information, so the host gets accurate feedback about the failure.

I also modified the error string by adding an `OK\n` at the end of the sync error message, since that appears to be what’s expected on the [Python side](https://github.com/Opentrons/opentrons/blob/ecb3fef2800fb0ecb1278373e6e2dd82a0151f6b/api/src/opentrons/drivers/asyncio/communication/serial_connection.py#L467C13-L467C21). But I kept the async error the same with only an "\n" at the end.

It’s not entirely clear if this is the intended behavior, since this pattern isn’t consistent across all modules. For example, the heater-shaker includes OK at the end of all errors, but the thermocycler and tempdeck do not. I’m planning to do some refactoring on the Python side to better standardize how stacker errors are handled.


sfoster1 (approved):
nice, ty!

vegano1:
I thought the reason why we don't want to terminate a sync error was to accumulate them in the case of many errors until we got an ok response.

PR #512, opened 2025-03-10
chore(stacker): update motor params for DVT

ahiuchingau:
Changing the default speeds/accelerations for motor axes, as well as updating default hold currents


vegano1 (approved):
Very nice, thank you.
Let's also disable stallguard for the L axis since we can't use it.
set `.diag0_stall = 0` and `sgt = 0`

PR #513, opened 2025-03-13
fix(stacker): check for estop and stall inside motor interrupt

ahiuchingau:
My changes from a previous PR tried to stop the motor interrupt from the motor task when the estop/stallguard is triggered... which just doesn't work! We should just check for those inside the interrupt like we were before and set the error when it occurs.


sfoster1 (approved):

vegano1 (approved):
thank you

PR #514, opened 2025-03-14
fix(flex-stacker): actually set short-range mode for the TOF sensors.

vegano1:
We were not setting the short-range mode for the TOF sensors because we were returning after setting the short-range register, but were not writing the change with the write config page command.


ahiuchingau (approved):
LGTM

PR #515, opened 2025-03-14
fix(stacker): enable IRQ handler for limit switches

ahiuchingau:
We now read the limit switch input every motor interrupt tick and it's significantly slowing the motor movements. This PR fixes that by setting the relevant limit switch as an external GPIO interrupt to alert the motor interrupt when it is triggered.


sfoster1 (approved):
cool cool

vegano1 (approved):
wow thats insane.

PR #516, opened 2025-03-17
fix(stacker): conditionally send acknowledgments for `SetStatusBarStateMessage`

ahiuchingau:
This PR updates the logic in the `UITask` class to conditionally send acknowledgments based on this new `from_host` field in the `SetStatusBarStateMessage`. This should reduce the occurrence of the `ERR005:bad message acknowledgement` error sent from the module.


vegano1 (approved):
good catch!

PR #517, opened 2025-03-17
fix(flex-stacker): various fixes for TOF sensor configuration.

vegano1:
# Overview

There were some issues with the histogram data we were collecting, it turns out some of the configurations we were using were not getting set.

# Changelog

- calculate the byte size to read/write from/to a register instead of a fixed 1-byte.
- fix an issue when setting the spad_size, the registers are 0x8F and 0x90 so they can't be written sequentially.
- consolidate the tof sensor configurations into a `configure_sensor` function.
- wait for the STAT_OK state when writing the config page
- always send the change config page as it resets after usage
- change how the active register is set by issuing value to cmd_stat 0x08

# Testing

- [x] Make sure that the tof sensors are configured on startup by validating the following (active_range, hist_dump, kila_iterations, report_period)
- [x] Make sure the tof histogram matches the data collected for EVT units
- [x] Make sure the tof histogram shows 1 tiprack detected around the 40-80 bin range

### DIFF
--- a/stm32-modules/include/flex-stacker/flex-stacker/tmf8820.hpp
+++ b/stm32-modules/include/flex-stacker/flex-stacker/tmf8820.hpp
@@ -185,11 +185,9 @@ class TMF8820 {
         }
 
         // Set sensor configuration
-        set_sensor_report_period(sensor_id, _config->report_period_ms);
-        set_sensor_kilo_iterations(sensor_id, _config->kilo_iterations);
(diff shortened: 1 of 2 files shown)


vegano1:
1 Tiprack labware detected at around 60-80 bin range

Screen Shot 2025-03-17 at 7 14 57 PM

ahiuchingau (approved):
Able to get data

PR #518, opened 2025-03-18
fix(stacker): latch limit switch responded even when not prompted

ahiuchingau:
This is a quick fix to enable dvt QC. We should diagnose why latch limit switch interrupt is behaving incorrectly when we have more time.


vegano1 (approved):
ok

PR #520, opened 2025-03-28
fix(flex-stacker): add correct PRIMARY_REVISION define to cmake to check install detect board rev.

vegano1:
# Overview

Install detect was introduced in rev b, so let's introduce PRIMARY_REVISION and SECONDARY_REVISION defines to check the board rev at compile time.

# Changes

- Add PRIMARY_REVISION and SECONDARY_REVISION to cmake
- Check PRIMARY_REVISION when compiling in INSTALL detect for rev b of flex stacker

# Testing

- Tested and works in sz with DVT Flex Stackers


djgomez-opentrons (approved):
Tested with DVT boards in SZ

PR #522, opened 2025-04-20
fix(flex-stacker): Various fixes + improvements for the Flex Stacker UI task and statusbar.

vegano1:
# Overview

The Flex Stacker UI task animation tick was done in the visit_message function, it was powered by a FreeRTOS timer sending a `UpdateUIMessage` message every n milliseconds. This resulted in a few fundamental issues, Firstly, the animations would stop if the motor task (with a hw interrupt) ticked, which resulted in out-of-sync states. Secondly, because the ui task tick was a message in a message queue, incoming messages from the host would sometimes timeout or get queue full messages because the queue was often filled with `UpdateUIMessage` messages, so the task could not respond. We also wanted to make sure you can tell when the stacker is done initializing, so we added back in the ability to set the statusbar internally. This pull request goes in conjunction with this [Opentrons/opentrons#18119](https://github.com/Opentrons/opentrons/pull/18119) pull request to sync statusbar state when connecting to a Flex.

# Changelog

- Remove message queue tick mechanism for the UITask and call the animation tick function (update_ui) directly in the FreeRTOS timer.
- Re-added `from_host` param to the `SetStatusBarStateMessage` message so we can set the statusbar led internally
- Pulse the statusbar LEDs white while the flex stacker is initializing, they will get set to static white (idle) by the tof sensor task or whatever state the Flex is in if connected.
- Unconditionally set the statusbar LED internally when the Tof sensors have initialized.
- Convert the ChannelMapping from a 4-channel array to a 16-channel array, which lets us create color combinations like yellow. This fixes an issue where only 2 of the 4 LEDs were used to create yellow, causing the statusbar to look terrible for color combinations.
- Only enable/disable TOF sensors in `EnableTOFSensorMessage` handler if the expected state differs from the actual state.
- Remove unused FreeRTOS timer variables from the motor task
- Use reference instead of redundant pointer when using the `get_statusbar_state` function. 

# Testing

- [ ] Power on the Flex Stacker and make sure the status bars are pulsing white while the tof sensors are initializing.
- [ ] Press the Estop on the Flex and make sure all channels are used when setting color combinations like yellow.
- [ ] In a protocol, while the Flex Stacker is dispensing, pause the protocol, which causes the statusbar to pulse blue. Make sure the Flex Stacker statusbar pulses blue WHILE dispensing/storing labware.

# Review Requests


sfoster1 (approved):
Looks good to me!

stm32-modules/include/flex-stacker/flex-stacker/tof_sensor_task.hpp
```diff
@@ -170,17 +174,19 @@ class TOFSensorTask {
...
-                sensor.driver.initialize(&sensor.config, _policy, sensor.kind);
-            sensor.state = sensor.ok ? IDLE : TOF_ERROR;
+        if (m.enable != sensor.ok) {
+            TOFSensorPolicy::enable_tof_sensor(m.sensor_id, m.enable);
+            sensor.driver.reset_custom_address();
+            sensor.state = DISABLED;
+            sensor.ok = false;
+            if (m.enable) {
+                sensor.state = INITIALIZING;
+                sensor.ok = sensor.driver.initialize(&sensor.config, _policy,
```
sfoster1, line 187: is this the thing that blocks the sensor task while it's running? can we set a freertos flag here and check it from the host comms task to send an error response if the host queries while this is happening so we don't fill the cache?

sfoster1 (changes requested):
Actually, after looking at the python side of things, IMO we need to do somethign so the expected case of "query while initializing" isn't "the gcode cache fills up until there's an error". Maybe add something to the sensor task that the host comms task can query - a freertos mutex or semaphore or something - and send back an early error?

stm32-modules/include/flex-stacker/flex-stacker/tof_sensor_task.hpp
vegano1: It is the `initialize` function, but when the tof_sensor task is first starting up in `run_once` entry function on line 102. But I agree, that is a better approach, will do.

ahiuchingau (commented):
If we're implementing a FreeRTOS flag for error, can we also implement one for update the UI task? I think maybe we should try to avoid as much as we can inter-task messages since the host will be sending them too?

vegano1:
> If we're implementing a FreeRTOS flag for error, can we also implement one for update the UI task? I think maybe we should try to avoid as much as we can inter-task messages since the host will be sending them too?

Yeah all tasks will have to notify the host comms task once they initialize, that being we should not have any issues sending inner-task messages, this was only a problem with the ui task.

ahiuchingau:
Do you think we should wait to start the UI task until all the other tasks are done initializing?

vegano1:
> Do you think we should wait to start the UI task until all the other tasks are done initializing?

no, we want to make comms available as soon as possible, but then give status responses when responding to a message from the host. Ths response would say something like "initializing" or similar.

stm32-modules/include/common/core/queue_aggregator.hpp
```diff
@@ -315,6 +327,25 @@ class QueueAggregator {
         return std::get<get_queue_idx<Queue>()>(_handles)._handle != nullptr;
     }
 
+    /**
+     * @brief Executes a function on a QueueHandle with the given index.
+     *
+     * @tparam Tuple The tuple to iterate through, the QueueHandles in this
+     * case.
+     * @return true if the index matches the queue index.
```
sfoster1: this isn't true though, right? Returns `f(tuple_member)`  if a match is found, or `false` if no match is found. 

started to go down a rabbit hole of making the return type of `dispatch_at_index` depend on the functor and then realized, as you probably did, that you would have to do it with a helper type and argument overloading to handle the index-doesn't-match case. yuck.

Either wya, maybe make the return value `std::optional<bool>` so you can tell them apart?

stm32-modules/include/flex-stacker/flex-stacker/errors.hpp
```diff
@@ -15,6 +15,7 @@ enum class ErrorCode : uint16_t {
     GCODE_CACHE_FULL = 4,
     BAD_MESSAGE_ACKNOWLEDGEMENT = 5,
     ESTOP_TRIGGERED = 6,
+    TASK_INITIALIZING = 7,
```
sfoster1: `TASK_NOT_READY`?

stm32-modules/include/common/firmware/freertos_message_queue.hpp
```diff
@@ -75,10 +75,14 @@ class FreeRTOSMessageQueue {
...
+    auto set_ready() -> void { ready = true; }
+    [[nodiscard]] auto task_ready() -> bool { return ready; }
+
   private:
     StaticQueue_t queue_control_structure;
     std::array<uint8_t, queue_size * sizeof(Message)> backing;
     QueueHandle_t queue;
     TaskHandle_t receiver_handle;
     uint8_t sent_bit;
+    bool ready = false;
```
sfoster1: This should be `std::atomic` or even a freertos [url=https://www.freertos.org/Documentation/02-Kernel/02-Kernel-features/06-Event-groups]event[/url] i think.

sfoster1 (approved):
Looks good (always love template recursion) but the `ready` flag needs a touch more atomicness

stm32-modules/include/common/core/queue_aggregator.hpp
vegano1: I went the easier route of depending on the function call itself to determine if the item at the index was found. Something like this

```c++
    template <std::size_t I = 0, typename Tuple, typename Func>
    auto dispatch_at_index(Tuple&& tpl, std::size_t index, Func&& f) -> void {
        if constexpr (I < std::tuple_size_v<std::decay_t<Tuple>>) {
            if (I == index) {
                f(std::get<I>(std::forward<Tuple>(tpl)));
                return;
            }
            return dispatch_at_index<I + 1>(std::forward<Tuple>(tpl), index,
                                            std::forward<Func>(f));
        }
    }
```
Now the caller can do something like this, where the function is called if the item exists
```c++
    auto task_ready(size_t idx) -> bool {
        dispatch_at_index(_handles, idx, [](auto& queue) {
            return queue._handle->task_ready();
        });
        return false;
    }
```
This delegates the return type to the caller function, in this case `task_ready` could also return optional<bool> but in this context I don't care about the difference between the task queue is not yet registered (index not found) vs the task is initializing (the return value of `task_ready`).

stm32-modules/include/common/firmware/freertos_message_queue.hpp
vegano1: cool, made it std::atomic<bool>

PR #523, opened 2025-05-06
fix(thermocycler): clear cache after motor-related errors

caila-marashaj:
## Overview
On the thermocycler, not clearing the message cache after errors can cause weird behavior on subsequent protocol runs, unless you power cycle the module. This can be fixed by clearing the cache after motor-related errors.


sfoster1 (approved):
Sounds good to me, I think, but can you expand on why we need it?

PR #524, opened 2025-05-12
feat(thermocycler): check lid status again after extending seal

caila-marashaj:
## Overview
In most of the remaining cases where the thermocycler lid is opened unexpectedly, it's caused by the seal pushing against the latch when the lid isn't really all the way closed. We can catch this and throw an error by checking the lid status one more time after the seal is extended.


sfoster1 (approved):
Nice, thank you!

ahiuchingau (approved):
LGTM once tested on hardware

ahiuchingau (approved):
The case is CLOSED

PR #525, opened 2025-05-15
chore(stacker): update motion params based on DVT testing findings

ahiuchingau:
## Changelog
1. Update default motor currents
2. Increase motor Z brake delay from 70ms to 100ms

This will also need a monorepo side PR to update the motor currents


vegano1 (approved):
thank you

PR #526, opened 2025-05-15
fix(flex-stacker): add atomics + remove the UNEXPECTED_LIMIT_SWITCH error from the motor interrupt.

vegano1:
Sometimes, the limit switch triggers when it's not supposed to, specifically during a Fixed movement, when we raise an UNEXPECTED_LIMIT_SWITCH error. We are not sure why this happens yet and need more time to investigate. For now, let's disable raising this error and use atomics for the variables to prevent OOO and irq issues.

Closes: [RABR-767](https://opentrons.atlassian.net/browse/RABR-767)

TODO:
Look into clearing the pending bit in EXTI, maybe there's a pending irq that triggers.

[RABR-767]: https://opentrons.atlassian.net/browse/RABR-767?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/stm32-modules/flex-stacker/firmware/motor_control/motor_hardware.c
+++ b/stm32-modules/flex-stacker/firmware/motor_control/motor_hardware.c
@@ -415,7 +415,9 @@ void hw_enable_lim_switch_irq(MotorID motor_id, bool direction) {
         return;
     }
     stepper_hardware_t motor = get_motor(motor_id);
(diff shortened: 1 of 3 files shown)


ahiuchingau (approved):
if it works, it works!

PR #527, opened 2025-06-11
fix(flex-stacker): use stall and estop interrupts instead of checking GPIO every motor tick

ahiuchingau:
We already have the motor driver `Diag0`and `estop` input as interrupts so we should make use of them to inform our motor timer interrupt logic, instead of checking the pins every time the timer ticks.

### DIFF
--- a/stm32-modules/flex-stacker/firmware/motor_control/freertos_motor_task.cpp
+++ b/stm32-modules/flex-stacker/firmware/motor_control/freertos_motor_task.cpp
@@ -78,6 +78,12 @@ static auto _top_task = motor_task::MotorTask(
     }
 }
 
+[[nodiscard]] static auto stall_callback_glue() {
+    // L motor does not have a stall detection mechanism
+    x_motor_interrupt.stall_detected();
+    z_motor_interrupt.stall_detected();
+}
+
 auto run(tasks::FirmwareTasks::QueueAggregator* aggregator) -> void {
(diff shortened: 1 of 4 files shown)


stm32-modules/flex-stacker/firmware/motor_control/freertos_motor_task.cpp
```diff
@@ -78,6 +78,18 @@ static auto _top_task = motor_task::MotorTask(
     }
 }
 
+[[nodiscard]] static auto stall_callback_glue() {
+    // L motor does not have a stall detection mechanism
+    x_motor_interrupt.stall_detected();
+    z_motor_interrupt.stall_detected();
+}
```
vegano1, line 85: Lets pass in the motor_id here so we know exactly which axis stalled

stm32-modules/flex-stacker/firmware/system/stm32g4xx_it.c
```diff
@@ -1,24 +1,25 @@
 /**
-  ******************************************************************************
-  * @file    stm32g4xx_it.c
-  * @brief   Interrupt Service Routines.
-  ******************************************************************************
-  * @attention
-  *
-  * Copyright (c) 2024 STMicroelectronics.
```
vegano1, line 8: weird formating

vegano1 (approved):
Cool, this should work, will need some hardware testing though.

PR #528, opened 2025-06-16
feat(tempdeck + magdeck): add generic handling for M114 gcode message

caila-marashaj:
## Overview
We're adding a new gcode message for all of the STM32 modules- `GetResetReason`. This message is intended to prompt modules to return the value of the flag the MCU sets on startup that contains the reason for the last shutdown. Since arduinos don't have this functionality, we should instead just add generic handling for the gcode message

## Changelog
- add M114 gcode definition to tempdeck + magdeck
- create a generic response function in the arduino code. I added a string argument that can just be an empty string in case we want to be more specific about our generic responses in the future

## Review Requests
The magdeck has a gcode message that's `M114.2`. That shouldn't interfere with `M114` right?


sfoster1 (commented):
let's .gitignore out those added files in `cmake-build-debug` but other than that looking good

sfoster1 (changes requested):
wrong button

vegano1 (changes requested):
@caila-marashaj @sfoster1 

This should probably be done at the Python layer because if a specific module does not implement a code, I don't think the Gcode needs to be defined by said module and no-op on the firmware. Instead, there should be a generic response of "We don't support that code," which I think should already be here with the no-response case. Then, we make sure we don't send unsupported codes from the Python driver, which we already do, since the supported module gcodes are passed into the module object when constructed [here](https://github.com/Opentrons/opentrons/blob/f8ee274ec7d4b8fc70c787ebfda5cbc38518b66f/api/src/opentrons/drivers/temp_deck/driver.py#L30).

Since the tempdeck and magdeck don't support the GET_RESET_REASON "M114" functionality, I think we should remove it from the Python driver instead of no-op here. 

Let me know what you think.

sfoster1:
I think it's reasonable to print a generic response of some kind for unknown gcodes, rather than just swallow them and not respond at all. i agree we should do that in a `default:` rather than explicit `case:`. I think we should do this here for reset reasons, and maybe also remove it from the python.

vegano1:
> I think it's reasonable to print a generic response of some kind for unknown gcodes, rather than just swallow them and not respond at all. i agree we should do that in a `default:` rather than explicit `case:`. I think we should do this here for reset reasons, and maybe also remove it from the python.

cool, then it would make sense to match the response to the other stm32-modules, something like [this](https://github.com/Opentrons/opentrons-modules/blob/6e136471fd59e20fe5999ed0210d1183b8dbbc82/stm32-modules/thermocycler-gen2/src/errors.cpp#L8)?
```const char* const UNHANDLED_GCODE = "ERR003:unhandled gcode OK\n";```

sfoster1 (approved):
This looks good and correct to me if it works!

vegano1 (approved):

PR #529, opened 2025-06-25
chore(stacker): update default board rev to B2 for PVT

ahiuchingau:
Adding B2 as PVT board rev.


djgomez-opentrons (approved):
Looks good

PR #530, opened 2025-08-26
fix(flex-stacker): Explicitly enable the limit switch IRQ on boot to clear any pending IRQ held after a FW update.

vegano1:
## Overview

The flex stacker was failing on the first dispense after a firmware update, which is caused by an interrupt not getting cleared between boots. Which is not usually a problem when cold booting, but when doing a firmware update (which should reset everything !), an interrupt line was being held, and the -Z limit switch was triggering when it was not supposed to. This translates to the stacker failing a dispense the first time after an update; however, subsequent dispenses are passing successfully.
 
Closes: [RQA-4536](https://opentrons.atlassian.net/browse/RQA-4536) [RABR-822](https://opentrons.atlassian.net/browse/RABR-822)

## Change log

- enable limit switch irq on boot to clear lingering signals

[RQA-4536]: https://opentrons.atlassian.net/browse/RQA-4536?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RABR-822]: https://opentrons.atlassian.net/browse/RABR-822?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/stm32-modules/flex-stacker/firmware/motor_control/motor_hardware.c
+++ b/stm32-modules/flex-stacker/firmware/motor_control/motor_hardware.c
@@ -187,39 +187,32 @@ void motor_hardware_gpio_init(void){
     // Z MOTOR
     init.Pin = Z_MINUS_LIMIT_PIN;
     HAL_GPIO_Init(Z_MINUS_LIMIT_PORT, &init);
-
+    HAL_NVIC_SetPriority(EXTI3_IRQn, 5, 0);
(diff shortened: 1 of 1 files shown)


sfoster1 (approved):
Gross, nice fix!

PR #531, opened 2025-09-02
fix(build): Set the TRAVIS_TAG env var in GitHub Actions to deploy Arduino modules to the correct directory.

vegano1:
# Overview

The TRAVIS_TAG env was never set, so the `set_build_vars.sh` script always defaulted to `modules-magdeck`.

Closes: [EXEC-1749](https://opentrons.atlassian.net/browse/EXEC-1749)

## Note
Once we create new module release builds, we should fix the tempdeck source URL in [oe-core](https://github.com/Opentrons/oe-core/commit/5e3be70474bab12594953dd7b1eee6dfc9134277) and [buildroot](https://github.com/Opentrons/buildroot/commit/ae4b488380ae59853836d8bc057ec23c9f2c48ce).

[EXEC-1749]: https://opentrons.atlassian.net/browse/EXEC-1749?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/arduino-modules.yaml
+++ b/.github/workflows/arduino-modules.yaml
@@ -58,7 +58,4 @@ jobs:
           TRAVIS_TAG: ${{ github.ref_name }}
         run: |
           . ./scripts/set_build_vars.sh
-
-          echo ${RELEASE_LOCAL_DIR}
-          echo ${RELEASE_UPLOAD_DIR}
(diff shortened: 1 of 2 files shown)


sfoster1 (approved):
Looks good but in a followup can we make it not `TRAVIS_TAG` (lets merge this one as is so it doesn't have to change code)

PR #532, opened 2025-09-15
feat(thermocycler-gen2): Implement ramp rate gcodes

ryanthecoder:
This adds the SetRampRate gcode message and expands the SetPlateTemp message to have an additional optional ramp rate argument.

It also updates the firmware side handling of these to connect the ramp rate to the already existing layers that would use it in the temperature controller tasks.


sfoster1 (approved):
Looks good to me!

PR #533, opened 2025-09-16
heater-shaker: add error control

sfoster1:
Add some new gcodes to the heater-shaker that will 
- Allow for querying the error state to drive asynchronous module errors (M411)
- Allow for setting the error state manually and clearing it manually (M412.D, M413) so we can test it

The goal here is that the robot needs to be able to ask, "are you in an error state?" Right now, the only way the module will tell you it's in an error state is if you tell it to do something - and we don't want to do that in a background poll.

There are of course more changes than it perhaps seems like there should be. That's for a couple largely unrelated reasons:

- In fe7aa32bb9c27cab9f122f0908d38aad19368c11 we add the new gcodes and hook them up to the tests. This is where all the new functionality comes from, and it's pretty straightforward.
One weird thing in this PR is I added a semaphore to the ack cache; in the heater shaker, lots of the gcodes want to get responses from both the motor and the heater tasks, and we don't want invalid-ack messages coming from the second response.
- d0bdb4edbc93e9d83afd461394577a8ac7aec4aa is less straightforward; while writing heater task tests, I realized there were a bunch of pairs and triplets of lines of code that were repeated over and over and over, so I put them in helper functions, and then had to change a bunch of stuff
- a354f4285fd49a0f90fc3209e3fed605a3817dea same but for the motors
- bca6ab73977bb7c62887eba0f3258fe86b605b54 042e31bb32a5963594f7f4e4ddb679fd7b7ac48c and 8f56a66d2415e34d2cf9744ca49fcc7adac86ab5 add tests for the actual new gcodes

## Testing
- Put this on a module
- Put the module into an error state via `M412.D`
- Query the module state via `M411`
- Clear the error state via `M413`


vegano1 (approved):
Straight forward, looks good on my end.

PR #534, opened 2025-09-22
thermocycler-gen2: add error control

sfoster1:
Implement some new gcodes as a followup to #533 but this time for the thermocycler.

As with that PR, M411 gets errors; M412.D Ennn Snnn sets with a delay; M413 clears.

This is a bit more particular and annoying because thermocycler errors are implemented a little unevenly; there's a lot of attention paid to how they work with the thermal task, and this work slots into there, but the lid errors are more ad-hoc and the motor errors don't really persist at all, so we just don't involve them at all.

Closes EXEC-1704


caila-marashaj (approved):
makes sense- the ack should end whatever task is currently going on, right?

sfoster1:
> makes sense- the ack should end whatever task is currently going on, right?

This is not really related to the concurrent execution stuff; commands sent by or data sent to the concurrent commands would end them without these changes. These changes are necessary to monitor problems that happen while no command or task is running.

PR #535, opened 2025-09-30
feat(reference-module): add reference module project

vegano1:
# Overview

The STM32 modules follow a similar architecture, use the same bootloader, and have the same general project structure. We are in the process of creating a Vacuum Module, which again is very similar to the other STM32 modules. So instead of just creating a vacuum module project, let's first create a “reference module“ project that can be used as a starting point for other modules in the future.

Closes: [EXEC-1912](https://opentrons.atlassian.net/browse/EXEC-1912)

# Testing

- [x] Flash the "reference-module-nff" target on an STM32 PCB like the Flex Stacker board
- [x] Make sure the status LED is blinking

[EXEC-1912]: https://opentrons.atlassian.net/browse/EXEC-1912?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


stm32-modules/reference-module/firmware/host_comms_task/freertos_comms_task.cpp
```diff
@@ -0,0 +1,165 @@
...
+                               local_task->tx_buf.accessible()->end());
+        if (!top_task->may_connect()) {
+            usb_hw_stop();
+        } else if (tx_end != local_task->tx_buf.accessible()->data()) {
+            local_task->tx_buf.swap();
+            usb_hw_send(
+                // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
+                reinterpret_cast<uint8_t *>(
+                    local_task->tx_buf.committed()->data()),
+                tx_end - local_task->tx_buf.committed()->data());
```
sfoster1, line 77: from bitter recent experience add a `vTaskDelay(1);` after the call to `usb_hw_send()`, because otherwise if multiple messages are in the host comms queue to send, you can end up swapping the to-send buffer back _out_ of the usb hardware before the next 125us tick hits and the data actually goes to the wire.

sfoster1 (approved):
Looks good, but you can save some heartache with a little tweak to the usb send

stm32-modules/reference-module/firmware/host_comms_task/freertos_comms_task.cpp
vegano1: wow thats insane.

PR #536, opened 2025-09-30
add usb transmission backpressure

sfoster1:
Sometimes the subsequent acks get lots (or, the first one does - there's one ack transmitted). My guess is that's because the acks are sent too fast and the usb hardware doesn't manage to actually send the buffer.

The usb middleware exposes status information; use that status information to wait on the data actually sending.

This change probably works; it definitely doesn't break anything.

### DIFF
--- a/stm32-modules/heater-shaker/firmware/host_comms_task/freertos_comms_task.cpp
+++ b/stm32-modules/heater-shaker/firmware/host_comms_task/freertos_comms_task.cpp
@@ -162,6 +162,7 @@ void run(void *param) {  // NOLINT(misc-unused-parameters)
                         local_task->tx_buf.committed()->data()),
                     tx_end - local_task->tx_buf.committed()->data());
             }
+            vTaskDelay(1);
         }
     }
 }
--- a/stm32-modules/heater-shaker/firmware/host_comms_task/freertos_comms_task.cpp
+++ b/stm32-modules/heater-shaker/firmware/host_comms_task/freertos_comms_task.cpp
@@ -152,7 +152,17 @@ void run(void *param) {  // NOLINT(misc-unused-parameters)
                 reinterpret_cast<uint8_t *>(
                     local_task->tx_buf.committed()->data()),
                 tx_end - local_task->tx_buf.committed()->data());
-            USBD_CDC_TransmitPacket(&_local_task.usb_handle);
+            uint8_t tx_result =
+                USBD_CDC_TransmitPacket(&_local_task.usb_handle);
+            while (tx_result == USBD_BUSY) {
+                vTaskDelay(1);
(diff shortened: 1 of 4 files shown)


stm32-modules/heater-shaker/firmware/host_comms_task/freertos_comms_task.cpp
```diff
@@ -152,7 +152,21 @@ void run(void *param) {  // NOLINT(misc-unused-parameters)
...
+                USBD_CDC_TransmitPacket(&_local_task.usb_handle);
+            while (tx_result == USBD_BUSY) {
+                vTaskDelay(1);
+                tx_result = USBD_CDC_TransmitPacket(&_local_task.usb_handle);
+            }
+            if (tx_result != USBD_FAIL) {
+                uint8_t retries = 0;
+                while ((((USBD_CDC_HandleTypeDef *)
+                             _local_task.usb_handle.pClassData)
+                            ->TxState == 1) &&
```
caila-marashaj, line 165: is there any chance this transmit state gets stuck at 1 and then we get stuck here?

stm32-modules/thermocycler-gen2/firmware/host_comms_task/usb_hardware.c
```diff
@@ -215,5 +215,17 @@ void usb_hw_send(uint8_t *buf, uint16_t len) {
...
+    while (tx_result == USBD_BUSY) {
+        vTaskDelay(1);
+        tx_result = USBD_CDC_TransmitPacket(&_local_config.usb_handle);
+    }
+    if (tx_result == USBD_FAIL) {
+        return;
+    }
+    uint8_t retries = 0;
+
+    while ((((USBD_CDC_HandleTypeDef*)_local_config.usb_handle.pClassData)->TxState == 1) && retries++ < 10) {
```
caila-marashaj, line 228: where does `retries` get incremented?

vegano1: ```retries++```it's a post-increment. It returns the value, then increments it by one.

caila-marashaj: oh yeah I missed that ty

stm32-modules/heater-shaker/firmware/host_comms_task/freertos_comms_task.cpp
vegano1: if `TxState == 0` then this always breaks,
if `TxState == 1` It will only retry at most 10 times, it will break out after.

vegano1 (approved):
thank you

PR #540, opened 2025-10-23
feat(reference/vacuum-module): Fix gpio init issues with i2c3 + add hal_i2c_master_read and hal_i2c_master_write

vegano1:
# Overview

The GPIO init was not initiated properly for I2C3 because it was not set up. So, let's fix that by adding the corresponding instance to the HAL_I2C_MspInit and HAL_I2C_MspDeInit functions. Let's also add a hal_i2c_master_read and hal_i2c_master_write functions that can be used for sending raw data through the i2c bus, where a specific transaction is required.

# Changes

- Add `hal_i2c_master_read` and `hal_i2c_master_write` functions and expose them through the I2C class
- Fix an issue when initializing the GPIO for the I2C3 bus

# Testing

- [x] Talk to a device on the i2c3 line using the `hal_i2c_master_read` and `hal_i2c_master_write` functions.


sfoster1 (approved):
Okay, cool. Nice work to do it on the reference module too.

PR #541, opened 2025-10-29
feat(vacuum-module): hardware setup for pressure sensors

caila-marashaj:
## Overview
Hardware setup for atmospheric and vacuum pressure sensors

## Changelog

- policy files for vacuum and pressure sensors
- drivers for vacuum and pressure sensors
- GPIO Pin setup for I2C1, and EOC(end of conversion) pin on the vacuum pressure sensor

### DIFF
--- a/stm32-modules/vacuum-module/firmware/system/system_hardware.c
+++ b/stm32-modules/vacuum-module/firmware/system/system_hardware.c
@@ -117,9 +117,7 @@ uint16_t system_hardware_reset_reason() {
 }
 
 bool system_hardware_read_eoc_pin() {
-    // have to write the actual port + pin val in here
     uint8_t pin_val = HAL_GPIO_ReadPin(GPIOA, GPIO_PIN_1);
-    // also gotta see if this is active low or hi
     return pin_val == GPIO_PIN_SET ? true : false;
 }
 
--- a/stm32-modules/vacuum-module/firmware/system/system_hardware.c
+++ b/stm32-modules/vacuum-module/firmware/system/system_hardware.c
@@ -116,6 +116,13 @@ uint16_t system_hardware_reset_reason() {
     return reset_reason;
 }
 
+bool system_hardware_read_eoc_pin() {
+    // have to write the actual port + pin val in here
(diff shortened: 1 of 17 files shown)


stm32-modules/include/vacuum-module/vacuum-module/.ui_task.hpp.swp
sfoster1, line 1: looks like a bunch of vim temporary files - should add *.swp to gitignore

stm32-modules/vacuum-module/firmware/system/system_hardware.c
```diff
@@ -116,6 +116,13 @@ uint16_t system_hardware_reset_reason() {
     return reset_reason;
 }
 
+bool system_hardware_read_eoc_pin() {
+    // have to write the actual port + pin val in here
+    uint8_t pin_val = HAL_GPIO_ReadPin(GPIOA, GPIO_PIN_1);
```
sfoster1, line 120: don't do this as a workaround - that pin might be connected to something else or even not connected to anything and just floating around. if you want to stub out a gpio reading function, just `return false;` or `return true;`

stm32-modules/vacuum-module/firmware/system/system_hardware.c
```diff
@@ -116,6 +116,13 @@ uint16_t system_hardware_reset_reason() {
...
+    // also gotta see if this is active low or hi
+    return pin_val == GPIO_PIN_SET ? true : false;
```
sfoster1, line 121:
```suggestion
    return pin_val == GPIO_PIN_SET;
```

stm32-modules/vacuum-module/firmware/system/system_hardware.c
caila-marashaj: this comment is actually old I didn't mean to leave it in, it's from before I found the pin value. It is actually connected to A1 on the NFF schematic and is active high, I'll take out this comment

vegano1 (approved):

sfoster1 (approved):
Nice, ty!

PR #543, opened 2025-11-06
feat(vacuum-module): more sensor hardware + sensor read functionality

caila-marashaj:
## Overview
With this code, once a FreeRTOSTask is connected to the sensor policies, you should be able to read from all sensors, as well as interact with the eoc and reset pins on the MPR pressure sensor. 

Another note: the atmosphere pressure sensor (lps22df)'s pins aren't actually connected to the MCU as of this hardware revision. There's just a TODO noting that we should add that when it's connected.

## Changelog

- added read functionality to both pressure sensor policies
- moved sensor hardware initialization from `system_hardware` to its own file
- defined pin values in `main.h` and reference those, rather than the `HAL_GPIO` defs directly


stm32-modules/include/vacuum-module/vacuum-module/MPRLL0025PA00001A.hpp
```diff
@@ -25,15 +31,42 @@ struct StatusByte {
...
         if (_policy == nullptr) {
             _policy = policy;
         }
     }
 
+    auto read_pressure(int retries) -> uint16_t {
+        uint8_t read_buff[4] = {0x00};
+        bool sensor_busy = true;
+
+        policy->i2c_write(0x18 << 1, 0xAA, read_buff, 0);
```
vegano1: lets use the DEV_ADDRESS and MEASURE_PRESSURE_COMMAND defines, lets add write_buff.

stm32-modules/include/vacuum-module/vacuum-module/MPRLL0025PA00001A.hpp
```diff
@@ -25,15 +31,42 @@ struct StatusByte {
...
   public:
-    auto initialize(VacuumPressureSensorPolicy* policy) -> void {
+    auto initialize(MPRPolicy* policy) -> void {
         if (_policy == nullptr) {
             _policy = policy;
         }
     }
 
+    auto read_pressure(int retries) -> uint16_t {
+        uint8_t read_buff[4] = {0x00};
```
vegano1: this gets allocated on the call stack, so lets create a global read/write buffer like so.

```cpp
constexpr uint8_t PRESSURE_FRAME_LEN = 10;
std::array<uint8_t, PRESSURE_FRAME_LEN> READ_BUFF = {0};
std::array<uint8_t, PRESSURE_FRAME_LEN> WRITE_BUFF = {0};
```

then we can pass these buffers to the i2c_write and i2c_read functions like so
```cpp
policy->i2c_write(DEVICE_ADDRESS, MEASURE_PRESSURE_COMMAND, READ_BUFF.data(), 4);
```

stm32-modules/include/vacuum-module/vacuum-module/lps22df.hpp
```diff
@@ -13,13 +13,65 @@ concept LPS22DFPolicy = requires(P p, uint16_t dev_addr, uint16_t reg,
     { p.i2c_write(dev_addr, reg, data, size) } -> std::same_as<RxTxReturn>;
 };
 
+constexpr uint8_t DEVICE_ADDRESS = 0xB8;
```
vegano1: - lets use the 7bit address (0x5C) and bitshift this instead, so we know explicitly what address we are targeting.
- lets also add a comment that we are grounding the SA0 pin to use 0x5C address

vegano1: - lets use the 7bit address (0x5C) and bitshift this instead, so we know explicitly what address we are targeting.
- lets also add a comment that we are grounding the SA0 pin to use 0x5C address

vegano1 (commented):
Left a few comments

stm32-modules/include/vacuum-module/vacuum-module/MPRLL0025PA00001A.hpp
```diff
@@ -7,14 +7,20 @@
...
     { p.i2c_read(dev_addr, reg, data, size) } -> std::same_as<RxTxReturn>;
     { p.i2c_write(dev_addr, reg, data, size) } -> std::same_as<RxTxReturn>;
 };
 
-constexpr uint8_t DEV_ADDRESS = 0x18;
-constexpr uint8_t MEASURE_PRESSURE_BYTE = 0xAA;
+constexpr uint8_t DEV_ADDRESS = 0x30;
+constexpr uint8_t MEASURE_PRESSURE_COMMAND = 0xAA;
+constexpr uint8_t STATUS_BUSY_FLAG = 0x20;
+constexpr uint16_t OUTPUT_MAX = 15099494;
```
sfoster1: this doesn't seem right, it's not a 16 bit number

stm32-modules/include/vacuum-module/vacuum-module/MPRLL0025PA00001A.hpp
```diff
@@ -7,14 +7,20 @@
...
+constexpr uint16_t OUTPUT_MIN = 1677722;
```
sfoster1: ditto

stm32-modules/include/vacuum-module/vacuum-module/MPRLL0025PA00001A.hpp
sfoster1: these are blocking reads (presumably with task sleeps), is it ok for the task that calls this to be in the call for a while?

sfoster1 (changes requested):
Confused by those min and max reading numbers - those need 24 or so bits, so either the data type is wrong or the numbers are wrong

stm32-modules/include/vacuum-module/vacuum-module/MPRLL0025PA00001A.hpp
caila-marashaj: you're right- the pressure output is 3 bytes, it's a 24 bit number. I've changed it to `uint32_t`

stm32-modules/include/vacuum-module/vacuum-module/MPRLL0025PA00001A.hpp
vegano1: Maybe, our control loop needs to be ~300hz, in that time we need to

- Read ABS Pressure B and apply FIR filter
- Read ATM Pressure
- Calculate gauge pressure
- Feed the gauge pressure into a pressure control PID
- Feed  pid output Duty Cycle into the pump RPM control
- Repeat

Do you think the blocking i2c calls might be slow for this?

stm32-modules/include/vacuum-module/vacuum-module/lps22df.hpp
```diff
@@ -13,13 +15,71 @@ concept LPS22DFPolicy = requires(P p, uint16_t dev_addr, uint16_t reg,
...
-    auto initialize(LPS22DFPolicy* policy) -> void {
+    auto initialize(
+        atmosphere_pressure_sensor::hardware::AtmospherePressureSensorPolicy*
+            policy) -> void {
         if (_policy == nullptr) {
             _policy = policy;
         }
     }
+
+    auto read_pressure(int retries) -> uint16_t {
```
vegano1: lets add a retries default as a #define, so the caller does not have to supply this

stm32-modules/include/vacuum-module/vacuum-module/MPRLL0025PA00001A.hpp
caila-marashaj: took out the sleep

stm32-modules/include/vacuum-module/vacuum-module/MPRLL0025PA00001A.hpp
```diff
@@ -4,17 +4,32 @@
...
+constexpr uint8_t DEV_ADDRESS = 0x18 << 1;
+constexpr uint8_t MEASURE_PRESSURE_COMMAND = 0xAA;
+constexpr uint8_t STATUS_BUSY_FLAG = 0x20;
+constexpr uint32_t OUTPUT_MAX = 15099494;
+constexpr uint32_t OUTPUT_MIN = 1677722;
+constexpr uint32_t OUTPUT_RANGE_COUNTS = OUTPUT_MAX - OUTPUT_MIN;
+// range for this particular model is 0-25 PSI
+constexpr uint16_t PRESSURE_RANGE_PSI = 25;
+
+constexpr uint8_t PRESSURE_FRAME_LEN = 4;
```
vegano1, line 30: - Lets make this 10 so we have enough for longer frames if we need to
- Lets create a WRITE_BUFF so i2c_read and i2c_write use seperate buffers

stm32-modules/include/vacuum-module/vacuum-module/MPRLL0025PA00001A.hpp
```diff
@@ -25,15 +40,42 @@ struct StatusByte {
...
+    auto initialize(hardware::VacuumPressureSensorPolicy* policy) -> void {
         if (_policy == nullptr) {
             _policy = policy;
         }
     }
 
+    auto read_pressure(int retries) -> uint16_t {
+        bool sensor_busy = true;
+
+        _policy->i2c_write(DEVICE_ADDRESS, MEASURE_PRESSURE_COMMAND, READ_BUFF,
```
vegano1: - this should use a WRITE_BUFF
- this should write 1 byte instead of 0

vegano1: also, should this be using i2c_master_write instead? i remember we were having issues with the i2c_write

vegano1 (approved):

stm32-modules/include/vacuum-module/vacuum-module/lps22df.hpp
```diff
@@ -13,13 +15,100 @@ concept LPS22DFPolicy = requires(P p, uint16_t dev_addr, uint16_t reg,
...
+constexpr uint8_t DEVICE_ADDRESS = 0x5C << 1;
+// address CTRL_REG2 to read pressure
+constexpr uint8_t CTRL_REG2 = 0x11;
+// pressure reading is 4 bytes, starting with status at 0x27
+constexpr uint8_t PRESSURE_OUTPUT_REGISTER = 0x27;
+// write bit 1 to CTRL REG 2 to read pressure once
+constexpr uint8_t ONE_SHOT_PRESSURE_READ = 0x01;
+// pressure in hectoPascals is the 3 byte output value divided by the
+// sensitivity
+constexpr uint16_t SENSOR_SENSITIVITY = 4096.0f;
```
sfoster1: float constant but integral datatype

stm32-modules/include/vacuum-module/vacuum-module/lps22df.hpp
```diff
@@ -13,13 +15,100 @@ concept LPS22DFPolicy = requires(P p, uint16_t dev_addr, uint16_t reg,
...
+// bit 0 in the status byte is for pressure reading available
+constexpr uint8_t PRESSURE_READY_FLAG = 0x01;
+
+constexpr uint8_t PRESSURE_FRAME_LEN = 10;
+
+// Frame retry defaults
+constexpr uint8_t DEFAULT_RETRIES = 3;
```
sfoster1: no need for this to be a uint8_t since it's used on line 51 for comparing with a regular int; the loop variable and this should be the same type, whatever that type happens to be

stm32-modules/include/vacuum-module/vacuum-module/lps22df.hpp
```diff
@@ -13,13 +15,100 @@ concept LPS22DFPolicy = requires(P p, uint16_t dev_addr, uint16_t reg,
...
+            auto status_byte = RD_BUFF[0];
+            pressure_reading_ready =
+                static_cast<bool>(status_byte & PRESSURE_READY_FLAG);
+
+            if (pressure_reading_ready) {
+                break;
+            }
+        }
+        if (!pressure_reading_ready) {
+            // raise an error here
```
sfoster1: even if we're not raising an error yet we should probably not parse whatever's randomly in the data buffer (last read?) even in dev since it might make the control loop run away. really recommend making this function return `std::optional<double>` or something

stm32-modules/include/vacuum-module/vacuum-module/lps22df.hpp
```diff
@@ -13,13 +15,100 @@ concept LPS22DFPolicy = requires(P p, uint16_t dev_addr, uint16_t reg,
...
+    // Formulate a CMD frame
+    auto prepare_cmd_frame(uint8_t cmd, const uint8_t* data, uint8_t len)
+        -> uint8_t {
+        if (len > PRESSURE_FRAME_LEN) {
+            return 0;
+        }
+        // Add the header
+        WR_BUFF[0] = cmd;
+        // Copy data to the buffer starting from the header len.
+        for (uint8_t i = 1; i < len; i++) {
```
sfoster1: `memcpy(&WR_BUFF[1], data, len)`?

stm32-modules/vacuum-module/firmware/system/i2c_comms.cpp
```diff
@@ -27,12 +27,12 @@ auto I2C::i2c_read(uint16_t dev_addr, uint16_t reg, uint8_t *data,
 
 auto I2C::i2c_master_write(uint16_t dev_addr, uint8_t *data, uint16_t size)
     -> RxTxReturn {
-    auto ret = hal_i2c_master_write(bus, dev_addr, data, size);
+    auto ret = hal_i2c_master_write(this->bus, dev_addr, data, size);
```
sfoster1: 🤨

stm32-modules/vacuum-module/firmware/vacuum-pressure-sensor/vacuum_pressure_sensor_hardware.c
```diff
@@ -0,0 +1,64 @@
...
+    GPIO_InitStruct.Mode = GPIO_MODE_OUTPUT_PP;
+    GPIO_InitStruct.Pull = GPIO_NOPULL;
+    GPIO_InitStruct.Speed = GPIO_SPEED_FREQ_LOW;
+    HAL_GPIO_Init(SENSOR_A_EOC_PORT, &GPIO_InitStruct);
+
+    // sensor_a_reset
+    GPIO_InitStruct.Pin = SENSOR_A_RESET_PIN;
+    GPIO_InitStruct.Mode = GPIO_MODE_INPUT;
+    GPIO_InitStruct.Pull = GPIO_NOPULL;
+   GPIO_InitStruct.Speed = GPIO_SPEED_FREQ_LOW;
```
sfoster1, line 28: [eyebrow twitching at the one line at the wrong spacing]

stm32-modules/vacuum-module/firmware/vacuum-pressure-sensor/vacuum_pressure_sensor_hardware.c
```diff
@@ -0,0 +1,64 @@
...
+
+void vacuum_pressure_sensor_hardware_init(void) {
+    GPIO_InitTypeDef GPIO_InitStruct = {0};
+
+    __HAL_RCC_GPIOA_CLK_ENABLE();
+    __HAL_RCC_GPIOB_CLK_ENABLE();
+    __HAL_RCC_GPIOC_CLK_ENABLE();
+
+    // sensor_a_eoc
+    GPIO_InitStruct.Pin = SENSOR_A_EOC_PIN;
```
sfoster1, line 18: are these names or pin modes switched? the resets are all input and the end-of-conversions are all outputs

stm32-modules/vacuum-module/firmware/vacuum-pressure-sensor/vacuum_pressure_sensor_policy.cpp
```diff
@@ -23,10 +23,30 @@ auto VacuumPressureSensorPolicy::i2c_read(uint16_t dev_addr, uint16_t reg,
     return i2c_comms->i2c_read(dev_addr, reg, data, size);
 }
 
-auto VacuumPressureSensorPolicy::conversion_ended() -> bool {
-    return system_hardware_read_eoc_pin();
+auto VacuumPressureSensorPolicy::i2c_master_write(uint16_t dev_addr,
+                                                  uint8_t* data, uint16_t size)
+    -> RxTxReturn {
+    auto ret = i2c_comms->i2c_master_write(dev_addr, data, size);
```
sfoster1, line 29: light preference to just `return i2c_comms->...` but it's aesthetic

stm32-modules/include/vacuum-module/vacuum-module/lps22df.hpp
```diff
@@ -13,13 +15,100 @@ concept LPS22DFPolicy = requires(P p, uint16_t dev_addr, uint16_t reg,
...
+            // NOLINTNEXTLINE
+            WR_BUFF[i] = data[i - 1];
+        }
+        return len + 1;
+    }
+
+    static auto parse_pressure(const uint8_t* raw) -> double {
```
sfoster1: this can be a `const` function right

stm32-modules/include/vacuum-module/vacuum-module/MPRLL0025PA00001A.hpp
```diff
@@ -31,9 +52,67 @@ class MPRLL0025PA00001 {
...
+        WR_BUFF[0] = cmd;
+        // Copy data to the buffer starting from the header len.
+        for (uint8_t i = 1; i < len; i++) {
+            // NOLINTNEXTLINE
+            WR_BUFF[i] = data[i - 1];
+        }
+        return len + 1;
+    }
+
+    static auto parse_pressure(const uint8_t* raw) -> double {
```
sfoster1: this can be a `const` function right

stm32-modules/include/vacuum-module/vacuum-module/MPRLL0025PA00001A.hpp
```diff
@@ -31,9 +52,67 @@ class MPRLL0025PA00001 {
...
+        // Calculation of pressure value according to equation 2 of datasheet
+        // Equation 2: Pressure Output Function
+        auto press_counts =
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic)
+            (double)((int32_t)raw[3] + (int32_t)raw[2] * (int32_t)256 +
```
sfoster1: this is doing the same 24 bit bitshifting as the other one and it's way more readable if it's `(int32_t)raw[3] + ((int32_t)raw[2]) << 8) + ((int32_t)raw[1]) << 16)` or the same thing with `|` or whatever.

stm32-modules/include/vacuum-module/vacuum-module/MPRLL0025PA00001A.hpp
```diff
@@ -31,9 +52,67 @@ class MPRLL0025PA00001 {
...
+    // Formulate a CMD frame
+    auto prepare_cmd_frame(uint8_t cmd, const uint8_t* data, uint8_t len)
+        -> uint8_t {
+        if (len > PRESSURE_FRAME_LEN) {
+            return 0;
+        }
+        // Add the header
+        WR_BUFF[0] = cmd;
+        // Copy data to the buffer starting from the header len.
+        for (uint8_t i = 1; i < len; i++) {
```
sfoster1: `memcpy(&WR_BUFF[1], data, len)`?

stm32-modules/include/vacuum-module/vacuum-module/MPRLL0025PA00001A.hpp
```diff
@@ -2,19 +2,39 @@
...
+// maximum value of pressure range [bar, psi, kPa, etc.]
+constexpr double PMAX = 25;
+// minimum value of pressure range [bar, psi, kPa, etc.]
+constexpr double PMIN = 0;
+constexpr double PSI2MBAR = 68.9475729318;
+
+constexpr uint8_t PRESSURE_FRAME_LEN = 10;
+
+// Frame retry defaults
+constexpr uint8_t DEFAULT_RETRIES = 3;
```
sfoster1: should be the same data type as the loop var on line 60 (don't care which)

sfoster1 (changes requested):
bunch of little things, but also it kind of seems like the end of conversion and reset pins are flipped?

sfoster1 (approved):
Looks good to me!

PR #544, opened 2025-11-20
feat(vacuum-module): Add the pressure and pump control architecture + base gcodes

vegano1:
# Overview

This pull request implements the overall pressure control loops for the vacuum module, consisting of two messaging tasks for message handling, with each task having its own task for accurate timing to drive the process. This also adds the GCODES needed to regulate pressure and rpm for the respective tasks. This PR goes in conjunction with this monorepo PR [Opentrons/opentrons#20459](https://github.com/Opentrons/opentrons/pull/20459).

```
M120 - Set Pressure State
M121 - Get Pressure State
M122 - Set Pump State
M123 - Get Pump State
M124 - Set Vent State
```

See full GCODE documentation [here](https://opentrons.atlassian.net/wiki/spaces/PER/pages/5309956097/Vacuum+Module+Communication+Protocol)

Closes: [EXEC-1923](https://opentrons.atlassian.net/browse/EXEC-1923)

[EXEC-1923]: https://opentrons.atlassian.net/browse/EXEC-1923?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

# Changelog

- Added Pressure Control task that reads the pressure sensors and commands the pump task to regulate pressure
- Added Pump Control task that reads the tachometer and regulates the motor to match the target rpm
- Merged the AtmospherePressureSensorPolicy and VacuumPressureSensorPolicy into the PressurePolicy since the i2c interface is the same.
- Added is_device_ready to check if the I2C device is ready to communicate when initializing 
- Added DebugMessage to send messages to the host when debugging
- Fixed timing issues with the pump_hardware.c that were causing invalid rpm


sfoster1 (changes requested):
This looks good, but can we do these two changes:

1. Collect documentation on the pressure control strategy to a comment at the top of the pressure loop or something; I think I get what it's doing, but it probably wants a prose explanation that's in one place
2. The debug message thing is very scary. it would be really easy to get yourself a probabilistic crash by making a stack-stored string that you pass to that message as a pointer. instead, let's give the message object a buffer to own. we can keep it null-terminated if we want since the copy code can terminate at the max size of the buffer. just looking out for future us

sfoster1 (approved):
nice, thanks for the comments!

PR #547, opened 2026-01-14
feat(vacuum-module): set up FIR filtering for vacuum pressure sensor readings

caila-marashaj:
## Overview
This code implements the structure for a finite-impulse response filter, to be used on the vacuum pressure sensor. 

This sets up a buffer of filter coefficient values, and after each pressure sensor reading, convolves the most recent set of unfiltered pressure readings with the filter's coefficients. Currently, the coefficients are configured to make the filter function as a simple unweighted moving average, but this is with the expectation that we’ll optimize and update these values later on.

### DIFF
--- a/stm32-modules/include/vacuum-module/vacuum-module/mprll0025pa00001a.hpp
+++ b/stm32-modules/include/vacuum-module/vacuum-module/mprll0025pa00001a.hpp
@@ -41,7 +41,6 @@ constexpr uint8_t PRESSURE_FRAME_LEN = 10;
 constexpr int UNFILTERED_PRESSURE_BUFFER_LEN = 13;
 constexpr int FILTER_LEN = 13;
 constexpr int FILTERED_PRESSURE_BUFFER_LEN = 13;
-//constexpr double FILTER[FILTER_LEN] = {(1.0 / FILTER_LEN)};
 constexpr std::array<double, FILTER_LEN> FILTER = {(1.0 / FILTER_LEN)};
 
 // Frame retry defaults
@@ -70,12 +69,6 @@ class MPRLL0025PA00001 {
             // check device status
             ok = _policy->is_device_ready(device_address << 1);
         }
-        // initialize filter buffer to be an unweighted moving average
-        // TODO: optimize these filter coefficients
-//        double moving_average_coefficient = 1.0 / FILTER_LEN;
-//        for (int i = 0; i < FILTER_LEN; i++) {
-//            FILTER.at(i) = moving_average_coefficient;
-//        }
 
         return ok;
     }
@@ -103,10 +96,8 @@ class MPRLL0025PA00001 {
                 static_cast<bool>(status_byte & STATUS_BUSY_FLAG);
             if (!sensor_busy) {
                 pressure_mbar = parse_pressure(RD_BUFF.data());
-                return pressure_mbar;
-//                filter_pressure(pressure_mbar);
-//                return FILTERED_PRESSURE_MBAR[
-//                    filtered_pressure_buffer_index];
+                filter_pressure(pressure_mbar);
+                return FILTERED_PRESSURE_MBAR[filtered_pressure_buffer_index];
             }
         }
 
@@ -124,7 +115,6 @@ class MPRLL0025PA00001 {
         WR_BUFF[0] = cmd;
         // Copy data to the buffer starting from the header len.
         for (uint8_t i = 1; i < len; i++) {
-            // NOLINTNEXTLINE
             WR_BUFF[i] = data[i - 1];
         }
         return len + 1;
@@ -136,7 +126,6 @@ class MPRLL0025PA00001 {
         // The pressure counts are in big-endian (most significant byte first)
         // order, not little-endian.
         auto press_counts =
-            // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic)
             static_cast<double>(raw[1] << 16 | raw[2] << 8 | raw[3]);
         auto pressure_psi = (((press_counts - OUTPUT_MIN) * (PMAX - PMIN)) /
                              (OUTPUT_MAX - OUTPUT_MIN)) +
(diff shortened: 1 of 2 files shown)


stm32-modules/include/vacuum-module/vacuum-module/mprll0025pa00001a.hpp
```diff
@@ -37,8 +37,11 @@ constexpr double PMAX = 25;
 // minimum value of pressure range [bar, psi, kPa, etc.]
 constexpr double PMIN = 0;
 constexpr double PSI2MBAR = 68.9475729318;
-
 constexpr uint8_t PRESSURE_FRAME_LEN = 10;
+constexpr int UNFILTERED_PRESSURE_BUFFER_LEN = 13;
+constexpr int FILTER_LEN = 13;
+constexpr int FILTERED_PRESSURE_BUFFER_LEN = 13;
+double FILTER[FILTER_LEN] = {};
```
sfoster1: i think we can static initialize this since everything is `constexpr`, right?

stm32-modules/include/vacuum-module/vacuum-module/mprll0025pa00001a.hpp
```diff
@@ -131,13 +141,37 @@ class MPRLL0025PA00001 {
...
+    }
+
     Policy* _policy{nullptr};
     std::array<uint8_t, PRESSURE_FRAME_LEN> RD_BUFF = {0};
     std::array<uint8_t, PRESSURE_FRAME_LEN> WR_BUFF = {0};
     PressureSensorID _sensor_id{};
     uint8_t device_address{};
 
-    double pressure_mbar = {0};
+    double FILTERED_PRESSURE_MBAR[FILTERED_PRESSURE_BUFFER_LEN] = {0};
```
sfoster1: nit - these aren't constants so they shouldn't be SHOUTED

stm32-modules/include/vacuum-module/vacuum-module/mprll0025pa00001a.hpp
```diff
@@ -131,13 +141,37 @@ class MPRLL0025PA00001 {
...
+    double UNFILTERED_PRESSURE_BUFFER_MBAR[UNFILTERED_PRESSURE_BUFFER_LEN] = {
+        0};
+    int filtered_pressure_buffer_index = -1;
```
sfoster1: don't love this... can we just make this a `size_t` and start the math with buffer element 1

sfoster1 (changes requested):
Couple nits but looks good overall

stm32-modules/include/vacuum-module/vacuum-module/mprll0025pa00001a.hpp
sfoster1: when we have an actual pretuned filter, these will just be constant values here so we also might as well start that off here with constant values of 1/13

stm32-modules/include/vacuum-module/vacuum-module/mprll0025pa00001a.hpp
```diff
@@ -37,8 +37,12 @@ constexpr double PMAX = 25;
 // minimum value of pressure range [bar, psi, kPa, etc.]
 constexpr double PMIN = 0;
 constexpr double PSI2MBAR = 68.9475729318;
-
 constexpr uint8_t PRESSURE_FRAME_LEN = 10;
+constexpr int UNFILTERED_PRESSURE_BUFFER_LEN = 13;
+constexpr int FILTER_LEN = 13;
+constexpr int FILTERED_PRESSURE_BUFFER_LEN = 13;
+//constexpr double FILTER[FILTER_LEN] = {(1.0 / FILTER_LEN)};
```
vegano1: comment

stm32-modules/include/vacuum-module/vacuum-module/mprll0025pa00001a.hpp
```diff
@@ -66,6 +70,12 @@ class MPRLL0025PA00001 {
             // check device status
             ok = _policy->is_device_ready(device_address << 1);
         }
+        // initialize filter buffer to be an unweighted moving average
```
vegano1: comment

stm32-modules/include/vacuum-module/vacuum-module/pressure_task.hpp
```diff
@@ -264,11 +264,7 @@ class PressureTask {
             }
         }
 
-        // Use EMA pressure filter. TODO: change this to FIR filter
-        auto raw_pressure = _pressure_control.pressure_abs_b;
-        auto previous_pressure = _pressure_control.current_pressure;
-        auto current_pressure = (SENSOR_ALPHA * raw_pressure) +
-                                ((1.0F - SENSOR_ALPHA) * previous_pressure);
```
vegano1, line 271: lets remove SENSOR_ALPHA if unused

stm32-modules/include/vacuum-module/vacuum-module/mprll0025pa00001a.hpp
```diff
@@ -37,8 +37,12 @@ constexpr double PMAX = 25;
 // minimum value of pressure range [bar, psi, kPa, etc.]
 constexpr double PMIN = 0;
 constexpr double PSI2MBAR = 68.9475729318;
-
 constexpr uint8_t PRESSURE_FRAME_LEN = 10;
+constexpr int UNFILTERED_PRESSURE_BUFFER_LEN = 13;
+constexpr int FILTER_LEN = 13;
+constexpr int FILTERED_PRESSURE_BUFFER_LEN = 13;
```
vegano1: Do we need all these? Can we have one `FILTER_LEN` constant for everything?

vegano1 (commented):
Left a few comments

stm32-modules/include/vacuum-module/vacuum-module/mprll0025pa00001a.hpp
```diff
@@ -131,13 +138,37 @@ class MPRLL0025PA00001 {
...
     }
 
+    auto filter_pressure(double input_pressure_mbar) -> void {
+        ++filtered_pressure_buffer_index %= PRESSURE_BUFFER_LEN;
+        ++unfiltered_pressure_buffer_index %= PRESSURE_BUFFER_LEN;
+        unfiltered_pressure_mbar.at(filtered_pressure_buffer_index) =
+            input_pressure_mbar;
+        double filter_output = 0;
+
+        for (int i = 0; i < PRESSURE_BUFFER_LEN; i++) {
```
sfoster1, line 153: wait i'm still not quite sure about this filter. the way you generally want a FIR filter to work is that you take a signal sequence `S` of some length `l_1` and a filter kernel `K` of some length `l_2` and you convolve `K` and `S` to produce an output sequence `F` of, depending on what you want your edge condition to be, length `l_1 - l_2` or `max(l_1, l_2)` or `l_1 + l_2`. this is taking `K` and `S` and producing a scalar, which isn't quite right. 

so you run the full filter across the full captured data trace every time you get a new piece of data, and the full filtered output may change (depending on what your filter is), though you often (but not always!) just take the single most recent element of the filtered data.

sfoster1 (changes requested):
I think the convolution setup is a little wrong still.

sfoster1 (approved):
if it works it works

vegano1 (approved):

stm32-modules/include/vacuum-module/vacuum-module/mprll0025pa00001a.hpp
caila-marashaj: just replying to keep track- this was a bug and its resolved now

PR #548, opened 2026-01-27
feat(vacuum module): read the vacuum pressure sensors based on GPIO interrupt

caila-marashaj:
## Overview
In the interest of adhering to the vacuum pressure sensor’s preferred timing and hopefully seeing less saturation errors, we’d like to only request an I2C read when the sensor’s data ready pin is triggered. This pr sets up the HAL GPIO interrupts and callbacks for this functionality, and separates the pressure driver’s initiating the sensor read, from reading the data once the interrupt is triggered.


stm32-modules/include/vacuum-module/vacuum-module/mprll0025pa00001a.hpp
```diff
@@ -81,33 +88,30 @@ class MPRLL0025PA00001 {
 
     [[nodiscard]] auto get_pressure() const -> double { return pressure_mbar; }
 
-    auto read_pressure() -> double {
```
vegano1, line 84: lets keep the synchronous read_pressure function and add `start_pressure_read` separately so we can have both options.

stm32-modules/include/vacuum-module/vacuum-module/pressure_task.hpp
```diff
@@ -188,17 +191,23 @@ class PressureTask {
...
-                    [&](auto&& driver) -> bool {
-                        auto ok = driver.initialize(comms, sensor_id);
-                        if (ok) {
-                            driver.read_pressure();
-                        }
-                        return ok;
-                    },
-                    sensor.driver);
-                sensor.state = sensor.ok ? IDLE : SENSOR_ERROR;
+//                auto comms = policy.get_i2c_comms(sensor_id);
```
vegano1, line 194: i think we still need this

stm32-modules/include/vacuum-module/vacuum-module/pressure_task.hpp
```diff
@@ -223,6 +232,34 @@ class PressureTask {
         std::visit(visit_helper, message);
     }
 
+    auto send_pressure_reading_it(uint16_t GPIO_Pin) -> void {
```
vegano1, line 235: is this for testing?

vegano1 (commented):
left a few comments

PR #550, opened 2026-02-18
feat(vacuum-module): add product id for vacuum module.

vegano1:
Add the corresponding product id `0xEF40` for the vacuum module, so udev can correctly identify this device when plugged into the Flex.


sfoster1 (approved):
sweet!

PR #551, opened 2026-02-19
feat(vacuum-module): add gcodes to tune the pressure control PID parameters.

vegano1:
# Overview

Hardware testing needs to tune the pressure regulation for smoother performance, so let's expose the pressure-control parameters that directly affect it. Those parameters are

- The PID parameters (Proportional, Integral, Derivative)
- The `K_Velocity` Feed-Forward Gain, How much RPM to add for every 1 mbar/sec drop requested
- The `K_Holding` Feed-Forward Gain, Max RPM required to hold a deep vacuum against leaks.
- The Overshoot error Determines the threshold when the k_velocity and k_holding are ignored if we overshoot

# Changelog

- Adds M125 SetPressurePID GCode
- Adds M126 GetPressurePID GCode

### DIFF
--- a/stm32-modules/include/vacuum-module/vacuum-module/gcodes.hpp
+++ b/stm32-modules/include/vacuum-module/vacuum-module/gcodes.hpp
@@ -535,4 +535,117 @@ struct SetVentState {
     }
 };
 
+struct SetPressurePID {
+    /*
+     * M125- SetPressurePID tune the pressure regulation parameters
+     * */
+    std::optional<double> kp;
+    std::optional<double> ki;
+    std::optional<double> kd;
+    std::optional<double> overshoot;
+    std::optional<double> k_velocity;
+    std::optional<double> k_holding;
+    bool reset = false;
+
+    using ParseResult = std::optional<SetPressurePID>;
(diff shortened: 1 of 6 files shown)


stm32-modules/include/vacuum-module/vacuum-module/gcodes.hpp
```diff
@@ -535,4 +535,113 @@ struct SetVentState {
...
 
+struct SetPressurePID {
+    /*
+     * M125- SetPressurePID tune the pressure regulation parameters
+     * */
+    std::optional<double> kp;
+    std::optional<double> ki;
+    std::optional<double> kd;
+    std::optional<double> overshoot;
+    std::optional<float> k_velocity;
```
sfoster1: there any particular reason these are floats and the others are doubles?

sfoster1 (approved):
looks good, though maybe just make everything a double if most of them are

stm32-modules/include/vacuum-module/vacuum-module/gcodes.hpp
vegano1: nope, fixed.

PR #552, opened 2026-02-19
fix(build): add vacuum-module to set_build_vars script to deploy the firmware to aws.

vegano1:
We need this so we can deploy the vacuum module firmware binaries to the correct directory in aws


TamarZanzouri (approved):
Thank you!

sfoster1 (approved):
lgtm if it works!

PR #553, opened 2026-02-20
feat(vacuum module): add solid state tracking + duration

caila-marashaj:
## Overview
This pr implements timekeeping for running internal pressure control for a given duration:

- adds timing for reaching target pressure as well as maintaining vacuum for a hold time given by the `SetPressureState` gcode
- determines when target pressure is reached and maintains, by checking whether all pressure readings in the last 5 seconds are within a given tolerance of the target pressure

This is tested and looks good on hardware

## TODO

- write a unit test for this
- handle not reaching pressure in time

### DIFF
--- a/stm32-modules/include/vacuum-module/vacuum-module/pressure_task.hpp
+++ b/stm32-modules/include/vacuum-module/vacuum-module/pressure_task.hpp
@@ -269,10 +269,13 @@ class PressureTask {
         // TODO : make sure B is the right sensor to use here
         auto& sensor = get_sensor(ABS_PRESSURE_B);
         bool solid_state_pressure = std::visit(
-            [&](auto&& driver) -> double { return driver.solid_state_target_pressure(_pressure_control.target_pressure,
-                                            SOLID_STATE_PRESSURE_SAMPLES, SOLID_STATE_PRESSURE_TOLERANCE); },
+            [&](auto&& driver) -> double {
+                return driver.solid_state_target_pressure(
+                    _pressure_control.target_pressure,
+                    SOLID_STATE_PRESSURE_SAMPLES,
+                    SOLID_STATE_PRESSURE_TOLERANCE);
+            },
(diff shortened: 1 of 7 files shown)


stm32-modules/include/vacuum-module/vacuum-module/pressure_task.hpp
```diff
@@ -246,15 +321,9 @@ class PressureTask {
...
-            policy.start_pressure_control(false);
-            set_pump_state(false, 0);
-
-            _pressure_control.pid.reset();
-            _pressure_control.slew.reset();
-            _pressure_control.current_pressure = 0;
-            _pressure_control.target_rpm = 0;
-            _pressure_control.last_tick = 0;
-            return;
+            stop_vacuum(policy);
```
vegano1: lets return here so we dont fall through

sfoster1 (changes requested):
Looks good to me but add some tests!

stm32-modules/include/vacuum-module/vacuum-module/pressure_task.hpp
```diff
@@ -235,6 +257,84 @@ class PressureTask {
...
+
+    auto handle_pressure_control_outcomes() -> void {
+        if (!_pressure_control.enable_vacuum) {
+            stop_vacuum();
+        }
+        if (_pressure_control.handle_open_vent) {
+            _policy->set_vent_state(static_cast<bool>(VentState::OPEN));
+        }
+        if (_pressure_control.pressure_error != Error::NO_ERROR) {
+            send_ack_message(_pressure_control.responding_to_id,
```
vegano1: This should use `send_error_message`  instead, so no need to keep track of responding_to_id

vegano1 (approved):
Looking good, just left one comment

sfoster1 (approved):
🧪 🔜

PR #554, opened 2026-02-22
fix(vacuum-module): convert unweighted filter to EMI when filtering mprll0025pa00001a pressure readings.

vegano1:
# Overview

Let's convert the unweighted moving average filter to an exponential moving average filter that prioritizes newer values so we can filter out noise from the pump oscillations. Let's also fix an issue when calculating the filtered pressure the first time we take a reading, where the mostly zeroed buffer is used to calculate the pressure. This caused the system to output much lower pressure, resulting in the pump overcompensating on startup.
# Changelog

- Convert Unweighted Moving Average to Exponential Moving Average (EMI) filter with coefficients producing alpha = 0.2 for greater smoothing.
- Fill the pressure buffer with the first pressure it reads, so we don't generate artificially low values.

### DIFF
--- a/stm32-modules/include/vacuum-module/vacuum-module/mprll0025pa00001a.hpp
+++ b/stm32-modules/include/vacuum-module/vacuum-module/mprll0025pa00001a.hpp
@@ -4,6 +4,8 @@
 #include <cmath>
 #include <cstdint>
 
+#include "firmware/hardware_iface.hpp"
+
 namespace vacuum_pressure_sensor {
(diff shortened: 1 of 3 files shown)


sfoster1 (approved):
Looks good but please add some tests!

caila-marashaj (approved):
LGTM. I can take my filtering fix out of [this pr](https://github.com/Opentrons/opentrons-modules/pull/548) and we can coalesce around this

PR #555, opened 2026-03-06
feat(vacuum-module): Add a waste full detection mechanism using pressure slope monitoring

vegano1:
# Overview

When the waste container is full, we need to stop the pump and signal the Flex to pause the protocol, enter error recovery, and guide the user to empty the container. The way we wanted to do that before was with a pressure differential between the pressure in the manifold and the container; a big enough difference would trigger the error, nice and easy. However, the hardware design required a significant change, so we cannot do that anymore; instead, we had to find another way to detect when the waste is full.

Closes: [EXEC-1853](https://opentrons.atlassian.net/browse/EXEC-1853)

## Slope Monitoring Algorithm

This pull request implements a waste-full detection algorithm that leverages the fact that air is compressible and the target pressure is reached faster / slower depending on how full the waste container is with liquid. This algorithm works in 2 phases: the ramp phase and the hold phase.

### Ramp Phase

This lets us detect when the waste is already full and the floater is either closing the valve or the suction force from the pump forces it closed, and we see a fast pressure drop. You can see below that we have 2 points, P_WINDOW_START and P_WINDOW_END, which determine when we read pressure and calculate how fast we get between these points, adjusted for how deep in vacuum we are.

Screenshot 2026-03-13 at 3 23 51 PM

The air in the container acts like a capacitor: the more liquid in the container, the less air there is, and the faster that air is to compress and reach the target pressure. When the floating valve in the container is closed, it blocks the airflow from the rest of the manifold, and we measure how fast the pressure gets to the target. Depending on this, we flag the waste as full.

There is also a post-boot first run where we track how fast the system waas able to reach the target pressure, and mark that as the "Baseline" that we use in future runs to determine how fast the pressure changes relative to that baseline. If the pressure changes N times faster than the baseline, then we flag the waste as full.

### Hold Phase

When we are holding, the system is regulating at the target pressure, while the waste container fills up slowly. In this case, we need to track the change of pressure over time and determine and flag as full when the delta spikes too quickly or we get a small accumulation of delta over time, which means the waste is filling up very slowly, see below.

WASTE_DETECTION_SLOW

Since we are taking a look at pressure deltas over time, we need to ignore huge positive pressure changes caused when the manifold runs out of liquid and a huge rush of air enters the system.

## Change log

- Move the pressure regulation from the Pressure Task into a dedicated PressureController class to
    - Decouple the FreeRTOS Task which orchestrates messages from the pressure-regulating algorithm
    - Makes it easier to test the pressure regulating algorithm
- Add a WasteDetection class that implements the Waste Detection Slope Monitoring algorithm
- Switch the set_vent_state from bool to VentState enum through the stack
- Change vent to OPENED after we are done with pressure regulation by default
- Reduce the Pump task ramp rate from 10 -> 5 and max pwm jump 3 -> 1 to stabilize the pump
- Decrease the filtering for the RPM used by the pressure task to respond faster to changes in motion.

## Review Requests

- Anything I'm missing?

## Notes

- This was validated using water
- We need to do more testing once we are in EVT to validate other liquids

[EXEC-1853]: https://opentrons.atlassian.net/browse/EXEC-1853?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


stm32-modules/include/vacuum-module/vacuum-module/pressure_task.hpp
```diff
@@ -553,13 +556,225 @@ class PressureTask {
...
+    auto check_waste_state(uint32_t timestamp) -> bool {
+        if (!_pressure_control.enable_vacuum || _pressure_control.waste_full) {
+            return false;
+        }
+
+        double current_p = _pressure_control.pressure_abs_b;
+        double p_atm = _pressure_control.pressure_atm;
+        double target_abs = _pressure_control.target_pressure;
+        double total_vacuum_range = p_atm - target_abs;
+        if (total_vacuum_range < 100.0) {
```
sfoster1: what's this constant?

sfoster1 (changes requested):
This looks like a correct and good approach to me, nice!

However, it's also pretty complex and stateful code, and the control system is mixed in with the task in a way that makes it pretty tough to read and more importantly tough to test, since the task only runs when it gets messages, and of course writing tests that command messages really sucks.

Instead, let's pull the controller out into something self contained, or even one unit for the controller and one unit for the waste monitoring. They can have interfaces that look like `run(timestamp, current_pressure)` and `target(target_pressure)` and now it's really easy to write tests that set up specific conditions.

Also let's write the tests afterward.

stm32-modules/include/vacuum-module/vacuum-module/pressure_task.hpp
```diff
@@ -31,12 +31,15 @@
...
- * The pressure control is stopped when the `_pressure_control.enable_pump`
+ * The pressure control is stopped when the `_control_state.enable_pump`
  * is false. We stop the `PressureControlMessage` emitter task, send a stop
  * `SetPumpMessage` message to the PumpTask, then reset all relevant internal
- *  _pressure_control variables.
+ *  _control_state variables.
  */
 
+#pragma GCC push_options
+#pragma GCC optimize("O0")
```
sfoster1: why?

sfoster1 (changes requested):
This looks good and I really like the split, but what's up with the O0 in the pressure task?

stm32-modules/include/vacuum-module/vacuum-module/pressure_task.hpp
vegano1: haha, sorry! This was leftover from debugging!

caila-marashaj (approved):
Nice!!

stm32-modules/include/vacuum-module/firmware/pressure_controller.hpp
```diff
@@ -0,0 +1,120 @@
...
+    auto get_state() -> ConfigState { return state; }
+
+    [[nodiscard]] auto get_target_rpm() const -> double { return target_rpm; }
+
+  private:
+    PID _pid;
+    SlewRateLimiter _slew;
+    ConfigState state{};
+    double prev_target_mbar_ = 0.0;
+    double target_rpm = 0.0;
```
vegano1, line 119: Reset these two in the reset function

sfoster1 (approved):
Looks good to me, nice! Love the refactors!

PR #556, opened 2026-03-07
feat(vacuum-module): add ability to parse temperature from lps22hb

vegano1:
The lps22hb pressure sensor has a built-in temperature sensor, so let's expose this functionality so we can get the temperature inside the module.


sfoster1 (approved):
🌡️

PR #559, opened 2026-03-23
feat(vacuum-module): add pressure not reached + relative tolerance errors

vegano1:
# Overview

This pull request adds the ability to set a timeout in seconds when setting the pressure state. We also need to be able to set the percent error that the target is off by, so the user can customize this.

# Change log

- Send `PRESSURE_NOT_REACHED` error when the target pressure is not reached within the given timeout_s
- Add `relative_tolerance_pct` to set the pressure target error tolerance percent
- Added `A1` revision for EVT build out


sfoster1 (approved):
nice

PR #560, opened 2026-03-27
fix: correct reset reason logic

sfoster1:
The reset reason logic had a couple issues
- Not all the RCC flags are about reset reasons; some of them are more in the CC than the R (of Reset and Clock Control) and are about monitoring the readiness of internal clock sources. Omit these
- The logic would only emit the first positive flag it found, which in the current implementation was always one of the clock readiness flags. Instead, emit all positive flags
- There are slightly different reset-reason flags available on the F3 (which is used in the heater-shaker) and the G4 (which is used everywhere else); adapt the code on the heater-shaker to that

## testing
- can't really unit test this one since it's all about parsing HAL flags
- put this on the specified modules; restart them; and see what M114 says. it should say 2, 8, or 128 (or possibly a mix of those - this is a bitflag field and it's possible that power-on reset and low power reset may collide, so 130)

### DIFF
--- a/stm32-modules/flex-stacker/firmware/system/system_hardware.c
+++ b/stm32-modules/flex-stacker/firmware/system/system_hardware.c
@@ -1,52 +1,39 @@
+#include "firmware/system_hardware.h"
+
+#include "main.h"
 #include "stm32g4xx_hal.h"
-#include "stm32g4xx_hal_rcc.h"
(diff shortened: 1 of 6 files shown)


vegano1 (approved):
I ran this on the vacuum module a few weeks ago, saw no issues.

PR #561, opened 2026-03-27
fix(vacuum-module): fix init issue with lps22hb resulting in bad pressure values.

vegano1:
We need to wait for both the pressure AND temperature states to be ready when initializing the LPS22HB atmospheric pressure sensor.


sfoster1 (approved):
ope

PR #563, opened 2026-05-18
feat(vacuum-module): add vacuum duration to GetPressureState.

vegano1:
We need to know how much time we have left under vacuum so we can monitor the task on the python side.


ryanthecoder (approved):
lgtm

caila-marashaj (approved):

PR #566, opened 2026-05-20
fix(heater-shaker): Fix race condition

ryanthecoder:
fix race condition in power control code. gets rid of false positive short/open circuit errors.


sfoster1 (approved):
Nice! Let's merge #560 before merging this so it ends up as two commits, though.

PR #567, opened 2026-06-01
fix(vacuum-module): set correct 8-bit DAC range of 256 instead of 4096 when setting the vent voltage.

vegano1:
We are using an 8-bit DAC, so the full range is 256 instead of 4096.


sfoster1 (approved):
oh lol

PR #568, opened 2026-06-04
feat(vacuum-module): add notify mechanism to the pump_task so we can use duration + waste detection logic of the pressure task.

vegano1:
# Overview

The PressureTask is responsible for tracking how long the vacuum should run when duration > 0, as well as for the waste detection mechanism. The pump task needs access to this logic, so let's create a new internal `NotifyPumpRunMessage` that the PumpTask can send to the PressureTask to enable monitor-only mode, so it does NOT send control messages to the pump since it's being controlled "manually".

# Changelog

- Add `duration`, `timeout`, `rate`, and `vent_after` args to the SetPumpState (M122) GCODE message.
- Add `NotifyPumpRunMessage` message so we can notify the PressureTask to go into `monitor-only` mode.
- When in `monitor-only` mode, set the target pressure by using the duty cycle 0-100 as a percent of the atmospheric pressure.


sfoster1 (approved):
ok, cool!

PR #569, opened 2026-06-08
fix(heater-shaker): fix PWM deadlock

mjhuff:
# Overview

Fixes a bug where the HS permanently stops heating after a `deactivate_heater()` call, requiring a hard power cycle to recover. The heater firmware reports no error (`M411 OK`) and retains its setpoint (`M105` shows `T:target`), but the PWM hardware is never activated.

The heater pad circuit check ISR (`HAL_TIM_OC_DelayElapsedCallback`) uses an `update_lock` flag to prevent PWM changes during open/short circuit checks. This check is a multi-step state machine spread across several ISR invocations:

- ISR invocation 1: sets `update_lock = true`, transitions state machine to `OPEN_CHECK_STARTED`
- ISR invocations 2-3: performs open/short checks, transitions to `PREP_RUNNING`
- ISR invocation 4: clears `update_lock = false`, restores overcurrent detection

If `heater_hardware_power_disable()` is called between steps 1 and 4 (ex., from an `M106` DeactivateHeater GCODE while the heater is actively running), the PWM channels are stopped and the ISR can never fire again. `update_lock` remains permanently stuck true.

When a subsequent `M104` (SetTemperature) re-enables heating, `heater_hardware_power_set()` sees `update_lock == true` and caches the PWM setting instead of starting the PWM hardware. The PID controller keeps computing, the task reports no error, but the heater pad never receives power.

It seems reasonable to assume timing window is approximately 1-3ms out of every ~1 second circuit check cycle, consistent with the observed rare occurrence rate.

## Test Plan

It's possible to validate the fix using a script sending an enormous amount of GCODES (see Seth's comment). One can also validate by running the following protocol enough times, each run has roughly a 95% chance of catching the bug. A few clean runs is sufficient for proving the fix:

```
from opentrons import protocol_api

metadata = {
    "protocolName": "HS Heater Race Condition Repro",
}

requirements = {"apiLevel": "2.28", "robotType": "Flex"}


def run(protocol: protocol_api.ProtocolContext) -> None:
    hs = protocol.load_module("heaterShakerModuleV1", "D1")
    plate = hs.load_labware("nest_96_wellplate_200ul_pcr_full_skirt")
    hs.close_labware_latch()

    target_temp = 65
    settle_seconds = 10
    max_cycles = 1000

    for i in range(1, max_cycles + 1):
        hs.set_and_wait_for_temperature(target_temp)

        protocol.delay(seconds=settle_seconds)

        hs.deactivate_heater()
        protocol.delay(seconds=5)

    hs.deactivate_heater()
    hs.open_labware_latch()

```

### DIFF
--- a/stm32-modules/heater-shaker/firmware/heater_task/heater_hardware.c
+++ b/stm32-modules/heater-shaker/firmware/heater_task/heater_hardware.c
@@ -294,6 +294,9 @@ void heater_hardware_power_disable(heater_hardware* hardware, uint8_t task_conte
     HAL_TIM_PWM_Stop_IT(&internal->pad_tim, HEATER_PAD_SHORT_CHECK_TIM_CHANNEL);
     HAL_TIM_PWM_Stop_IT(&internal->pad_tim, HEATER_PAD_OPEN_CHECK_TIM_CHANNEL);
     internal->heater_started = false;
+    internal->update_lock = false;
+    internal->update_while_locked = false;
+    internal->period_count = 0;
     if (!heatpad_in_error_state()) {
         internal->heatpad_cs_status = IDLE;
     }


sfoster1 (approved):
yeah, this is a valid bad path. it's actually a max of two (full) timer periods of vulnerability; that ISR fires when each compare channel so configured, in this case the open and short circuit pulse channels, hit their compare value. looks something like this:

- Let the timer base period be 18000 at 18kHz (this means the timer rolls over 1000/sec)
- Let the OC compare be channel 2 with compare value of 1800 (10%)
- Let the SC compare be channel 4 with compare value of 16200 (90%)
- Also the actual pad PWM is channel 3, compare varies

Then when the checks are enabled and the heater is enabled,
- `HAL_TIM_OC_DelayElapsedCallback` fires three times per period (once for each channel's compare value matching the timer). One is ignored because the channel is the timer active channel
- After 2001 interrupts, which should be after the open circuit invocation of the 1000th timer periods, which should be 1 second, the checks are active
- That first 90% check sets `update_lock = true` and moves to `OPEN_CHECK_STARTED`
- Next period, the first interrupt should be the OC interrupt and we're in `OPEN_CHECK_STARTED`, so we execute, move to `SHORT_CHECK_STARTED`, defer to next period
- Next period, the second interrupt is the SC check, we move to `PREP_RUNNING`, and immediately (since it's not an else if) clear `update_lock = false`

so a `heater_hardware_disable()` in there would definitely cause the failure by leaving `update_lock = true` forever. also the period count is never reset so probably false positives would happen more often than we think.

this will be pretty annoying to repro, but one way to do it is to have a script that connects to the HS serial and sends `M104.D S0.25` (set heater power 25%; this avoids waiting for temperature conversion callbacks for more accurate timing, and is above the 20% power minimum threshold for OC checks to widen the window); waits 1s with like +-20ms synthetic jitter; and then sends `M106`. If no error happens in a couple hours, mess with the jitter.

We could also cut the period threshold down to like 100ms to speed this up 10x, the limiting factor on how long it takes is waiting 1s after the heater turns on.

PR #570, opened 2026-06-15
fix(vacuum-module): Fixes some issues with the SetPumpState message logic and acks.

vegano1:
# Overview

Addresses several issues with the SetPumpState message handling found while running the QC test script on the DVT units

# Changelog

- Fixes an issue with the SetPumpState ignoring the ramp logic for the duty cycle
- Fixes an issue with the SetPumpState ack getting blocked by the PumpControl messages.
- Fixes an issue with the SetPumpState ramp rate being ignored.


caila-marashaj (approved):
This fixed the issues we were seeing. Thanks!

PR #571, opened 2026-06-17
fix(vacuum-module): fixes an issue with NotifyPumpRunMessage not working correctly for SetPumpMessage.

vegano1:
- Forgot to set vent_after in NotifyPumpRunMessage
- Make sure we always call set_pump_state(false, 0) when we stop the vacuum_pump so the vacuum is stopped when running in "manual" mode via the `SetVacuumPower` command.


caila-marashaj (approved):
looks good!

PR #572, opened 2026-06-23
fix(vacuum-module): always report current gauge pressure as calculated by abs - atm.

vegano1:
Don't report false current pressure by clamping gauge pressure to 0 when the pump is not running, as we still need the current pressure when the system is not running but still venting.


sfoster1 (approved):
makes sense to me!

PR #573, opened 2026-07-16
fix(vacuum-module): enter DFU via option bytes

vegano1:
## Summary

The vacuum module was not actually getting into DFU after a `dfu` gcode, so host updates could not find it with `dfu-util`. Jumping straight into the system bootloader from the running app does not bring up USB DFU on this board.

This change enters DFU by programming the boot option bytes and resetting into the system bootloader instead. On the next app boot we put the option bytes back so normal flash boot still works. Also fixed the `dfu OK` path so the response can go out before USB is disconnected.

Verified on hardware with ST-Link and with `update_module_fw.py` on a Flex.

## Test plan

- [x] Flash full image, send `dfu`, confirm `0483:df11` with `dfu-util -l`
- [x] Leave DFU / reboot app, confirm normal boot option bytes restored
- [x] Full Flex update via `update_module_fw.py --module vacuum-module --force`

### DIFF
--- a/stm32-modules/vacuum-module/firmware/system/system_hardware.c
+++ b/stm32-modules/vacuum-module/firmware/system/system_hardware.c
@@ -3,6 +3,8 @@
 #include "main.h"
 #include "stm32g4xx_hal.h"
 #include "stm32g4xx_hal_cortex.h"
+#include "stm32g4xx_hal_flash.h"
+#include "stm32g4xx_hal_flash_ex.h"
 #include "stm32g4xx_hal_rcc.h"
 #include "stm32g4xx_hal_tim.h"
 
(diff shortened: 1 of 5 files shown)


stm32-modules/vacuum-module/firmware/system/system_hardware.c
```diff
@@ -73,48 +75,124 @@ static void save_reset_reason() {
...
-    // addresses) and to have a naked function you need to define it as a
-    // function, not a function pointer, and we don't statically know the
-    // address here since it is whatever's contained in that second word of the
-    // bsystem memory region.
-    asm volatile("bx %0"
-                 :  // no outputs
-                 : "r"(*sysmem_boot_loc)
-                 : "memory");
+    __enable_irq();
+    asm volatile(
```
caila-marashaj, line 186: extreme times call for extreme measures

caila-marashaj (approved):

PR #574, opened 2026-07-17
vacuum-module: pressure control and pump coordination improvements

vegano1:
## Summary

Improves vacuum closed-loop control: correct timebases and sample timing, fix vacuum-depth percent mapping, and clean up target-reached detection. Also softens the approach to the final setpoint (slew, feed-forward, integral, overshoot brake) and keeps pump RPM ramping aligned with the pressure trajectory.

## Test plan

- [x] set vacuum, confirm target reach and hold without large overshoot
- [x] Manual pump / vacuum-depth percent modes still match expected depths

### DIFF
--- a/stm32-modules/include/vacuum-module/vacuum-module/pump_task.hpp
+++ b/stm32-modules/include/vacuum-module/vacuum-module/pump_task.hpp
@@ -27,7 +27,7 @@ static constexpr const double PUMP_STOP_RPM_THRESH = 500;
 static constexpr const float MIN_RAMP_RATE = 1;      // rpm/s
 static constexpr const float DEFAULT_RAMP_RATE = 5;  // rpm/s
 static constexpr const float MAX_RAMP_RATE = 500;    // rpm/s
-static constexpr const double MAX_PWM_JUMP = 1;      // pwm/tick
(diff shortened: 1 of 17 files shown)


stm32-modules/include/vacuum-module/vacuum-module/pump_task.hpp
```diff
@@ -142,7 +142,7 @@ class PumpTask {
         auto last_tick =
             _pump_control.last_tick ? _pump_control.last_tick : timestamp;
         auto delta_s = (timestamp - last_tick) * MS_TO_SECONDS;
```
ryanthecoder: Not related to this PR but:
if NOT _pump_control.last_tick, then delta_s will be (timestamp-timestamp * MS_TO_SECONDS)  == 0
Then when we go to compute the pid later, we will create a divide by 0 error when calculating the D term

caila-marashaj (approved):
looks good

ryanthecoder (approved):
The changes look good but we should put some safeguards around that potential divide by 0 error.

stm32-modules/include/vacuum-module/vacuum-module/pump_task.hpp
vegano1: nice call

PR #575, opened 2026-08-06
fix(vacuum-module): stabilize pressure hold regulation

vegano1:
## Summary

- Keep integral active while still short of the final vacuum target so hold no longer sits several mbar shallow
- Soft-land the pressure slew near the target (`slew_end_fraction` default 0.30) to cut deep overshoot/ripple at -600…-800
- Expose approach band (`A`) and slew end fraction (`S`) on M125/M126 for runtime tuning

## Test plan

- [x] Bench hold sweep -100…-800 mbar (60s/90s) on water, 3×0.2 mm fixture
- [x] Confirm M126 reports `A`/`S` and M125 can set them

### DIFF
--- a/stm32-modules/include/vacuum-module/firmware/pressure_controller.hpp
+++ b/stm32-modules/include/vacuum-module/firmware/pressure_controller.hpp
@@ -27,6 +27,8 @@ static constexpr const double OVERSHOOT_DEPTH_PCT = 0.02F;
 static constexpr const double MAX_OVERSHOOT_MBAR = 20.0F;
 // Feed-forward tapers to zero within this band above the slewed trajectory.
 static constexpr const double APPROACH_BAND_MBAR = 80.0F;
+static constexpr const double MIN_APPROACH_BAND_MBAR = 1.0F;
+static constexpr const double MAX_APPROACH_BAND_MBAR = 500.0F;
 static constexpr const double ATM_PRESSURE_MBAR = 1013.25;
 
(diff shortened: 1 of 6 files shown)


stm32-modules/include/vacuum-module/firmware/pressure_controller.hpp
```diff
@@ -21,11 +21,29 @@ static constexpr const double K_VELOCITY = 20.0F;
...
 static constexpr const double K_HOLDING = 43.0F;
-// Disables Velocity and Holding Gain if target is overshot
+// Disables Velocity and Holding Gain if target is overshot.
+// Keep the brake band tight so hold regulation does not wander several mbar
+// past the final setpoint before the pump turns off.
 static constexpr const double OVERSHOOT_ERROR = -2.0F;
-static constexpr const double OVERSHOOT_DEPTH_PCT = 0.02F;
-static constexpr const double MAX_OVERSHOOT_MBAR = 20.0F;
-// Feed-forward tapers to zero within this band above the slewed trajectory.
+static constexpr const double OVERSHOOT_DEPTH_PCT = 0.005F;
```
sfoster1: is this a percent (as in, 0.005%) or is it a ratio (as in, 0.5%)?

sfoster1 (approved):
ok, sure. i kind of hate these control systems that are a PID controller with more and more epicycles layered on top, but sometimes you've just got phase delays and if it works it works

PR #576, opened 2026-09-03
fix(vacuum-module): detect waste-full as sealed deadhead, not ramp timing

vegano1:
## Summary

The Waste-full detector checks if we are in a sealed deadhead for long enough. This means we are near the target, but the pump needs to do relatively little work to hold the required vacuum.

Companion host PR: https://github.com/Opentrons/opentrons/pull/22401

### How does this work

When waste-full detection is enabled, every tick we check:

1. Are we near target?: `|current − target| < PRESSURE_TOLERANCE` mbar
    - `current` is the EMA-smoothed pressure of B
    If false, Reset `near_target_ms_` and `sealed_hold_ms_`
2. Debounce: Stay near target for `NEAR_TARGET_MS`
    If false, don't add or subtract from `sealed_hold_ms_`
3. Sealed?: All the following conditions must be true to be a sealed deadhead
    - Commanded RPM ≥ 1 - Pump is actually being asked to run
    - |A−B| < H mbar    - Veto for large leaks; this is mostly sensor-to-sensor offset, but when there is a wide-open path, the sensors show different pressures. Not the main discriminator; that's G below.
    - (G = RPMcmd / vacuum) < Gcap - The work the pump is doing relative to the current vacuum.

        A lower G means a full deadhead; a higher G means empty or leaky. Gcap is our tuning knob; lowering it decreases sensitivity, making it easier to avoid false positives but harder to trip when actually full. Raising the value increases the sensitivity, so it makes it easier to trip when full, but increases the chances of false positives when empty.

    If false, the sealed clock decays by the same dt. This way 1-tick PID blips don't reset the whole thing. We subtract dt from `sealed_hold_ms_`, AND we don't do the trip check on this tick.
4. Hold time: Sealed clock ≥ `stable_hold_ms`; have we stayed "sealed" for N
    seconds?
    - If false, keep watching.
    - If true, trip the system and return Waste-Full error. This stays latched until a reset or a new vacuum command comes in.

## Test plan

- [ ] Have the hardware testing team validate the new algorithm

### DIFF
--- a/stm32-modules/include/vacuum-module/firmware/waste_detector.hpp
+++ b/stm32-modules/include/vacuum-module/firmware/waste_detector.hpp
@@ -1,103 +1,56 @@
 #pragma once
-#include <array>
+#include <algorithm>
 #include <cmath>
 #include <cstdint>
 
-#include "systemwide.h"
-
 namespace waste_detector {
+
+// Waste-full is inferred, not measured. Sensors A and B sit in series
+// inside the module, downstream of the waste-jug floater, so |A-B| is
+// orifice flow after the jug.
+//
+// Full: tiny sealed volume. Hold is a deadhead (low conductance, no
+// orifice flow). Empty leaky: higher RPM/depth in hold.
 
 enum class WasteFullError : uint8_t {
     NO_ERROR = 0,
-    MAX_DELTA_TIC_ERROR = 1,
-    OVER_TARGET_ERROR = 2,
-    RISE_TOO_FAST_ERROR = 3,
-    FAST_BASELINE_ERROR = 4,
-    FIRST_RUN_SLOW_ERROR = 5,
-    SUDDEN_BLOCKED_ERROR = 6,
-    CUMMULATIVE_BLOCKED_ERROR = 7,
-    FLOW_STABLE_FULL_ERROR = 8,
+    FLOW_STABLE_FULL_ERROR = 1,
 };
 
-// Window thresholds: Measure from p1 depth to p2 depth
-static constexpr const double WASTE_WINDOW_START_PCT = 0.10F;
-static constexpr const double WASTE_WINDOW_END_PCT = 0.95F;
-// Resets the baseline if the new target pressure is this percent difference
-static constexpr const double BASELINE_DEPTH_RESET = 0.30F;
-// Compare to the learned baseline. Faster than this factor is full.
-static constexpr const double BASELINE_FAST_FACTOR = 0.50F;
-// Hold-phase vacuum increase per tick that flags a slam-shut.
-static constexpr const double MAX_RISE_PER_TICK = 30.0;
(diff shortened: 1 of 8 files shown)


stm32-modules/include/vacuum-module/firmware/waste_detector.hpp
```diff
@@ -1,82 +1,56 @@
...
+    double g_sealed_max = G_SEALED_MAX;
+    double flowing_dp_mbar = FLOWING_DP_MBAR;
+    double stable_hold_ms = STABLE_HOLD_MS;
+    double stable_hold_deep_ms = STABLE_HOLD_DEEP_MS;
+    double min_waste_depth_mbar = MIN_WASTE_DEPTH_MBAR;
 };
 
 auto inline calculate_flow_per_second(double pressure_a, double pressure_b,
-                                      double smoothed_delta_p) -> double {
+                                      double& smoothed_delta_p) -> double {
```
sfoster1: is this intentional? if so, it's a bugfix that probably needs to be highlighted more in the docs

stm32-modules/include/vacuum-module/firmware/waste_detector.hpp
```diff
@@ -85,274 +59,153 @@ auto inline calculate_flow_per_second(double pressure_a, double pressure_b,
...
-            config.p_filter_alpha = c.p_filter_alpha;
-        }
-        if (c.min_window_time > 0) {
-            config.min_window_time = c.min_window_time;
-        }
-        if (c.max_window_time > 0) {
-            config.max_window_time = c.max_window_time;
-        }
+        config = c;
+        config.p_filter_alpha = (c.p_filter_alpha <= 0.0)
```
sfoster1: this is a bit of a code smell. is this meant to detect "unconfigured"?

sfoster1 (changes requested):
I think I see what this is doing, but the PR description and a lot of comments are way too much jargon and need to be something that somebody looking at this in a year can understand. also a couple nits with some parameter typing and config stuff.

stm32-modules/include/vacuum-module/firmware/waste_detector.hpp
vegano1: This was preventing 0 from basically turning off incoming pressures altogether since the EMA would ignore them fully. However, this is not a concern since the alpha we set is > 0. I added clamps to these valid config values, but they might not be right, since we want to allow hardware testing to map bad values to good values.

stm32-modules/include/vacuum-module/firmware/waste_detector.hpp
vegano1: it was intentional, but now ive made the function part of the class so we dont have to pass this in by ref.

sfoster1 (approved):
Nice, thank you!

PR #577, opened 2026-09-16
fix(vacuum-module): Remove waste-full detection from pump control path.

vegano1:
[Waste-detection](https://github.com/Opentrons/opentrons-modules/pull/576) now works by detecting a deadhead situation, which requires pressure control. So lets only do waste-full detection when we are pressure regulating and not when we are doing pump (power) control.


sfoster1 (approved):
ah, makes sense!

